PR #38, opened 2015-03-02
Minor fix for autotest script

rohgarg:
This patch adds a fix for a bug in the autotest.py script
that would preclude slower CPUs from having extra time
for process start up. The problem was because of the
extra whitespace characters in the output of `uname -m`
used by the script to detect the architecture.


karya0:
LGTM.

PR #41, opened 2015-03-16
Better vDSO handling

karya0:
Assumption: the layout and contents of the vDSO (and vvar) sections has not changed between checkpoint and restart.

- The addresses of vDSO (and vvar) sections are stored in ProcessInfo during initialization.
- During checkpoint, the contents of vDSO, vvar, and vsyscall sections are not written to the checkpoint image.
- The vDSO (and vvar) sections are checked for layout compatibility at restart.
- If the layout is correct, the new vDSO (and vvar) section(s) are mremap()'d to the older addresses. (mremap just updates the page tables to establish a new virtual page to physical page mapping, thus the vDSO (and vvar) pages still point to the backing kernel pages).

TODO: If the vDSO (and vvar) section(s) have a different layout (e.g., restarting on a different kernel), one potential fix is to keep the older vDSO section, but modify the _vdso_gettimeofday, etc., to point into the new vDSO. This can be done by a simple `jmp` instruction on x86.

### DIFF
--- a/src/mtcp/mtcp_header.h
+++ b/src/mtcp/mtcp_header.h
@@ -44,6 +44,10 @@ typedef union _MtcpHeader {
     void *saved_brk;
     void *restore_addr;
     size_t restore_size;
+    void *vdsoStart;
+    void *vdsoEnd;
+    void *vvarStart;
+    void *vvarEnd;
     void (*post_restart) ();
     ThreadTLSInfo motherofall_tls_info;
     int tls_pid_offset;
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -581,6 +581,7 @@ static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
       // Do not unmap vdso.
       vdsoStart = area.addr;
       vdsoEnd = area.endAddr;
+      DPRINTF("***INFO: vDSO found (%p-%p)\n", area.addr, area.endAddr);
     } else if (mtcp_strcmp(area.name, "[vvar]") == 0) {
       // Do not unmap vvar.
       vvarStart = area.addr;
@@ -588,12 +589,12 @@ static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
     } else if (mtcp_strcmp(area.name, "[vsyscall]") == 0) {
       // Do not unmap vsyscall.
     } else if (area.size > 0 ) {
+      DPRINTF("***INFO: munmapping (%p-%p)\n", area.addr, area.endAddr);
       if (mtcp_sys_munmap(area.addr, area.size) == -1) {
         MTCP_PRINTF("***WARNING: munmap(%p, %d) failed: %d\n",
                     area.addr, area.size, mtcp_sys_errno);
         mtcp_abort();
       }
-      MTCP_PRINTF("***INFO: munmap(%p, %d) worked\n", area.addr, area.size);
       // Rewind and reread maps.
       mtcp_sys_lseek(mapsfd, 0, SEEK_SET);
     }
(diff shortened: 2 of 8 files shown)


gc00:
On dekaksi, this is failing for me with   Error: failed to mremap vvar
(Details below):

gene@dekaksi:/tmp/gene/tmp/dmtcp-master$ bin/dmtcp_launch -i5 test/dmtcp1
dmtcp_coordinator starting...
    Host: dekaksi (155.33.213.218)
    Port: 7779
    Checkpoint Interval: 5
    Exit on last client: 1
Backgrounding...
  1   2   3   4 ^C
gene@dekaksi:/tmp/gene/tmp/dmtcp-master$ bin/dmtcp_restart ckpt_dmtcp1_1d4a852a5f139a6-40000-550731f2.dmtcp 
dmtcp_coordinator starting...
    Host: dekaksi (155.33.213.218)
    Port: 7779
    Checkpoint Interval: disabled (checkpoint manually instead)
    Exit on last client: 1
Backgrounding...
[20010] mtcp_restart.c:596 unmap_memory_areas_and_restore_vdso:
  ***INFO: munmap(0x400000, 45056) worked
[20010] mtcp_restart.c:596 unmap_memory_areas_and_restore_vdso:
  ***INFO: munmap(0x60a000, 8192) worked
[20010] mtcp_restart.c:596 unmap_memory_areas_and_restore_vdso:
  ***INFO: munmap(0x60c000, 8192) worked
[20010] mtcp_restart.c:596 unmap_memory_areas_and_restore_vdso:
  ***INFO: munmap(0x7ffffffde000, 135168) worked
[20010] mtcp_restart.c:636 unmap_memory_areas_and_restore_vdso:
  ***Error: failed to mremap vvar.
Segmentation fault (core dumped)

gc00:
It failed for me on dekaksi.  I did a small diagnosis via:
  cd src/mtcp; make -f Makefile.debug gdb
It shows that when mtcp_restart reads the restore_info header,
it's not getting the full information:
  vdsoStart and vdsoEnd are good, but
    I see these fields:  vvarStart = 0x0, vvarEnd = 0x0,
and so mremap fails when given those arguments.

- Gene

karya0:
Thanks for the catch. Just updated the pull request with the fix.

gc00:
Hi Kapil,
    I appreciate the new implementation.  But there is also something that
I worry about.
    In the one earlier commit, you "cleaned up VDSO" and got rid
of a lot of the code that MTCP had used to check various corner
cases over the years.
    Then in the most recent commit, you added something like:
"if vvar not found, then call the old-style code" (most of which
no longer exists now).

    It seems to me that this will cause many users who are not
using vvar to discover regressions of old bugs.  Can we bring back
the old code from MTCP (with a comment to get rid of it when vvar
is standard for most users)?  This seems a lot safer.
    In particular, some of that code we got rid of is also used
in special cases in supporting ARM-32 and ARM-64.  At this time,
Linux ARM does not appear to be using vvar at all.

Thanks,
- Gene

karya0:
Two things:
1. The code that we used to check various corner cases over the years was mostly duct-tape.  We started by ignoring vsyscall, then vdso, then stack/vdso ordering, and now finally vvar.  Further, a bunch of that code was leftover from when we had libmtcp.so that we used to load from the ckpt image.  In the current setup, all of that was irrelevant. I am quite certain that we haven't lost any functionality or compatibility with older kernels/glibcs.

2. I am not sure where the "if vvar not found, then call the old-style code" is coming from.  The bug that I had in the implementation was that it didn't check whether vvar was available or not, before doing a mremap.  The fix was to do a mremap only if vvar exists.

Having said that, the current code doesn't assume that `vvar` is always present. It does vvar related stuff (mremap on restart) only if it finds `vvar`, and skips it otherwise.

Similarly, on even older systems, if `vdso` is missing, the current code is harmless and doesn't have any undesired side-effects.

The current patch might still have some silly bugs that can be detected by a careful code-review. However, I do not suspect any hard bugs to show up (except the ones already mentioned in the description).

karya0:
The next thing to figure out now is to see what to do with `vectors` section for ARM machines. @rohgarg: Can you take a look at it and see how should the code be altered to handle vectors in the new design?

gc00:
Hi Kapil,
    Sure there's a lot of duct tape in mtcp_restart.  What's making
me nervous is to tear away the duct tape on the day before a release.
We're already late on this release.  Wouldn't we want to do some testing
after we tear away the duct tape and before a release?

    Anyway, I propose that we release rc1 as is, and then merge
this pull request into 'master', for rc2.
    Everything's ready to go, if you would like to do the rc1 release.
If not, please let me know soon, and I'll do it.

    It makes me nervous that our support for matlab has been failing
for more than a year.  And we promised the MVAPICH people that we
would have a release supporting MVAPICH, since their release
is pointing toward DMTCP.

Thanks,
- Gene

karya0:
I would suggest the we release rc1 with this patch. People downloading rc1 would know that there might be bugs. And in the meanwhile, let's try to build on all known machines that we have access to and fix any bugs that might have seeped in.

karya0:
The trouble with release rc1 without this patch is that DMTCP would be broken on most distros that use Linux 3.16 (released in August 2014) or newer.

gc00:
Hi Kapil,
My preferred code for rc1 would be:
  if vvar found:
    do it the right way
  else:
    keep using the old duct tape

What I'm objecting to is to tear out the duct tape
the day before the release.  What is the downside of
using the above pseudo-code for rc1, and tearing out the
duct tape the day _after_ the release?

Best,
- Gene

----- Original Message -----

karya0:
I don't think there is an easy solution for just vvar. It goes hand in hand with vDSO. So, let's just tag rc1 _without_ this patch and then merge this into master followed by rc2 by the end of the week after some more testing.

gc00:
Agreed.  Thanks, Kapil.  that's a good solution.
- Gene

----- Original Message -----

rohgarg:
Good news! This patch works for me (OpenSuse-13.2, Linux-3.16.7), and I'm no longer seeing the kernel hang-up. make check passes for all the test.

gc00:
Hi Kapil,
    When you're ready, let's do the pull for this.  I didn't see any other issues in my review, and we'll have a few days for testing now before rc2.  Thanks.

karya0:
Committed.

PR #42, opened 2015-03-21
Fix compare of mtcpHdr with MTCP_SIGNATURE

gc00:
Previously on fail, it would read beyond the MTCP header, looking for  MTCP_SIGNATURE.  Now, it reads the header only, and does not read further.

MINOR:  Also, adding a FIXME comment for case when can't restore MAP_SHARED file.
MINOR: Fixed a typo in comment

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -239,9 +239,11 @@ MTCP_PRINTF("Attach for debugging.");
       mtcp_abort();
     }
     // This assumes that the MTCP header signature is unique.
-    do {
-      mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
-    } while (mtcp_strcmp(mtcpHdr.signature, MTCP_SIGNATURE) != 0);
+    mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
+    if (mtcp_strcmp(mtcpHdr.signature, MTCP_SIGNATURE) != 0) {
+      MTCP_PRINTF("***ERROR: ckpt image doesn't match MTCP_SIGNATURE\n");
+      mtcp_abort();
+    }
   }
 
   DPRINTF("For debugging:\n"
@@ -826,7 +828,7 @@ static int read_one_memory_area(int fd)
     }
 
     if (area.prot & MAP_SHARED) {
-      imagefd = mtcp_sys_open (area.name, flags, 0);  // Can we open file.?
+      imagefd = mtcp_sys_open (area.name, flags, 0);  // Can we open file?
       if (imagefd < 0 && mtcp_sys_errno == ENOENT) {
         // File doesn't exist.  Do we have perm to create it and write data?
         imagefd = mtcp_sys_open (area.name, O_CREAT|O_RDWR, 0);
@@ -835,6 +837,8 @@ static int read_one_memory_area(int fd)
         } else {
           // We don't have permission to re-create shared file.
           // Open it as anonymous private, and hope for the best.
+          // FIXME:  Or maybe we do have permission to re-create shared file,
+          //         but it requires us to also re-create the parent directory.
           area.flags ^= MAP_SHARED;
           area.flags |= MAP_PRIVATE;
           area.flags |= MAP_ANONYMOUS;


src/mtcp/mtcp_restart.c
```diff
@@ -239,9 +239,11 @@ MTCP_PRINTF("Attach for debugging.");
       mtcp_abort();
     }
     // This assumes that the MTCP header signature is unique.
-    do {
-      mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
-    } while (mtcp_strcmp(mtcpHdr.signature, MTCP_SIGNATURE) != 0);
+    mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
```
karya0: I am not sure if I understand this patch.  This 'else' branch is taken only by readmtcp in which case, there might still be DMTCP data (ProcessInfo for parent-child maps, etc.) that needs to be read and ignored. With this patch, readmtcp would always fail because the first page of the ckpt image is always going to be related to the ProcessInfo metadata.

I do understand the problem you are trying to address though. We should instead check the return value of mtcp_readfile and abort if it's unable to read `sizeof mtcpHdr` data.

gc00: Hi Kapil,
    I don't understand your modification of my patch.
    In the current code, mtcp_readfile() either succeeds in reading the
header, or it calls mtcp_abort().  The existing do/while loop says
that if we succeeded in reading the header (e.g., 4096 bytes) and if it
does not match MTCP_SIGNATURE, then call mtcp_readfile() another time
and read the next 4096 bytes and see if you can find an MTCP_SIGNATURE
there.  After repeatedly failing to find MTCP_SIGNATURE,
mtcp_readfile() is called when the fd is at the end of the ckpt image,
and it reports the obscure error mesage
" failed to read after 10 tries in a row.\n", and then aborts.  :-)
    I patched the original behavior to call mtcp_readfile() only
once, and then abort with a relevant error message if the MTCP_SIGNATURE
doesn't match.  In your newest patch, you seem to be proposing that if
it doesn't match with MTCP_SIGNATURE, then don't abort at all, and try
to process it as an MTCP header even though it's not one.  :-(

Best,
- Gene

On Sat, Mar 21, 2015 at 09:02:47AM -0700, Kapil Arya wrote:
> > @@ -239,9 +239,11 @@ MTCP_PRINTF("Attach for debugging.");
> >        mtcp_abort();
> >      }
> >      // This assumes that the MTCP header signature is unique.
> > -    do {
> > -      mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
> > -    } while (mtcp_strcmp(mtcpHdr.signature, MTCP_SIGNATURE) != 0);
> > +    mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
[ Above line is Kapil's proposed alternative patch.  - Gene ]

gc00: I was writing too fast, while distracted earlier.  There was only one
patch (mine).  In any case,  I hope my explanation was clear.

karya0: The problem with calling `mtcp_readfile` only _once_ is that it would always fail to find the `MTCP_SIGNATURE` because the first 4096 bytes (could be more) represent DMTCP metadata (i.e. ProcessInfo). Here is a test run of this patch:

```
## readdmtcp.sh on a compressed ckpt image:
dekaksi:~/dmtcp> util/readdmtcp.sh ckpt_sleep_1d4a852a5f139a6-40000-5511c7a1.dmtcp 
*** ckpt_sleep_1d4a852a5f139a6-40000-5511c7a1.dmtcp is a gzipped file. Will uncompress it into ckpt_tmp.dmcp first.
[20953] mtcp_restart.c:202 main:
  Considering 'ckpt_tmp.dmtcp' as a ckpt image.
[20953] mtcp_restart.c:244 main:
  ***ERROR: ckpt image doesn't match MTCP_SIGNATURE
Segmentation fault (core dumped)

## Uncompress the ckpt image:
dekaksi:~/dmtcp> mv ckpt_sleep_1d4a852a5f139a6-40000-5511c7a1.dmtcp{,.gz}
‘ckpt_sleep_1d4a852a5f139a6-40000-5511c7a1.dmtcp’ -> ‘ckpt_sleep_1d4a852a5f139a6-40000-5511c7a1.dmtcp.gz’
dekaksi:~/dmtcp> gunzip ckpt_sleep_1d4a852a5f139a6-40000-5511c7a1.dmtcp.gz 

## readmtcp.sh on uncompressed image
dekaksi:~/dmtcp> util/readdmtcp.sh ckpt_sleep_1d4a852a5f139a6-40000-5511c7a1.dmtcp
[20992] mtcp_restart.c:202 main:
  Considering 'ckpt_sleep_1d4a852a5f139a6-40000-5511c7a1.dmtcp' as a ckpt image.
[20992] mtcp_restart.c:244 main:
  ***ERROR: ckpt image doesn't match MTCP_SIGNATURE
Segmentation fault (core dumped)
```

gc00: This is now made obsolete by Issue #46, where I try again.  We can close this issue.

PR #46, opened 2015-03-25
mtcp_restart exits if no match, MTCP_SIGNATURE

gc00:
* It used to verbosely make 20 tries at end of file, and then abort.
  Now, it prints a simple error statement, and exits with 'return 1'.

I've tested this on:  bin/mtcp_restart --simulate IMG
for the cases when IMG is a gzipped ckpt image, an unzipped ckpt image, and a random file still ending in .dmtcp.  It seems to do the right thing in all cases, and without the previous bug of repeatedly printing:
  " failed to read after 10 tries in a row.\n"

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -235,6 +235,7 @@ MTCP_PRINTF("Attach for debugging.");
   if (rinfo.fd != -1) {
     mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
   } else {
+    int rc = -1;
     rinfo.fd = mtcp_sys_open2(ckptImage, O_RDONLY);
     if (rinfo.fd == -1) {
       MTCP_PRINTF("***ERROR opening ckpt image (%s): %d\n",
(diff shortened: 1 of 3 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -243,8 +244,12 @@ MTCP_PRINTF("Attach for debugging.");
     }
     // This assumes that the MTCP header signature is unique.
     do {
-      mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
-    } while (mtcp_strcmp(mtcpHdr.signature, MTCP_SIGNATURE) != 0);
+      rc = mtcp_readfile(rinfo.fd, &mtcpHdr, sizeof mtcpHdr);
+    } while (rc > 0 && mtcp_strcmp(mtcpHdr.signature, MTCP_SIGNATURE) != 0);
+    if (rc == 0) { /* if end of file */
+      MTCP_PRINTF("***ERROR: ckpt image doesn't match MTCP_SIGNATURE\n");
```
karya0: I wonder if we should rename `MTCP_SIGNATURE` with `DMTCP_SIGNATURE` in the error message so as to not confuse the user with MTCP/DMTCP.

We should keep this in mind for other such error strings as well.

karya0:
LGTM. You can push it to dmtcp/dmtcp.

Can you also create an issue on github to check the return values of `mtcp_readfile` everywhere else?

src/mtcp/mtcp_restart.c
gc00: > I wonder if we should rename `MTCP_SIGNATURE` with `DMTCP_SIGNATURE` in the error message so as to not confuse the user with MTCP/DMTCP.

It's a good point, but here we're referring to the name in the
code:  MTCP_SIGNATURE
In the code, we distinguish between the DMTCP header, which comes first,
followed by the MTCP header.  So MTCP_SIGNATURE reflects the code better,
and hopefully, users will recognize the resemblance in the names.

By the way, what does LGTH mean?  (I like systems, because there are
always new things to learn.)

Best,
- Gene

karya0:
Merged as 0d3b22b. 

PS: L[S]GTM: Looks [Sounds] Good To Me :-).

Issue #48, opened 2015-03-26
BUG diagnosis: User files saved in DMTCP_TMPDIR are being blacklisted.


gc00:
DMTCP_TMPDIR and TMPDIR can coincide.  A user will put files in TMPDIR, and DMTCP considers that all files in DMTCP_TMPDIR are owned by DMTCP.  This bug apparently affects any user who sets DMTCP_TMPDIR or who uses the --tmpdir flag of dmtcp_launch.  In particular, it affects test/file1, and it affects Open MPI version 1.6.

The following fails:
    DMTCP_TMPDIR=/tmp/dmtcp-gene@dekaksi make check-file1

This issue is that test/file1 creates and then unlinks a file while 
continuing to use the file descriptor pointing to it.

Apparently, _isBlacklistedFile() considers this _user_ file to be
blacklisted because it is in DMTCP_TMPDIR, and the program logic assumes
that all files in DMTCP_TMPDIR are owned by DMTCP.

Because the file is blacklisted, the file type is never set to FIKLE_DELETED.
So, on restart, the restart fails.

In order to demonstrate this diagnosis, try for example:
  DMTCP_TMPDIR=/tmp/dmtcp-USER@HOST make check-file1
and restart will fail.
Then go to: FileConnection::handleUnlinkedFile() and comment out the call to !_isBlacklistedFile(_path), and the bug will go away when again executing:
  DMTCP_TMPDIR=/tmp/dmtcp-USER@HOST make check-file1

I'll need to think carefully about a re-design of the file plugin to fix this.  Any suggestions are welcome.

gc00:
I'm thinking that DMTCP_TMPDIR should be the base directory, and DMTCP_TMPDIR/dmtcp-USER@HOST should be calculated by us internally.  So, the help in configure/dmtcp_launch would be modified, get_tmp_dir(), and the rm plugin uses DMTCP_TMPDIR directly.  We should also check what happens if DMTCP_TMPDIR is not present upon launch, upon restart, etc.  I'll go forward with this, if I don't hear anything.

karya0:
DMTCP_TMPDIR should _NOT_ be the base directory. The base directory is supposed to be TMPDIR.  DMTCP_TMPDIR is just a convenience variable to provide an alternate to TMPDIR/dmtcp-user@host.

[...]

PR #50, opened 2015-03-28
 DMTCP_TMPDIR is base dir only, dmtcp-USER@HOST added internally [CORRECTED]

gc00:
This pull request has two commits.  They correspond to the first two items in Issue #48.
1.  Previously, even if the user does not set DMTCP_TMPDIR, DMTCP would set it anyway.  DMTCP should not be creating additional environment variables visible to the user program.  The extra environment variable has now been removed.
2.  DMTCP_TMPDIR (and the value for the --tmpdir flag) are now considered base directories only.  Internally, DMTCP appends "dmtcp-USER@HOST" as a subdirectory to the base directory in order to create the true tmpdir.  This is hidden, and not documented to the user.  The user should not be writing files here.


src/dmtcp_coordinator.cpp
```diff
@@ -1796,7 +1799,7 @@ int main ( int argc, char** argv )
       setenv(ENV_VAR_CHECKPOINT_DIR, argv[1], 1);
       shift; shift;
     }else if(argc>1 && (s == "-t" || s == "--tmpdir")){
-      setenv(ENV_VAR_TMPDIR, argv[1], 1);
+      tmpDir = Util::calcTmpDir(argv[1]);
```
karya0: We need to use the `tmpdir_arg` trick here as well (just like dmtcp_restart and dmtcp_launch below).

gc00: Do we need the tmpdir_arg trick?  This code is inside:  dmtcp_coordinator.cpp:main().
Because it's invoked at the top level function, it should be enough to set the local variable tmpDir, and then pass that value into other functions.

karya0: If '--tmpdir' argument is not passed in then, the value of `tmpdir` would be "" and that would be a problem :-). Bottom line is that we need to call calcTmpDir() even if no --tmpdir flag was specified.

karya0:
PS: I am still not quite okay with the help string. Will try to come up with a suggestion by tomorrow, if I can.

src/dmtcp_coordinator.cpp
gc00: I fixed this now, but I don't know if github saved my comment.  Thanks for the catch.

gc00:
I've fixed the --tmpdir issue ("tmpdir_arg trick").  I've also now updated the help on it in dmtcp_restart.cpp, to more accurately reflect what's happening.

karya0:
The code looks good. However, I have a concern about the following sentence in the help message:
 "Can preempt previous tmpdir of coord or earlier launch.\n"

Apparently, we need to have the same tmpdir for all processes on each individual host because the pid plugin and others rely on files in tmpdir for publish/subscribe.  So if there are two processes A and B with two different tmpdirs tmpA and tmpB respectively, they won't be able to share the pids or participate in leader election for files, etc.

Thus I think we should make it clear in the help message that the tmpdirs for dmtcp_launch/dmtcp_restart SHOULD be the same for all processes on a given node.

gc00:
Good point.  I'll change the help message to:
  "      Can preempt previous tmpdir of coord.  (Behavior is undefined\n"
  "        if two launched processes specify different tmpdirs.)\n"
and then check it in.

karya0:
Do we need to say anything about coord tmpdir in the help messages for dmtcp_launch and restart? I am just thinking whether or not it is going to be confusing for the user.

gc00:
Suppose I also put the "coord tmpdir" inside the parentheses:
"        Directory to store temporary files (default: $TMDPIR or /tmp)\n"
"        (Can preempt previous tmpdir of coord.  Behavior is undefined\n"
"         if two launched processes specify different tmpdirs.)\n"
But if you still prefer removing the sentence "Can preempt ...", tell me,
and I'm find with removing it.

karya0:
What if we never specified a --tmpdir for the coordinator?

Alternatively, what if a --tmpdir for the coordinator but we do not specify a --tmpdir for launch/restart. In this case, doesn't this somewhat gives the wrong message that coord tmpdir is inherited by launch/restart?

This is less apparent in single node computations, but in distributed computations, this might be a bit confusing. 

Having said all this, what if we change the wording to something like:
" Directory to store temporary files for dmtcp_launch ONLY (default: ...)\n
  Behavior is undefined if two ...."

karya0:
I didn't read the whole message earlier. Yes, I would like to remove the "Can preempt .." sentence :-).

gc00:
So, for now, I'll remove the "Can preempt ..." sentence.  The sentence:
   (Behavior is undefined if two launched processes specify different tmpdirs.)\n"
will remain.
Are there any other changes we should make before I commit this?

Best,
- Gene

karya0:
Sounds good. Let's commit it.

gc00:
This has now been pushed into origin/master at:
  c458e6a43dbc30167927798287e674b86bac2e90
  b4d3b1b29b55ea9e05651fe6a511194b8bf563ee

PR #52, opened 2015-04-03
In mtcp, fix restoring of deleted directory tree (and add new tests)

gc00:
If an application like Open MPI created a directory tree in a temp directory, and then unlinked its files and directories, DMTCP fails to restore.  (DMTCP does work correctly, if there is only a deleted file and no deleted directories.)  This addition will call fix_parent_directories() both for MAP_SHARED and for the standard files mapped private.
This addition also creates shared-memory2.c and file2.c to test the code.  Unfortunately, the file2.c test exposes a corresponding bug in fileconnection.cpp.  So, file2 has not yet been added to autotest.py.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -112,6 +112,7 @@ static void mmapfile(int fd, void *buf, size_t size, int prot, int flags);
 #endif
 static void read_shared_memory_area_from_file(int fd, Area* area, int flags);
 static void lock_file(int fd, char* name, short l_type);
+static char* fix_parent_directories(char* filename);
 // static char* fix_filename_if_new_cwd(char* filename);
 static int open_shared_file(char* filename);
 static void adjust_for_smaller_file_size(Area *area, int fd);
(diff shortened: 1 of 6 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -864,6 +865,10 @@ static int read_one_memory_area(int fd)
         imagefd = mtcp_sys_open (area.name, O_CREAT|O_RDWR, 0);
         if (imagefd >= 0) {
           mtcp_sys_unlink(area.name);
+        } else if (fix_parent_directories(area.name) == area.name) {
```
karya0: Shouldn't we proactively create the directory tree, i.e. why not call fix_parent_directories() before we do mtcp_sys_open for the shared file? That way we can retain the `} else {` logic here.

With this patch, the `else` branch is dead code. `fix_parent_directories` will always return `area.name` or it will abort, in both cases the `else` branch is never taken.

src/mtcp/mtcp_restart.c
```diff
@@ -970,14 +975,20 @@ static void read_shared_memory_area_from_file(int fd, Area* area, int flags)
 
   /* Check to see if the filename ends with " (deleted)" */
   if (mtcp_strendswith(area_name, DELETED_FILE_SUFFIX)) {
+    // FIXME:  Shouldn't we also unlink area_name after re-mapping it?
     area_name[ mtcp_strlen(area_name) - mtcp_strlen(DELETED_FILE_SUFFIX) ] =
       '\0';
+    if (fix_parent_directories(area_name) != area_name) {
```
karya0: Since we are already calling `fix_parent_directories` above, why do we need it here as well?

src/mtcp/mtcp_restart.c
```diff
@@ -1262,10 +1273,32 @@ static void lock_file(int fd, char* name, short l_type)
...
+        MTCP_PRINTF("error %d re-creating directory %s\n",
+		    mtcp_sys_errno, filename);
+        filename[i] = '/';  // Restore original filename
+	mtcp_abort();
+      } // else parent directory already exists or was re-created
+      // FIXME:  We should unlink parent directories if we re-created them.
+      filename[i] = '/';  // Restore original filename
+    }
+  }
+  return filename;  // Success.  All parent directories now exist.
```
karya0: Since it always succeeds or aborts(), why not make it a `void` function?

src/mtcp/mtcp_restart.c
```diff
@@ -1262,10 +1273,32 @@ static void lock_file(int fd, char* name, short l_type)
...
+}
+
 #if 0
 /* Adjust for new pwd, and recreate any missing subdirectories. */
 static char* fix_filename_if_new_cwd(char* filename)
 {
+  int mtcp_sys_errno;
```
karya0: Github doesn't allow adding comment on unmodified lines so adding it here:

Lower in the function (lines 1314-1330), we have the logic to recreate directory tree. Why not replace this with the new function that you introduced?

src/mtcp/mtcp_restart.c
gc00: I don't want to put it ahead of mtcp_sys_open().  Primarily, I was thinking about efficiency.  fix_parent_directories is not necessarily cheap, since it starts at the root directory and tries to add each subdirectory in turn, testing for EEXIST.  Hence, I call it only if mtcp_sys_open() fails.

However, the point about creating dead code is excellent.  Nice catch!  I'll change the logic so that after calling fix_parent_directories(), I'll again try to open the file, and if I still fail, then I'll follow the existing "else" condition and change from MAP_SHARED to MAP_PRIVATE.

src/mtcp/mtcp_restart.c
gc00: Good point.  I'll remove this second call to fix_parent_directories() and add comments about the logic, for clarification.  You're two-for-two so far on catching my mistakes.  Thanks!

And in case we have to review this again, here was the previous logic:
  if MAP_SHARED:
    if ENOENT (open failed):
      if we can successfully fix parent directories:
        Go on and call read_shared_memory_area_from_file (in code below)
     else change MAP_SHARED to MAP_PRIVATE, and try again from the top:  goto read_data
   Still inside MAP_SHARED, read_shared_memory_area_from_file():
     fix_parent_directories (but it's redundant; we guaranteed the file does exist before going here)

src/mtcp/mtcp_restart.c
gc00: I'm primarily thinking of future extensions.  In one earlier variant, I had fix_parent_directories() returning NULL on failure, so that the parent function could do something further.  In a previous commit of mine to mtcp_restart.c, I had to do that treatment for mtcp_readfile(), changing it from succeed-or-abort to succeed-or-pass-error-to caller.  So, let's keep the potentially more general logic here.

src/mtcp/mtcp_restart.c
gc00: If we do change this code, it should be a separate commit, of course.  For now, I propose we simply add a FIXME comment for this fnc:  /* Could be re-written to use fix_parent_directories in both parts of the function */
The code is essentially, try to recreate old directories;  If we fail, then chop off the prefix for which there really is such a directory, and treat the rest as a relative pathname.  Then try to create the relative pathname.  Both the first half and the second half (with a relative pathname) could be made to use fix_parent_directories().  Note, by the way, in that case fix_parent_directories() really should be revised to return NULL on failure, rather than abort on failure.

gc00:
Nice catches!  The first two comments by @karya0 found a bug and a simplification of the code.  This latest pull request adopts two fixes for that.  I left "line notes" for the outdated pull request commenting in detail.

karya0:
Please update the pull request after doing a proper rebase.

gc00:
I've updated the pull request now.

gc00:
I updated the pull request primarily for mtcp_restart.c.  The call to fix_parent_directories() now comes one level earlier.  So, it applies to files that are mmap'ed as MAP_SHARED or MAP_PRIVATE if they are not MAP_ANONYMOUS.
There is still a test/file2.c that is not handled by DMTCP.  This test is for the case when an open file descriptor references a deleted directory tree, and there is no mmap'ed file.  In this case, mtcp can't handle it, since it only handles mmap'ed files.  So, the file plugin will have to be extended with the same logic as in mtcp_restart.c:fix_parent_directories().

src/mtcp/mtcp_restart.c
```diff
@@ -858,42 +859,54 @@ static int read_one_memory_area(int fd)
...
-            MTCP_PRINTF("Found old underlying file %s\n"
-                        " that no longer exists, but it's mapped shared,"
-                        " with write permission,\n and we don't have permission"
-                        " to re-create it.  We will map it read-only\n"
-                        " and hope for the best.\n", area.name);
-          }
-          goto read_data;
-        }
-      }
+    int can_create_area = 1;
```
karya0: I hate to do this to you, but could you explain why we moved the directory creation outside of `if (area.prot & MAP_SHARED) { }`? 

AFAIK, the directory tree matters only for the shared memory areas, so it's better to keep it all inside the 'if' block.

gc00: I can go either way.  But my thinking at the time was that the case of a deleted directory tree can occur both for a file mapped MAP_SHARED and for a file mapped MAP_PRIVATE.  Even though we don't see so many examples of a process mapping files into memory as MAP_PRIVATE from inside an unlinked directory tree, I figured we might as well also handle that case by moving the code outside the  `if (area.prot & MAP_SHARED) { }`.

karya0: For the MAP_PRIVATE case, we don't need to create the directory tree. If a file is missing, it will automatically be updated to MAP_ANONYMOUS and restart will proceed as expected. Recreating the directory tree in this case doesn't help. In fact, it's less optimal because we are unnecessarily creating the directory tree, recreating the file and populating it's contents from the file. However, once the file has been created, it won't be used again by the restarted process.

gc00:
I've now updated the pull request inline with the earlier comment.  It is now inside the `MAP_SHARED` if statement again.  I also tested against 'make check-shared-memory2' to guard against regressions.

karya0:
Thanks @gc00, let's commit it.

PR #55, opened 2015-04-10
mtcp_restart: print useful debugging info if munmap fails

rohgarg:


rohgarg:
@karya0: Let me know what you think. I think this will be useful for the future.

karya0:
:+1:  This would have saved us some time in the latest dmtcp-forum issue :-). Let's get it committed!!

PR #58, opened 2015-04-16
Add git-bisect.sh to util

gc00:
This creates a file that, with slight modifications, can be used for the git bisect command:
   `git bisect run git-bisect.sh`
At the beginning of the file, git-bisect.sh has comments on how to use this file.

### DIFF
--- a/util/README
+++ b/util/README
@@ -1,4 +1,18 @@
 This directory contains some useful scripts to work with DMTCP on clusters.
+This directory is becoming somewhat crowded, but some particularly useful
+utilities to highlight are:
+
+* readdmtcp.sh - read the memory map of a DMTCP checkpoint image (ckpt_*.dmtcp)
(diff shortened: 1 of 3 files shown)


karya0:
@gc00: Looks like your branch isn't in sync with dmtcp/dmtcp:master. Could you fix that and update the PR?

gc00:
It's now in sync again, and the pull request is updated.

gc00:
I've now updated the template for git-bisect.sh to also include git-bisect-m32.sh.  In both cases, the template now starts the search from tag 2.0, tests against dmtcp1 only, and assumes that "good" means everything passes, and "bad" means that we got at least as far as autotest.py, which printed "ckpt:".  Any other result is considered a "skip" for "git bisect.

karya0:
Merged via 15c312e36d843823c0072573658a522e3722d490.

Issue #64, opened 2015-04-22
JTRACE/JNOTE doesn't print for SLURM/Intel MPI-4.0/PMI


jiajuncao:
At the U. Buffalo HPC supercomputing center, we use --enable-debug.  The jassertlog is created for ckpt/resume, but not after restart.  Also, JNOTE does not print to stderr after restart.  This is a minor bug, which can be diagnosed and fixed later.

gc00:
This might be fixed by pull request #67.  We should test using the code from that pull request.

gc00:
@jiajuncao:  Pull request #67 has now been pushed into master.  This pull request might fix the issue noted here, but we won't be sure until it's tested.  Could you do this whenever you have the free time?  (It can even be done after the 2.4.0 release if you're busy now.)

Issue #63, opened 2015-04-19
Jalib shouldn't use dmtcp plugin API


karya0:
I just noticed that some part of Jalib is now using the plugin API. This shouldn't be the case. Whatever is needed by jalib should be supplied via `jalib_init`. The whole point of `jalib_init` was to eliminate two-way communication between jalib and rest of dmtcp.

PR #67, opened 2015-04-25
Jalib cleanup

karya0:
This is an attempt to remove Jalib's dependency on dmtcp.h to get closer to the ideal world where jalib is oblivious to DMTCP.

Addresses #63.

### DIFF
--- a/jalib/jalib.cpp
+++ b/jalib/jalib.cpp
@@ -34,28 +34,55 @@
 #include "jalib.h"
 #include "jassert.h"
 
-jalib::JalibFuncPtrs jalib::jalibFuncPtrs;
-int jalib::jalib_funcptrs_initialized = 0;
-const char *jalib::elfInterpreter = NULL;
-int jalib::stderrFd = -1;
-int jalib::logFd = -1;
-int jalib::dmtcp_fail_rc = -1;
+static jalib::JalibFuncPtrs jalibFuncPtrs;
+static int jalib_funcptrs_initialized = 0;
+
+static struct {
+  const char *elfInterpreter;
+  int stderrFd;
+  int logFd;
+  int dmtcp_fail_rc;
+} dmtcpInfo = {NULL, -1, -1, -1};
+
 extern "C" void initializeJalib();
 
-extern "C" void jalib_init(jalib::JalibFuncPtrs jalibFuncPtrs,
-                      const char *elfInterpreter,
-                      int stderrFd,
-                      int jassertLogFd,
-                      int dmtcp_fail_rc)
+extern "C" void jalib_init(jalib::JalibFuncPtrs _jalibFuncPtrs,
+                           const char *elfInterpreter,
+                           int stderrFd,
+                           int jassertLogFd,
+                           int dmtcp_fail_rc)
 {
-  jalib::jalibFuncPtrs = jalibFuncPtrs;
-  jalib::elfInterpreter = elfInterpreter;
-  jalib::stderrFd = stderrFd;
-  jalib::logFd = jassertLogFd;
-  jalib::jalib_funcptrs_initialized = 1;
-  jalib::dmtcp_fail_rc = dmtcp_fail_rc;
+  dmtcpInfo.elfInterpreter = elfInterpreter;
+  dmtcpInfo.stderrFd = stderrFd;
+  dmtcpInfo.logFd = jassertLogFd;
+  dmtcpInfo.dmtcp_fail_rc = dmtcp_fail_rc;
+
+  jalibFuncPtrs = _jalibFuncPtrs;
+  jalib_funcptrs_initialized = 1;
+
   JASSERT_INIT();
 }
+
+const char *jalib::elfInterpreter()
+{
+  return dmtcpInfo.elfInterpreter;
+}
+
+int jalib::stderrFd()
+{
+  return dmtcpInfo.stderrFd;
+}
+
+int jalib::logFd()
+{
+  return dmtcpInfo.logFd;
+}
+
+int jalib::dmtcp_fail_rc()
+{
+  return dmtcpInfo.dmtcp_fail_rc;
+}
+
 
 #define REAL_FUNC_PASSTHROUGH(type,name) \
   if (!jalib_funcptrs_initialized) { \
--- a/jalib/jassert.cpp
+++ b/jalib/jassert.cpp
@@ -259,15 +259,11 @@ void jassert_internal::jassert_safe_print(const char* str, bool noConsoleOutput)
   if (theLogFileFd != -1) {
     int rv = jwrite(theLogFileFd, str);
 
-    if (rv < 0) {
+    if (rv < 0 && theLogFileFd == EBADF) {
       if (errConsoleFd != -1) {
-        jwrite(errConsoleFd, "JASSERT: write failed, reopening log file.\n");
+        jwrite(errConsoleFd, "JASSERT: failed to write to log file.\n");
       }
-      set_log_file(theLogFilePath());
-      if (theLogFileFd != -1) {
-        jwrite(theLogFileFd, "JASSERT: write failed, reopened log file:\n");
-        jwrite(theLogFileFd, str);
-      }
+      theLogFileFd = -1;
     }
   }
 }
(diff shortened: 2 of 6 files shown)


jalib/jassert.cpp
```diff
@@ -46,6 +43,9 @@ int jassert_quiet = 0;
 static int theLogFileFd = -1;
 static int errConsoleFd = -1;
 
+static jalib::string& tmpDir() {static jalib::string s;return s;};
+static jalib::string& uniquePidStr() {static jalib::string s;return s;};
```
gc00: I guess the idea here is that the function `tmpDir()` is a method in the jalib class (although to verify this, the reader has to go to the beginning of the file and look for 'namespace').  One could have used the better known paradigm of a field member, `jalib::_tmpdir`, with methods, `jalib::get_tmpdir()` and `jalib::set_tmpdir()`.  Here, we are using an alternative paradigm that is special to C++ only, and allows one to use `tmpDir()` both as a fixed return value (what everybody is familiar with in the standard paradigm), and also as a tricky way to hide a hidden field member (the local variable `static jalib::string s`) and then use the unusual syntex `tmpDir() = ...`.
However, since we will have many developers who are not trained in the intricacies of C++, I propose that we use the better known forms `get_tmpdir()` and `set_tmpdir()`.  Similarly, for `uniquePidStr()` and `theLogFilePath()`.

jalib/jalib.cpp
```diff
@@ -34,29 +34,56 @@
...
-int jalib::dmtcp_fail_rc = -1;
+static jalib::JalibFuncPtrs jalibFuncPtrs;
+static int jalib_funcptrs_initialized = 0;
+
+static struct {
+  const char *elfInterpreter;
+  int stderrFd;
+  int logFd;
+  int dmtcp_fail_rc;
+} dmtcpInfo = {NULL, -1, -1, -1};
```
gc00: Because dmtcpInfo occurs before the call to `namespace jalib`, it is part of the std namespace.  This is also the case for `jalib_funcptrs_initialized`.  However, `jalibFuncPtrs` is explicitly made a member of the `jalib` class.  All three are static, and so we can do anything we like, since they won't be exposed outside of this file.
However, I would prefer a common convention.  Either we should make everything part of `jalib` or none of this.  (Naturally, it's fine to break this convention when we have a good reason, but we should start with a standard convention.)

gc00:
I added two inline comments about details of the code.  But aside from that, the general concept looks good, and is ready to go.  I really appreciate separating out jalib from the dmtcp subdirectory.
I was curious about the history.  The connection between jalib and dmtcp crept in on March 13, 2013 (two years ago) at a time when we had other priorities.  It's nice to finally clean this up.

gc00:
It may be that this pull request also fixes issue #64 .  We should ask @jiajuncao to do a quick test of this version, and see if the bug goes away.

jalib/jassert.cpp
karya0: This whole library uses heavy C++ so I wouldn't worry about the intricacies.  C++ only paradigms aren't bad and I would _strongly_ suggest that we keep it the way it is. After all we are not talking about rewriting it all to use C++11, ;-).

The problem with creating a static variable is that it gets us into weird dependency problems during initialization. This particular invocation (static variable inside a function body) is a well known and accepted strategy to avoid polluting the global namespace and weird dependency chains.

karya0: Next, let's avoid mixing C and C++ styles. It's a bad engineering practice and our code base is already getting fairly ugly that way :-(.

jalib/jalib.cpp
karya0: That's a good point. I would suggest that we wrap everything inside `namespace jalib { ... }` and avoid using `using namespace jalib;`. This has an added benefit of keeping the global namespace clean and is the preferred way.

karya0: I would suggest that we do it for the entire code base. I will try to come up with a pull request soon.

jalib/jassert.cpp
gc00: I'm not proposing to mix C and C++ styles.  I just want to limit the code to a simple enough C++ style that new developers can understand the code without too much effort.  You wrote "This whole library uses heavy C++ so I wouldn't worry about the intricacies."  I found the rest of the code in the file very readable.
In this case, I'm speaking from my experience at working with many students, and observing what they do understand and what they do not understand.  If we ask students to add code to DMTCP using a style that they do not understand, we are asking for trouble.  The convention used here is not taught in standard C++ courses.  _Isn't there any common convention in C++ that would be understood by students who have had only C, Java, and perhaps one semester of C++?_

If this conversation goes further, let's talk on the phone rather than messaging.  But the issue that students face is that the statement `tmpDir() = ...` does not have any mention within the current function of what the destination type is.  Students know that C++ will do appropriate type conversions, but that's a hard concept to apply when the destination type is not specified.

gc00: I deleted some of the previous messages, which were distracting and due to my own misunderstanding of the technical issue involved here.
First, we agree that this usage of `foo() = "newValue";` with `foo()` containing a static local variable in cases where `foo()` would be declared static within the same file, and where `foo()` was just a function, and not a method inside some other class.
Second, a preferred option in an ideal world would be to promote the static local variable of `foo()` into a static global variable.  However, in C++, this would cause the global variable to be allocated, and any default constructor would be invoked, even before main() is called.  That creates problems if the constructor invokes any wrapper functions from DMTCP.  In C++, this initialization of the static local variable is deferred until the first call to `foo()`, and so it will now occur only after main() is called, and only after DMTCP has been initialized.

CONCLUSION:  Even though programmers new to C++ will initially be confused by seeing a function on the left-hand side, this is a necessary evil.  As long as we document the definition of `foo()` with a comment ahead of it, this will mostly avoid any confusion among programmers new to C++.

gc00: @karya0:  Here is one possible comment to add in front of the definition of `tmpDir()`.  Feel free to modify it, if you have something better.  Whatever comment we use will then be re-used in occurrences of this paradigm in other files.
`// This function is a "wrapper" around a static local variable.  We will use the function
 // on the left-hand side of an assignment statement.  We cannot promote the static local
 // variable to a static global variable and use the variable on the left-hand side, because
 // in C++, the default constructor would then be called even before main().  That default
 // constructor might call a DMTCP wrapper function, even before DMTCP is fully initialized.`

rohgarg: @gc00: Also, with multiple global (non-primitive) objects, the order in which their constructors are called is hard to reason about. It is not defined by the language specification. However, in this case, we want precise control over when the initialization of `tmpDir` happens. This is important because the DMTCP libraries, and the executables that statically link with `libjalib.a` can define their own global C++ objects. (In fact, libdmtcp relies on this trick to inject a ckpt thread in every process).

As @karya0 mentioned earlier, wrapping global C++ objects in functions is the idiomatic way of implementing this. This is portable and avoids non-deterministic behavior. See: https://isocpp.org/wiki/faq/ctors#static-init-order, and https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use.

gc00: This is another good point.  The comment that I suggested was just
a first draft.  Probably @kara0 and you can improve on it.  Maybe,
we could make the comment shorter by pointing to:
https://isocpp.org/wiki/faq/ctors#static-init-order,
and https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use
I'm good with however you choose to word the comment.

gc00:
Hi all. For this pull request, here's what we're waiting on:
1. A comment in front of where we declare a static global fnc with a static local variable for functions on left hand side of comments:  https://github.com/dmtcp/dmtcp/pull/67#discussion_r29300350 and https://github.com/dmtcp/dmtcp/pull/67#discussion_r29303859
2. Putting functions inside jalib namespace consistently:  https://github.com/dmtcp/dmtcp/pull/67#discussion_r29106963 and https://github.com/dmtcp/dmtcp/pull/67#discussion_r29106967
3. Asking @jiajuncao to do a test to see if this fixes issue #64 : https://github.com/dmtcp/dmtcp/pull/67#issuecomment-96323848  (we could postpone this by adding this comment to issue #64)

karya0:
Added an issue #82 to handle the namespace consistency. Merging the rest now.

karya0:
Merged.

PR #60, opened 2015-04-16
Fix random restart bug; seen in MVAPICH @ Stampede

gc00:
Proposal to solve bug in issue #56

PR #71, opened 2015-04-27
Fixes multi-arch bug from: dec2c26c1eb13eb1c12ed

gc00:
This fix finally make multi-arch work again.  (It last worked for dmtcp-2.4.0-rc1.)
This will also make --enable-m32 standalone work, when it is combined pull request #66 and also with a fix for dmtcp_restart.cpp. (I'm doing that fix as an independent pull request: #72.  It has similar logic to:  9445ba7ef682876ad09483e31c4c6e5f10556289 for dmtcp_launch.cpp.)  The final standalone m32 will also require `(mkdir -p bin && cd bin && ln -s ../lib/dmtcp/32/bin/* .)`

This bug has taken a long time to fix.  Here are two lessons for the future:
1.  Whenever we modify MTCP in a way that could affect assembly, or different versions of the kernel or libc for different CPUs, we need to first test on all of:  __x86_64__, __i386__, __arm__, __aarch64__
2.  We need to split commits into smaller units.  The commit that created this bug ( dec2c26c1eb13eb1c12ed ) was both changing the restart algorithm to use mremap, and tearing away the duct tape.  Analyzing what caused the bug required separating this commit into the two halves.


gc00:
As a separate comment, I'll note that this pull request requires ENABLE_VDSO_CHECK to be defined.  Eventually, we may be able to get rid of that by separately including inline certain functionality from mtcp_check_vdso().  Probably we'll have to worry about two issues:
`mtcp_sys_personality((pers | ADDR_NO_RANDOMIZE) & ~ADDR_COMPAT_LAYOUT)`
and making sure that we are using the `AT_SYSINFO` vector correctly.

src/mtcp/mtcp_restart.c
```diff
@@ -668,6 +669,24 @@ static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
...
+    //   code above was added, and then caused a segfault on restart for
+    //   __i386__ in CentOS 7.  In that case ENABLE_VDSO_CHECK was not defined.
+    //   This version was needed in that commit for __i386__
+    //   (i.e., for multi-arch.sh) to succeed.  But more recently,
+    //   the mremap version seems to succeed even without ENABLE_VDSO_CHECK. 
+    void *vdso = mtcp_sys_mmap(rinfo->vdsoStart,
+                                 vdsoEnd - vdsoStart,
+				 PROT_EXEC | PROT_WRITE | PROT_READ,
+                                 MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    mtcp_memcpy(rinfo->vdsoStart, vdsoStart, vdsoEnd - vdsoStart);
```
karya0: mmap+memcpy != mremap. This is wrong. What's wrong with using mremap here?

gc00: > mmap+memcpy != mremap. This is wrong. What's wrong with using mremap here?

What's wrong is that multi-arch fails when you use mremap.  Also,
--enable-m32 standalone fails when you use mremap.  We can spend some more
weeks trying to figure out why, but that means we're moving even further
away from the principle of "release often".  Let's release something that
works, and then do the further testing in parallel with our users running
the latest release.

karya0: I agree that we are moving further away from the principle of "release often", but releasing broken code is even worse :-). Let's at least document very clearly what's broken and at that point, I won't have any objections to a quick release.

gc00: Great.  Do you agree that I should put back the duct tape?  At least the
code is less broken with the duct tape.  Otherwise, we have to document
that we no longer support multi-arch, x86, arm, or arm-64.  :-(  (joke)

I can put back the duct tape tonight, and we can do an rc4.
Then we can use rc4 for testing against the various CPU architectures.

gc00: I deleted the last two comments, because our interactive conversation has now made things clear, and those comments would be a distraction.
I like the bug fix that you found.  Just in case you're short of time, I'm going to modify this pull request to reflect the bug fix that you described.  Let me know if I don't get it right.  This will give us something concrete on which we can do testing across architectures.

gc00:
I've now replaced the original pull request with a new one based on the comments of @karya0.  This one preserves the mremap, and for __i386__, it does an mmap of the new vdso address, and copies from the old vdso address (where the new vdso code resides) to the new vdso address.  This worked correctly for multi-arch on Ubuntu 13.10.
After this, I will test 32-bit standalone in a 64-bit Linux, and also I'll test on a 32-bit native Linux.

src/mtcp/mtcp_restart.c
```diff
@@ -668,6 +668,37 @@ static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
     }
     MTCP_ASSERT(vdso == rinfo->vdsoStart);
   }
+#if defined(__i386__)
```
karya0: The new code needs to go inside the previous `if  (vdsoStart != NULL) { ... }` block.

You also need to do the same i386 specific trick to the `if  (vvarStart != NULL) { ... }` block.

src/mtcp/mtcp_restart.c
```diff
@@ -668,6 +668,37 @@ static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
...
+    //   for kernel calls while using vdso for system calls that can be
+    //   executed purely in user space through kernel-specific user-space code.
+    // On older kernels like __x86__, both purposes are squeezed into vdso.
+    //   Since vdso will use randomized addresses (unlike the standard practice
+    //   for vsyscall), this implies that kernel calls on __x86__ can go through
+    //   randomized addresses, and so they need special treatment.
+    void *vdso = mtcp_sys_mmap(rinfo->vdsoStart,
+                                 vdsoEnd - vdsoStart,
+				 PROT_EXEC | PROT_WRITE | PROT_READ,
+                                 MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
```
karya0: We need to also add MAP_FIXED or else the kernel might disregard our addr request.

gc00:
Hi @karya0 and @rohgarg:
These are all great comments!  (How embarassing on my part -- especially to get the mmap/memcpy totally wrong.)

I think I've applied all the suggestions (except for `#if defined(ENABLE_CHECK_VDSO)`.  `#ifdef` seemed just as good.  But if you really prefer the other syntax, tell me, and I'll change it.

I just re-tested on the 32-bit CCIS machine.  Everything is passing, with these minor caveats for the shared-memory tests and emacs:

`shared-memory1 ckpt:PASSED rstr:FAILED (first process didn't die) retry:PASSED; ckpt:PASSED rstr:FAILED (first process rec'd signal 11) retry:PASSED`
`shared-memory2 ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:FAILED (first process didn't die) retry:PASSED`
`emacs          FAILED`
`               root-pids: [30077] msg: failed to write 's' to coordinator (pid: 20021)`
[ I'll look more closely at these three isolated cases later. ]

Later, I'll also test on 32-bit RedHat6 and RedHat5 at batlab.  I'll also re-test for multi-arch and standalone m32 (which are almost the same thing).  But at this point, it seems that we will be able to leave the duct tape off.

Well done!

src/mtcp/mtcp_restart.c
```diff
@@ -184,11 +184,11 @@ MTCP_PRINTF("Attach for debugging.");
...
 #endif
 
-// TODO(karya0): Remove vDSO checks after 2.4.0-rc3 release.
+// TODO(karya0): Remove vDSO checks after 2.4.0-rc3 release, and after testing.
 // Without mtcp_check_vdso, CentOS 7 fails on dmtcp3, dmtcp5, others.
 #define ENABLE_VDSO_CHECK
 // TODO(karya0): Remove this block and the corresponding file after sufficient
-// testing.
+// testing:  including testing for __i386__, __arm__ and __aarch64__
 #ifdef ENABLE_VDSO_CHECK
```
karya0: I meant to replace `#ifdef ENABLE_VDSO_CHECK` with `#if defined(__i386__)` because we dont want vDSO check for 64-bit.

gc00: But I thought we were keeping `#ifdef ENABLE_VDSO_CHECK` along with `#define ENABLE_VDSO_CHECK` because otherwise MVAPICH fails on Stampede (in 64-bit mode).  This was the essence of this comment:  https://github.com/dmtcp/dmtcp/pull/60#issuecomment-94137963 .  That pull request had been closed when we added  `#define ENABLE_VDSO_CHECK` to the master branch.

karya0: Is it possible for me to observe the 64-bit failure somewhere?

For the current timing, let's keep the code as is for 2.4.1, but open a new issue (targeted for 2.4.1) about replacing  `#ifdef ENABLE_VDSO_CHECK` with `#if defined(__i386__)`. I would like to figure out the problem and potentially fix it.

gc00: @karya0: That's a good solution (keeping the code as is for now, and fixing it for 2.4.1).  Both @jiajuncao and @rohgarg have accounts on Stampede, and know how to observe this bug.  They can also give you a sub-account on Stampede, with instructions on observing the bug.  I assume you're busy enough that we can wait until after the 2.4.0 release for that.

gc00:
I've done further testing on this patch now.  Recall that it mostly works on the CCIS 32-bit Ubuntu machine, with shared-memory1, shared-memory2, and emacs having problems.  For 32-bit RedHat5, all tests are passing.  For 32-bit RedHat6, sysv-shm2 and posix-mq1 are failing on the second restart only.

When I tested multi-arch just now, something was wrong upon dmtcp_launch, but I'm sure that can be fixed.  I'll look into it next.  dmtcp_launch is independent of the mtcp_restart issues addressed by this pull request.

CONCLUSION:  We have 32-bit DMTCP working now without the "duct tape".  :-)  Let me know when you're happy with the code, @karya0 and @rohgarg, and I'll push this into the master branch.

karya0:
LGTM. Let's merge it.

gc00:
Done.  :-)

gc00:
Done again!  It didn't work for me, the first time I did it.  (But I'm getting better.  This time I didn't need @karya0 to bail me out.  :-) )

PR #72, opened 2015-04-27
Fix standalone --enable-m32 for dmtcp_restart

gc00:
This is the final piece needed to allow standalone --enable-m32.
It is to be used in combination with pull requests #66 and #71,
and along with `(mkdir -p bin && cd bin && ln -s ../lib/dmtcp/32/bin/* ./)`

However, note that the following tests are continuing to fail under
--enable-m32:
  plugin-example-dbckpt, waitpid, perl, python, bash, ssh, vim, emacs
[COMMENT ADDED ON Apr. 30:  The last few tests concern 64-bit applications.  That's why 32-bit standalone mode is failing to checkpoint them.]

### DIFF
--- a/src/dmtcp_restart.cpp
+++ b/src/dmtcp_restart.cpp
@@ -199,6 +199,18 @@ class RestoreTarget
         string installDir =
           jalib::Filesystem::DirName(jalib::Filesystem::GetProgramDir());
 
+#if defined(__i386__)
+        if (Util::strEndsWith(installDir, "/lib/dmtcp/32")) {
+          // If dmtcp_launch was compiled for 32 bits in 64-bit O/S, then note:
+          // DMTCP_ROOT/bin/dmtcp_launch is a symbolic link to:
+          //    DMTCP_ROOT/bin/dmtcp_launch/lib/dmtcp/32/bin
+          // GetProgramDir() followed the link.  So, need to remove the suffix.
+          char *str = const_cast<char*>(installDir.c_str());
+          str[strlen(str) - strlen("/lib/dmtcp/32")] = '\0';
+          installDir = str;
+        }
+#endif
+
         /* We need to initialize SharedData here to make sure that it is
          * initialized with the correct coordinator timestamp.  The coordinator
          * timestamp is updated only during postCkpt callback. However, the
@@ -295,6 +307,8 @@ class RestoreTarget
 
 #if defined(__x86_64__) || defined(__aarch64__)
       is32bitElf = (_pInfo.elfType() == ProcessInfo::Elf_32);
+#elif defined(__i386__) || defined(__arm__)
+      is32bitElf = true;
 #endif
       runMtcpRestart(is32bitElf, _fd, &_pInfo);
 


karya0:
Let's document the policy before we do any m32 merges :-).

gc00:
We're making good progress.  When this pull request is included with the current master, the following recipe can now build a standalone m32 in the local directory:
`./configure --enable-m32 CXXFLAGS="-g3 -O0" && make -j clean && rm -rf lib bin && make -j && (cd src && make -j) && (cd lib/dmtcp && ln -s 32/lib/dmtcp/* ./) && (mkdir -p bin && cd bin && ln -s ../lib/dmtcp/32/bin/* ./) && make -j check`
This is excellent progress!  (By the way, note that an m32 build can never checkpoint a 64-bit app, and so some tests will always fail.)
Standalone m32 is now failing on second checkpoint, which seems to be a new bug, but let's check in this restart fix, so that we have firm ground on which to build.

Finally, it's not clear to me what policy we need to document for this bug fix commit.

gc00:
I think this is ready to go.  Let me know if it looks good.  (It fixes a bug in dmtcp_restart work for standalone m32.)

karya0:
Let's wait for my m32 fix to arrive before we merge this. It's quite likely that we won't need this patch after all. I will add a comment once I have the PR ready.

gc00:
I think what you're proposing would be a big mistake, @karya0.  It assumes that your commit will do everything right, and we will never ever want to see a previous working standalone m32.  Here's the issue:
I propose that we push in my pull request #72 (here) anyway.  And then you apply your commit on top of that -- even if your commit undoes my changes.  The issue is this:
1.  standalone m32 didn't work since 2.4.0
2.  Through a series of pull requests, I've gotten m32 working.
3.  This is the last pull request in that series.  All the others are already in master.  For example, what I do here to dmtcp_restart was already done to dmtcp_launch.
4.  Suppose your patch has a bug.  With my commit, you can compare my working standalone m32 with anything that goes wrong in your commit.  Otherwise, we have to compare DMTCP 2.4.0 with your commit.  _This is not egoless programming._

So, in summary, the key to evolving complex software is to take many small steps.  I think it's a mistake for you to use DMTCP 2.4.0 as your reference copy of a working standalone m32, and throw away all m32-related commits between 2.4.0 and the present.

Yes, in theory, we have a mostly working m32 right now that is missing one pull request.  So, if we get into a lot of trouble, we can save the missing pull request in a github issue right now as insurance.  Then we can go back to the commit before yours, and put the last brick in place (the missing pull request) o see what a working m32 would have looked like if we hadn't decided that DMTCP-2.4.0 was the preferred reference version.  But I stiill think that would be a mistake.

In sum, the last building block for a working m32 has been under review for 6 days.  It is the last piece of three or four related pull requests.  I think it is a misuse to allow the first three pull requests to go into master, and then to stall the last one for 6 days, and then to decide that no one should ever be able to see what the full set of pull requests would look like, because we're going to tear out all four pull requests and place it with a new version.

I'll also say that this is how we got into trouble when multi-arch was broken on every CPU architecture for a month.  It would have been possible to split "tearing away the duct tape" and "adding vvar support" in two different commits.  Instead it was all in one commit, because we were sure that the commit was correct.  When it turned out that the commit was not correct, we had no easy way to keep the vvar support and put back the duct tape, because those were both inside one very large commit.  _This also is not egoless programming._

karya0:
Ok, let's merge it.

gc00:
Done.  Thanks. :-)

PR #77, opened 2015-04-28
make clean-64: keeps lib/dmtcp/32; for multi-arch

gc00:
Added clean-64 target.  With this target, the following recipe for a local multi-arch works:
  ./configure --enable-m32 && make clean && make -j
  ./configure && make clean-64 && make -j
Things like the following then work:
  gcc -m32 -o dmtcp-tmp test/dmtcp1.c
  bin/dmtcp_launch -i6 ./dmtcp-tmp
  bin/dmtcp_restart ckpt_dmtcp-tmp_*.dmtcp

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -189,6 +189,12 @@ clean: tidy
 	 rm -rf $$DMTCP_TMPDIR
 	rm -rf $(targetdir)/lib $(targetdir)/bin
 
+# clean-64 preserves files in lib/dmtcp/32.  This is useful for a local multi-arch.
+clean-64: tidy
+	mv ./lib/dmtcp/32 ./32
+	${MAKE} clean
+	mkdir -p ./lib/dmtcp && mv ./32 ./lib/dmtcp/32
+
 distclean: clean
 	- cd src && $(MAKE) distclean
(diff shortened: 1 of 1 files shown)


karya0:
Before we jump into helper scripts/targets, let's document what do we want from --enable-m32 and how do we want to achieve multi-arch support. I would rather prefer a simpler `--enable-m32` even if it means that the multi-arch build requires a bit more complex procedure. Do you agree?

gc00:
At this time, I only want to add the target clean-64.  My intention was not to create any new scripts  So, I was hoping this would be non-controversial.  (Don't use the target if you don't like it.)

 In the DMTCP FAQ, http://dmtcp.sourceforge.net/FAQ.html#intel32-64 , I will probably quote the new recipe that lets one experiment with multi-arch locally without needing any kind of 'install'.  It always seemed sttrange to me that we can't use multi-arch unless we install into some other location.  glibc is the only other software I know of, that requires that.

karya0:
Let's commit it. At a later time when we have a better design for solving the m32 standalone/multi-arch stuff, we can revisit this patch.

gc00:
Done.  :-)

PR #81, opened 2015-04-29
Cleaning up <DMTCP_ROOT>/test/plugin

gc00:
Look at README, and one Makefile to see the pattern.
Three commits:
* README:  make clean tidy -> make clean
* Makefile:  dmtcp_NAMEhijack.so -> libdmtcp_NAME.so (and use word 'plugin library' in comments; not 'hijack library')
* Makefile:  LINKER = ${CC}    [ and the .so target uses ${LINKER} ... ]

QUESTION:  Is there a convention for the macro name LINKER (used to create a .so file from .o files)?

### DIFF
--- a/test/plugin/applic-delayed-ckpt/Makefile
+++ b/test/plugin/applic-delayed-ckpt/Makefile
@@ -17,6 +17,7 @@ DMTCP_INCLUDE=${DMTCP_ROOT}/include
 
 CFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g
 CXXFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g
+LINK = ${CC}
 
 DEMO_PORT=7781
 
@@ -45,7 +46,7 @@ check: ${LIBNAME}.so applic
 
 # We link the library using C++ for compatibility with the main libdmtcp.so
 ${LIBNAME}.so: ${LIBOBJS}
-	${CXX} -shared -fPIC -o $@ $^
+	${LINK} -shared -fPIC -o $@ $^
 
 .c.o:
 	${CC} ${CFLAGS} -o $@ $<
--- a/test/plugin/applic-delayed-ckpt/Makefile
+++ b/test/plugin/applic-delayed-ckpt/Makefile
@@ -3,10 +3,10 @@
 # The name will be the same as the current directory name.
 NAME=${shell basename $$PWD}
 
-# By default, your resulting library will have this name.
-LIBNAME=dmtcp_plugin-to-announce-events
+# By default, your resulting plugin library will have this name.
+LIBNAME=libdmtcp_plugin-to-announce-events
 
-# As you add new files to your hijack library, add the object file names here.
+# As you add new files to your plugin library, add the object file names here.
 LIBOBJS = plugin-to-announce-events.o
 
 # Modify if your DMTCP_ROOT is located elsewhere.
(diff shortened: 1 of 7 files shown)


karya0:
We use `CXXLINK` as the macro name for `LINKER` that you used above. http://www.gnu.org/software/automake/manual/html_node/How-the-Linker-is-Chosen.html

test/plugin/applic-delayed-ckpt/Makefile
```diff
@@ -17,6 +17,7 @@ DMTCP_INCLUDE=${DMTCP_ROOT}/include
 
 CFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g
 CXXFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g
+LINKER = ${CC}
```
karya0: s/LINKER/LINK ?

http://www.gnu.org/software/automake/manual/html_node/How-the-Linker-is-Chosen.html

test/plugin/applic-initiated-ckpt/Makefile
```diff
@@ -17,6 +17,7 @@ DMTCP_INCLUDE=${DMTCP_ROOT}/include
 
 CFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g
 CXXFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g
+LINKER = ${CC}
```
karya0: Same as above.

test/plugin/applic-initiated-ckpt/Makefile
```diff
@@ -44,7 +45,7 @@ check: ${LIBNAME}.so applic
 
 # We link the library using C++ for compatibility with the main libdmtcp.so
 ${LIBNAME}.so: ${LIBOBJS}
-	${CXX} -shared -fPIC -o $@ $^
+	${LINKER} -shared -fPIC -o $@ $^
```
karya0: ditto

test/plugin/example-db/Makefile
```diff
@@ -17,6 +17,7 @@ DMTCP_INCLUDE=${DMTCP_ROOT}/include
 
 CFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g
 CXXFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g
+LINKER = ${CC}
```
karya0: same as above

karya0:
BTW, we use the following definition for `CXXLINK`:
 
```CXXLINK = $(CXX) $(CXXFLAGS) $(LDFLAGS) -o $@                                    
```

Let's use a similar one for `LINK`.

test/plugin/applic-delayed-ckpt/Makefile
gc00: Agreed (LINKER -> LINK).  Thanks.  I was looking for the right convention.

gc00:
I changed LINKER -> LINK.  (Nice.  I was looking for the conventional name.)
As for `CXXLINK = $(CXX) $(CXXFLAGS) $(LDFLAGS) -o $@ `,
the rest of that Makefile does things like:  `CFLAGS += -I${DMTCP_INCLUDE} -DDMTCP -fPIC -c -g`
followed by:  `${CC} ${CFLAGS} -o $@ $<`
In that spirit, I could define `LDFLAGS = -shared -fPIC`
and follow it by: ` ${LINK} ${LDFLAGS} -o $@ $^`

Whatever we do, this is intended as an easy example, readable instantly by a novice.  That means, keeping the recursion of macro expansion only one level deep.  Note that `LINK = ... ${LDFLAGS}` followed by `LDFLAGS = ...` would be two levels deep.

My first thought would be to leave it as is, but I don't mind introducing `LDFLAGS` if you prefer that.  Tell me what you'd prefer.

karya0:
Let's keep it simple for this case. For new real plugins, we want to follow the same convention as with existing plugins.

karya0:
If nothing else needs to be done, this PR can be merged.

gc00:
Done. :-)

PR #83, opened 2015-05-01
Cleanup exitInProgress logic in DmtcpWorker

rohgarg:


src/threadsync.cpp
```diff
@@ -368,6 +368,10 @@ bool ThreadSync::wrapperExecutionLockLock()
 {
   int saved_errno = errno;
   bool lockAcquired = false;
+  // Ignore locks if we are about to exit
+  if (DmtcpWorker::exitInProgress()) {
+    return true;
```
karya0: We should return `false` instead. This will then avoid the need to check for exit status in the unlock code.

src/threadsync.cpp
```diff
@@ -439,6 +443,10 @@ bool ThreadSync::wrapperExecutionLockLockExcl()
...
```
karya0: ditto.

src/threadsync.cpp
```diff
@@ -461,14 +469,14 @@ bool ThreadSync::wrapperExecutionLockLockExcl()
...
-            "       This should not be happening, something is wrong.\n",
-            __FILE__, __LINE__, __PRETTY_FUNCTION__);
-    _exit(1);
+  // Ignore locks if we are about to exit
+  /* XXX:
+   *   One of the user threads could have called exit(), setting the
+   *   exitInProgress flag, and changing the WorkerState to Unknown. Does it
+   *   matter if we ignore the lock?
+   */
+  if (DmtcpWorker::exitInProgress()) {
```
karya0: We can get rid of this block now.

rohgarg: I'm sorry I didn't understand which block you're referring to. The original if clause has been replaced by a new clause that causes the function to return if exitInProgress is set.

karya0: I meant we can delete the `if (DmtcpWorker::exitInProgress()) { ... }` since there is no harm in unlocking the lock if it was acquired earlier. However, if the lock was not acquired (due to exiInProgress being set), this function won't be called anyways.

rohgarg: I would keep this check. It's like saying that the arguments to system call will always be correct, and that a client (a wrapper) will always do the right thing. But we have seen cases, for example the wrappers in the IB plugin, where a client can ignore the return values of the lock/unlock functions, not necessarily on purpose.

This just serves as an additional check.

karya0: In that case, let's add a comment on top of it. Also, let's add a JWARNING here to indicate such a situation so we can catch this scenario. We can merge it after that :-).

rohgarg: Sounds good!

rohgarg: Actually, adding a warning is not straightforward. :-)

There are two difficulties with that:
 1) There can be a scenario where a user thread called exit, when another thread was in the middle of a wrapper, after having acquired the wrapperExecution lock. I would argue that it is safe to return from the unlock function w/o having done anything because the kernel is going to kill the process anyway.
 2) Even in the most simple case, we need some way for determining that a client is calling the unlock function even though the wrapperExecution lock was never acquired.

Adding a meaningful warning message requires maintaining a non-trivial global state that addresses the above issues. Please let me know what you think.

karya0: We already have a global per-thread global variable that counts The number of times this lock was acquired (_wrapperExecutionLockLockCount). If it's trivial to ise that variable to add the warning, let's do that, otherwise please add a Todo in there and then we can commit it.

rohgarg:
Closed via 3235760e8914c5ae329df1d6621e8ff2c34e8d56

PR #84, opened 2015-05-01
Print extra info for util/readdmtcp.sh

gc00:
This improves on util/readdmtcp.sh by having `mtcp_restart --simulate` also print out most of the information in `mtcpHdr`, such as vdso, vvar, postRestart entry point, etc.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -123,7 +123,7 @@ static void restart_slow_path(void);
 static int doAreasOverlap(VA addr1, size_t size1, VA addr2, size_t size2);
 static int hasOverlappingMapping(VA addr, size_t size);
 static void getTextAddr(VA *textAddr, size_t *size);
-static void mtcp_simulateread(int fd);
+static void mtcp_simulateread(int fd, MtcpHeader *mtcpHdr);
 void restore_libc(ThreadTLSInfo *tlsInfo, int tls_pid_offset,
                   int tls_tid_offset, MYINFO_GS_T myinfo_gs);
 static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo);
@@ -207,7 +207,8 @@ MTCP_PRINTF("Attach for debugging.");
   while (argc > 0) {
     // Flags for standalone debugging
     if (argc == 1) {
-      MTCP_PRINTF("Considering '%s' as a ckpt image.\n", argv[0]);
+      // We would use MTCP_PRINTF, but it's also for output of util/readdmtcp.sh
+      mtcp_printf("Considering '%s' as a ckpt image.\n", argv[0]);
       ckptImage = argv[0];
       break;
     } else if (mtcp_strcmp(argv[0], "--use-gdb") == 0) {
@@ -269,7 +270,7 @@ MTCP_PRINTF("Attach for debugging.");
             " `text_offset.sh mtcp_restart`\n    in the mtcp subdirectory.\n");
 
   if (simulate) {
-    mtcp_simulateread(rinfo.fd);
+    mtcp_simulateread(rinfo.fd, &mtcpHdr);
     return 0;
   }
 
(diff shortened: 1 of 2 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -207,7 +207,7 @@ MTCP_PRINTF("Attach for debugging.");
   while (argc > 0) {
     // Flags for standalone debugging
     if (argc == 1) {
-      MTCP_PRINTF("Considering '%s' as a ckpt image.\n", argv[0]);
+      mtcp_printf("Considering '%s' as a ckpt image.\n", argv[0]);
```
karya0: Why this change? We decided to use MTCP_PRINTF and DPRINTF exclusively, unless we are printing some raw data where we don't want to print line/function information with every line.

src/mtcp/mtcp_restart.c
```diff
@@ -463,11 +463,25 @@ static void restart_slow_path()
...
-static void mtcp_simulateread(int fd)
+static void mtcp_simulateread(int fd, MtcpHeader *mtcpHdr)
 {
   int mtcp_sys_errno;
+
+  // Print miscellaneous information:
+  char buf[MTCP_SIGNATURE_LEN+1];
+  mtcp_memcpy(buf, mtcpHdr->signature, MTCP_SIGNATURE_LEN);
+  buf[MTCP_SIGNATURE_LEN] = '\0';
+  mtcp_printf("\nMTCP: %s", buf);
```
karya0: Here it is okay to use `mtcp_printf` because we are printing a lot of raw data. Can you also add a MTCP_PRINTF at the top of this function that would act as a header for this raw data?

src/mtcp/mtcp_restart.c
```diff
@@ -463,11 +463,25 @@ static void restart_slow_path()
...
+              mtcpHdr->restore_addr + mtcpHdr->restore_size);
+  mtcp_printf("**** DMTCP entry point (ThreadList::postRestart()): %p\n",
+              mtcpHdr->post_restart);
+  mtcp_printf("**** brk (sbrk(0)): %p\n", mtcpHdr->saved_brk);
+  mtcp_printf("**** vdso: %p-%p\n", mtcpHdr->vdsoStart, mtcpHdr->vdsoEnd);
+  mtcp_printf("**** vvar: %p-%p\n", mtcpHdr->vvarStart, mtcpHdr->vvarEnd);
+
   Area area;
-  MTCP_PRINTF("Listing ckpt image area:\n");
+  mtcp_printf("\n**** Listing ckpt image area:\n");
```
karya0: This should be promoted to `MTCP_PRINTF` as well since it's a header.

src/mtcp/mtcp_restart.c
gc00: The primary use of this function is to print to stdout when called by readdmtcp.sh.  The routine MTCP_PRINTF() will print the file, function, line number, etc., followed by an indented message on the next line.  I don't think people calling readdmtcp.sh would want to see what function and line number is delivering the information to them.  (For example, readelf doesn't do that.)

src/mtcp/mtcp_restart.c
gc00: Same comment as before.  readelf.sh doesn't report the internal function and line number delivering the information to us.  util/readdmtcp.sh shouldn't be doing that either.

src/mtcp/mtcp_restart.c
karya0: Could you also add a comment here that states the fact that this condition is true only for readmtcp and that's why it's okay to use mtcp_printf here. Alternatively, could we just remove this line from here and move it in `mtcp_simulateread`?

src/mtcp/mtcp_restart.c
```diff
@@ -207,7 +207,8 @@ MTCP_PRINTF("Attach for debugging.");
   while (argc > 0) {
     // Flags for standalone debugging
     if (argc == 1) {
-      MTCP_PRINTF("Considering '%s' as a ckpt image.\n", argv[0]);
+      // We would use MTCP_PRINTF, but it's also for output of util/readdmtcp.sh
```
gc00: This comment was added, based on the review.

gc00:
I've added the comment on line 10 as proposed.  Let me know if it's ready to go.

karya0:
It's ready to go. If you get a chance, can you add a short comment at the top of `mtcp_simulateread` explaining the mtcp_printf vs MTCP_PRINTF choice.

gc00:
Done.  :-)

PR #88, opened 2015-05-02
ib-plugin: check return value of dmtcp_plugin_disable_ckpt

rohgarg:
The dmtcp_plugin_disable_ckpt() function makes an attempt
to acquire the wrapperExecutionLock and returns true if
it was successful.

We should check of its return value before trying to
release the lock using dmtcp_plugin_enable_ckpt().

### DIFF
--- a/contrib/infiniband/infinibandwrappers.c
+++ b/contrib/infiniband/infinibandwrappers.c
@@ -23,121 +23,121 @@ void *dlopen(const char *filename, int flag) {
 
 int ibv_fork_init(void)
 {
-  dmtcp_plugin_disable_ckpt();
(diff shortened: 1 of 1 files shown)


karya0:
Good catch. However, as for the fix, there is already a similar construct (`DMTCP_PLUGIN_DISABLE_CKPT`) defined in dmtcp.h and we should be using that instead of defining a similar construct. So, I would suggest that we replace dmtcp_plugin_{dis,en}able_plugin with the corresponding macros.

karya0:
Let's squash the two commits together and merge into the master.

rohgarg:
Closed via f2d2a7c6feba38ab0b0cb8e09a4ad6cc37d9f330

PR #89, opened 2015-05-02
If m32, skip tests of autotest on 64-bit apps

gc00:
Check Python variable USE_M32, and skip any tests in m32 mode that invoke 64-bit processes.


gc00:
This one is a small change, and should be non-controversial.  Can we push this in, if the rc4 release is going to be supporting standalone m32?  Thanks.

karya0:
This looks good; let's merge it. I have also created an issue (#90) to auto-disable 64-bit tests on a standalone 32-bit build.

PR #96, opened 2015-05-08
ib-plugin: Fix compilation error introduced in f2d2a7

rohgarg:


karya0:
The correct fix is to change `bool` to `int` in dmtcp.h for the following line:
```   bool __dmtcp_plugin_ckpt_disabled = dmtcp_plugin_disable_ckpt()
```

This way we still keep dmtcp.h favorable for .c files.

rohgarg:
Updated the PR.

karya0:
Let's commit it (after squashing the two commits).

rohgarg:
Closed via 229add881d22b5bbc372c260483352fbc5925dbd.

Issue #79, opened 2015-04-29
Rename `is32bit` with a more precise name


karya0:
Looks like `is32bit` is a bad variable name when dealing with 32-bit binaries on 64-bit systems. We want something that says "it's a 32-bit binary on a 64-bit system".  Is there a common convention that people use? If so, we can create a new variable that is set to true only in the said case and remains false otherwise.

gc00:
@karya0:  Really good point.
What about isM32 or isMultiarch32?
 
For inspiration, I started looking at https://wiki.debian.org/Multiarch and: https://wiki.debian.org/Multiarch/LibraryPathOverview .
 
I found this paragraph:
To solve this problem, the GCC back-end may provide a secondary OS  
multilib suffix which is used in place of the primary multilib suffix 
for all library directories derived from *system* paths as opposed to 
GCC paths. For example, in the typical bi-arch setup, the -m32 option is associated with the OS multilib suffix "../lib". Given the that primary
system library directory is /usr/lib64 on such systems, this has the
effect of causing the toolchain to search ...

gc00:
Still thinking about the right name.  What about `is32bitElfIn64bitOS`?

gc00:
Also, a better design would be to set is32bitElf (or the new name) always to `false` when we're in a 32-bit Linux O/S.  After pull request #97 (to fix a bug in 32-bit Linux), let's do a later pull request with this fix.
&nbsp;&nbsp;&nbsp;&nbsp;(It's better to get DMTCP working correctly first, and then do polishing of the code in a later commit.)

PR #97, opened 2015-05-08
is32bitElf should not be used with a 32-bit Linux

gc00:
An alternative design would be to set is32bitElf to false in a 32-bit OS.  All of this is also related to issue #79 (choosing a better name than is32bitElf).

Let's push in this fix, and do the polishing later.  As of now, rc4 does not work on 32-bit Linux without this commit.


karya0:
Ship it.

gc00:
Done.  :-)

PR #99, opened 2015-05-08
Fix capitalizations for some ipc method names

gc00:
CreateDirectoryStructure(), Create(), Null(), and Self() are all methods (not constructors) that start with capital letters.  It's been that way probably since 2012.  This seems to violate standard C++ conventions.  So, I'm changing them to start with lower case.

Also, Self() seems never to be used.  I added a FIXME comment in front of it.  Should we just remove the Self() method?


karya0:
Ship it.

gc00:
Done.  :-)

PR #100, opened 2015-05-08
Fix 3b06a815 : buggy condition in 'test' command

gc00:
The '==' operator is not legal for the 'test' command.  Also, I added a test if the 'id' command exists, as extra insurance against this blowing up.


karya0:
LGTM

gc00:
Done.  :-)

PR #103, opened 2015-05-10
Env var DMTCP_GDB_ATTACH_ON_RESTART allows attach

gc00:
Linux distros of the last few years don't allow 'gdb PROGNAME PID' when the PID is not a child process of the current process.  This allows:

    DMTCP_GDB_ATTACH_ON_RESTART=1 dmtcp_restart ckpt_a.out_*.dmtcp &
    gdb a.out `pgrep -n a.out`
(I plan to include the recipe above in the DMTCP FAQ.)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,7 +4,7 @@ Please send DMTCP bug reports via <dmtcp-forum@lists.sourceforge.net>
 
 Version 2.4 release notes
 =========================
-Several important changes and enancements were added:
+Several important changes and enhancements were added:
 * dmtcp_launch/restart/command/coordinator now take the flags
     -h, -p, --coord-host/port and environment variables
     DMTCP_COORD_HOST/PORT.  The older --host, --port, DMTCP_HOST/PORT
@@ -26,6 +26,11 @@ Several important changes and enancements were added:
      cat /proc/self/maps | grep '\[vvar]'
 * Support for glibc version 2.21 added.  To see if this affects you, do:
      ls -l /lib*/libc.so.6 /lib/*/libc.so.6
+* The environment variable DMTCP_GDB_ATTACH_ON_RESTART was added.  Setting
+     this permanently is a security risk. But on a temporary basis,
+     it can enable easier debugging of restarted processes:
+       DMTCP_GDB_ATTACH_ON_RESTART=1 dmtcp_restart ckpt_a.out_*.dmtcp &
+       gdb a.out `pgrep -n a.out`
 * Enhancements added for newer 32-bit ARM (armv7) CPUs
 * Experimental support is now provided for 64-bit ARM (armv8)
 * Bug fixes
@@ -338,7 +343,7 @@ Version 1.2.4 release notes
   checkpoint-restart service of Open MPI.  The usage will be documented soon
   through the Open MPI web site.  As before, an alternative is to simply start
   Open MPI inside DMTCP, and let DMTCP treat all of Open MPI as a "black box"
-  that happens to be a ditributed computation
+  that happens to be a distributed computation
 - A new --prefix command line flag has been added to dmtcp_checkpoint.  It
   operates similarly to the flag of the same name in Open MPI.  For distributed
   computations, remote processes will use the prefix as part of the path to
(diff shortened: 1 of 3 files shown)


NEWS
```diff
@@ -26,6 +26,10 @@ Several important changes and enancements were added:
      cat /proc/self/maps | grep '\[vvar]'
 * Support for glibc version 2.21 added.  To see if this affects you, do:
      ls -l /lib*/libc.so.6 /lib/*/libc.so.6
+* The environment variable DMTCP_GDB_ATTACH_ON_RESTART was added.  Setting
```
karya0: Not sure if we should be exposing internal debugging env vars to the end user. I haven't seen any other release notes from other projects following this suit.

karya0: We should instead be adding the description of this flag to the debugging documentation.

gc00: I had meant to add something like the following to the print statement:
```Setting this can be a security risk.  But on a temporary basis, it can enable
easier debugging of restarted processes:```
[ I agree that we should also add this to:  `DMTCP_ROOT/doc/debugging-dmtcp.txt` .  I'll include the revised print statement and modified `debugging-dmtcp.txt` in the next iteration.

**Comment by gc00** [comment]
@karya0:  This revised version also includes a mostly rewritten version of doc/debugging-dmtcp.txt.  You'll want to read that as a new file, and not as a diff with the old file.

File: NEWS (hunk 1 above)
Comment (karya0) [inline NEWS]: Hmm. I am not sure if we can ever support `gdb --args dmtcp_restart ckpt*.dmtcp` to work normally. However, I agree that the user should be able to "attach" to their processes after restart. In that case, wouldn't it be better to also supply a command line flag to dmtcp_restart to complement this env var? I personally would prefer to use a flag over an environment variable :-). Having said that, let's commit it and create a new issue to add the command line flag.

File: src/dmtcp_restart.cpp [hunk 2]
```diff
@@ -345,6 +349,20 @@ static void runMtcpRestart(int is32bitElf, int fd, ProcessInfo *pInfo)
   sprintf(fdBuf, "%d", fd);
   sprintf(stderrFdBuf, "%d", PROTECTED_STDERR_FD);
 
+#ifdef HAS_PR_SET_PTRACER
+  if (getenv("DMTCP_GDB_ATTACH_ON_RESTART")) {
+    JNOTE("\n     *******************************************************\n"
```
karya0: JASSERT_STDERR would have been a better choice here instead of JNOTE. It would have looked like:

JASSERT_STDERR << "....." << getpid();

However, since we don't have a uniform policy about it, let's commit it as is.

karya0:
The code looks good to me, but I haven't had a chance to look at the documentation. Could @rohgarg or @jiajuncao take a look at it and then we can commit it.

doc/debugging-dmtcp.txt
```diff
@@ -2,35 +2,128 @@ This is intended as one of a series of informal documents to describe
...
+Ideally, we would like to just use:  gdb --args dmtcp_restart ckpt_*.dmtcp
+This does not work (unless you just want to debug the command dmtcp_restart
+itself).  But the command dmtcp_restart overlays the old memory from
+ckpt_*.dmtcp into the current running 'dmtcp_restart' process.  As you would
+expect, GDB has difficulty following this memory overlay.  Therefore,
+the best way to debug a restarted process is to use the 'attach' feature
+of GDB.
+
+If you want to debug DMTCP functions also, be sure to compile DMTCP with
+debugging support.  A convenient recipe from the root directory of DMTcP is:
```
rohgarg: Minor typo: DMTcP --> DMTCP

doc/debugging-dmtcp.txt
```diff
@@ -2,35 +2,128 @@ This is intended as one of a series of informal documents to describe
...
+       # DMTCP will then pause for 15 seconds, and wait for a GDB attach.
+
+3. DMTCP_RESTART_PAUSE:
+If you need to attach with GDB _very_ early in the restart process (usually
+needed only by DMTCP developers), do:
+       DMTCP_RESTART_PAUSE=1 bin/dmtcp_launch a.out
+       # Continue as above.
+In this case, you will attach at a time when only the primary thread
+exists, and the DMTCP checkpoint thread is not yet distinguished from
+the original user thread executing main().
```
rohgarg: DMTCP_RESTART_PAUSE2 pauses DMTCP immediately after all the threads (including the checkpoint thread) have been restored?

rohgarg:
Except for the two minor nits, looks good to me.

gc00:
I added the two changes and pushed it in.  :-)

Issue #62, opened 2015-04-18
--enable-m32 doesn't work standalone


gc00:
`./configure --enable-m32` is supposed to work standalone, not just as part of a multi-arch build.  Currently, to make it work, one must do something like:
 `make -j clean && rm -rf bin lib && make -j && (cd src && make -j || true) && 
    (cp -prf lib/dmtcp/32/bin/ bin || true) &&
   make -j check-dmtcp1`
to make `--enable-m32` work with all DMTCP commands built for m32.  We should modify the make files to do this automatically, and `rm -rf bin lib` as necessary if this is part of a multi-arch build.
[ In addition to supporting other users, it makes it much easier for the developers to test m32 mode. ]

karya0:
At the first read, your argument sounds convincing.  However, this is rather confusing from the point of view of packaging. As of our current design, there are no 32-bit binaries (except mtcp_restart_32), only 32-bit libraries. Commands like dmtcp_launch and dmtcp_restart automatically do the correct thing by linking the required 32/64 bit libraries as needed. In essence, the 32-bit libraries are supposed to act as auxiliary units for the 64-bit binaries and libraries.

Having said that, we definitely need something to improve automated build of 32-bit binaries. One possibility is to write a short script to put in util/ that builds 64-bit and 32-bit rather seamlessly. Any thoughts?

gc00:
I encountered this issue in trying to do a 'git bisect'
through the history, using --enable-m32.  I have a version
of what mostly works in my testing, as part of my pull request
for git-bisect-example (git-bisect-m32.sh).  The core is:

`./configure --enable-m32 && make -j clean && rm -rf bin lib && (PATH=$PATH:. make -j) && \
  make -j clean && rm -rf bin lib && make -j && (cd src && make -j || true) && \
    (mkdir -p bin && cd bin && ln -sf ../lib/dmtcp/32/bin/* ./ || true) && \
    (cd lib/dmtcp && ln -sf 32/lib/dmtcp/* ./ || true) && \
    (cd lib/dmtcp/32/bin && ln -s mtcp_restart-32 mtcp_restart || true) && \
  make -j check-dmtcp1`

If we had --enable-m32 do this, it should work standalone.
(The link of `mtcp_restart-32` to `mtcp_restart` is not needed.  That's
only for an older revision.)

I believe that this version of --enable-m32 would not interfere
with our multi-arch build.  I would test the multi-arch build as
part of this enhancement.

[...]

PR #105, opened 2015-05-10
Standalone m32 now builds easily:  configure --enable-m32 && make clean && make

gc00:
@karya0:  We had wanted to talk together about a clean design to support both standalone --enable-m32 and multi-arch.  There hasn't been time to do it yet.  So, I created his commit as one possible design.  This can serve as a _baseline_ for further discussions.

My goal here is to push into the development branch one clean way to do this.  Then we can talk together about changing _this_ baseline to find the cleanest possible design.  That will be easier than comparing several new designs against a baseline in the development branch that is only partially working.

I've tested this commit in four modes:
* Native 32-bit Ubuntu distro
* 64-bit distro, standalone m32:

    `configure --enable-m32 && make clean && make`
* multi-arch:

    `./configure --enable-m32 && make clean && make -j`
    `./configure && make clean-64 && make -j`
    `gcc -m32 test/dmtcp1.c`
    [ And test DMTCP on the `a.out` generated above. ]`
* multi-arch with `--prefix=.../dmtcp-build`, while adding `&& make install`` to above recipe

All four of the above cases are now working when I use this commit.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -59,7 +59,12 @@ MANPAGES=dmtcp.1.gz dmtcp_coordinator.1.gz dmtcp_command.1.gz dmtcp_launch.1.gz 
 
 default: display-build-env build
 
+# If M32, 'rm -rf lib bin'.  Can't do a 32-bit build on top of a 64-bit build,
+#   neither './configure --enable-m32' nor 'make clean' will remove these dirs.
 mkdirs:
+ifeq ($(M32),1)
+	rm -rf lib bin
+endif
 	$(MKDIR_P) $(targetdir)/bin
 	$(MKDIR_P) $(targetdir)/lib/dmtcp
 
@@ -100,6 +105,10 @@ display-config:
 
 dmtcp:
 	cd src && $(MAKE)
+ifeq ($(M32),1)
+	cd lib/dmtcp && ln -s 32/lib/dmtcp/* ./
+	mkdir -p bin && cd bin && ln -s ../lib/dmtcp/32/bin/* ./
+endif
 
 plugin: dmtcp
 	cd plugin && $(MAKE)
--- a/doc/multi-arch.txt
+++ b/doc/multi-arch.txt
@@ -27,11 +27,13 @@ A. Compiling DMTCP for multi-arch support:
    ./configure --enable-m32
    make clean
    make -j
-   make install
 2. Do the usual 64-bit compilation
    ./configure
+   make clean-64
    make -j
-   make install
+
+Optionally, add 'make install' at the end of each of the above stages,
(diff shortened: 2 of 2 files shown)


karya0:
I understand your motivation, however, I don't think this is getting us any closer to the baseline. For example, with the current patch, if you do make twice, it will delete the lib and bin dirs on the second make and will have to rebuild the binaries and libs even if there were no changes to the corresponding files.

A clean fix is to restore the original -m32 (pre multi-arch) behavior and then add multi-arch support as a special case. The multi-arch should be the one that deals with 'lib/dmtcp/32' stuff, not --enable-m32. Does this seem reasonable?

gc00:
@karya0 wrote:  
_> The multi-arch should be the one that deals with 'lib/dmtcp/32' stuff, not --enable-m32. Does this seem reasonable?_

I agree with what you write.  It's completely reasonable.  But here's my point again.  (If we end up in a tight loop, let's switch to a phone call.)

The short story is:  I'm proposing a reference case with code that's verified to work in all four modes.  I am not proposing the final solution.  We will put it into rc5, and then tear it out again when we have our clean design.  But an interim reference case is useful for avoiding bugs when complex logic is involved.  Or, to put it another way, the principle is:
&nbsp;&nbsp;&nbsp;&nbsp;_When dealing with complex logic, take small steps._ (This change involves just seven executable lines in one Makefile.)

The long story is:  
1. Right now, standalone m32 doesn't work, and it hasn't worked since version 2.0
2. I'd like to see this commit pushed in, in time for 2.4.0-rc5.  (We will tear it out after rc5, when the clean design is ready.)
3.  It's trivial to back out of this small change after 2.4.0-rc5.
4.  In the meantime, we will have a landmark, with which to compare functionality.
5.  Getting the logic for standalone m32 right is not trivial, since it currently involves many #define's, and since there are four modes (above) against which to test any new code.  So it's good to have a working reference case.  Besides this reference case, the only other possible reference case is version 2.0 (yuck!  :-) )
6.  So, the downside is minimal.  It won't create new bugs, and it's trivial to tear it out later.  The upside is that if we make a mistake in the clean design that we push in after 2.4.0-rc5, then both we and our users can compare 2.4.0-rc4 against 2.4.0-rc5, and point to a definite bug, not simply some weird mis-feature.

karya0:
Let's ship it :-).

I have now created an issue (#108) to track the high level design ideas for M32 and multi-arch (aka multilib support). Let's use that to converge on a medium/long term design goal :-).

gc00:
This closes #62.  Note that a new issue(referenced above) has been created for the high-level design ideas for M32.  Eventually, that issue will be used to create a cleaner long-term design.

PR #107, opened 2015-05-11
Fix logic for --slow flag of autotest

gc00:
Previously, the `--slow` flag had a pause after the first launch/ckpt, and that's all.  The new logic adds a pause:
a. after launch;
b. after checkpointing and killing the old process; and
c. after restart.

### DIFF
--- a/test/autotest.py
+++ b/test/autotest.py
@@ -116,8 +116,8 @@ for i in sys.argv:
   if i=="--stress":
     CYCLES=100000
   if i=="--slow":
-    SLOW=5
-    TIMEOUT *= 2
+    SLOW *= 5  # SLOW was initialized to 1 ; --slow --slow => SLOW==25
+    TIMEOUT *= SLOW
   if i=="--retry-once":
     RETRY_ONCE = True
   #TODO:  Install SIGSEGV handler with infinite loop, and add to LD_PRELOAD
@@ -220,10 +220,10 @@ def master_read(fd):
   os.read(fd, 4096)
   return ''
 
-#launch a child process
+#run a child process
 # NOTE:  Can eventually migrate to Python 2.7:  subprocess.check_output
 devnullFd = os.open(os.devnull, os.O_WRONLY)
-def launch(cmd):
+def runCmd(cmd):
   global devnullFd
   global master_read
   if VERBOSE:
@@ -335,8 +335,8 @@ if os.getenv('LD_LIBRARY_PATH'):
 else:
     os.environ['LD_LIBRARY_PATH'] = os.getenv("PWD")+"/lib"
 
-#launch the coordinator
-coordinator = launch(BIN+"dmtcp_coordinator")
+#run the coordinator
+coordinator = runCmd(BIN+"dmtcp_coordinator")
 
 #send a command to the coordinator process
 def coordinatorCmd(cmd):
@@ -477,8 +477,8 @@ def runTestRaw(name, numProcs, cmds):
       global coordinator
       coordinatorCmd('q')
       os.system("kill -9 %d" % coordinator.pid)
-      print "Trying to kill old coordinator, and launch new one on same port"
-      coordinator = launch(BIN+"dmtcp_coordinator")
+      print "Trying to kill old coordinator, and run new one on same port"
+      coordinator = runCmd(BIN+"dmtcp_coordinator")
     for x in procs:
       #cleanup proc
       try:
@@ -509,14 +509,26 @@ def runTestRaw(name, numProcs, cmds):
 
     #wait for files to appear and status to return to original
     WAITFOR(lambda: getNumCkptFiles(ckptDir)>0 and \
-                    (CKPT_CMD == 'xc' or doesStatusSatisfy(getStatus(), status)),
+                   (CKPT_CMD == 'xc' or doesStatusSatisfy(getStatus(), status)),
             wfMsg("checkpoint error"))
+    #we now know there was at least one checkpoint file, and the correct number
+    #  of processes have restarted;  but they may faily quickly after restert
+
+    if SLOW > 1:
+      #wait and give the processes time to write all of the checkpoint files
+      sleep(S*SLOW)
 
     #make sure the right files are there
     numFiles=getNumCkptFiles(ckptDir) # len(os.listdir(ckptDir))
     CHECK(doesStatusSatisfy((numFiles,True),status),
           "unexpected number of checkpoint files, %s procs, %d files"
           % (str(status[0]), numFiles))
+
+    if SLOW > 1:
+      #wait and see if some processes will die shortly after checkpointing
+      sleep(S*SLOW)
+      CHECK(doesStatusSatisfy(getStatus(), status),
+            "error: processes checkpointed, but died upon resume")
 
   def testRestart():
     #build restart command
@@ -525,9 +537,14 @@ def runTestRaw(name, numProcs, cmds):
       if i.endswith(".dmtcp"):
         cmd+= " "+ckptDir+"/"+i
     #run restart and test if it worked
-    procs.append(launch(cmd))
+    procs.append(runCmd(cmd))
     WAITFOR(lambda: doesStatusSatisfy(getStatus(), status),
             wfMsg("restart error"))
+    if SLOW > 1:
+      #wait and see if process will die shortly after restart
+      sleep(S*SLOW)
+      CHECK(doesStatusSatisfy(getStatus(), status),
+            "error:  processes restarted and then died")
     if HBICT_DELTACOMP == "no":
       clearCkptDir()
 
@@ -543,11 +560,12 @@ def runTestRaw(name, numProcs, cmds):
 
     #start user programs
     for cmd in cmds:
-      procs.append(launch(BIN+"dmtcp_launch "+cmd))
-      sleep(S*SLOW)
+      procs.append(runCmd(BIN+"dmtcp_launch "+cmd))
 
+    #TIMEOUT in WAITFOR has also been multiplied by SLOW
     WAITFOR(lambda: doesStatusSatisfy(getStatus(), status),
             wfMsg("user program startup error"))
+    #Will sleep(S*SLOW) in the following for loop.
 
     for i in range(CYCLES):
       if i!=0 and i%2==0:
@@ -556,6 +574,8 @@ def runTestRaw(name, numProcs, cmds):
       printFixed("ckpt:")
       # NOTE:  If this faile, it will throw an exception to CheckFailed
       #  of this function:  testRestart
+      #wait for launched processes to settle down, before we try to checkpoint
+      sleep(S*SLOW)
       testCheckpoint()
       printFixed("PASSED ")
       testKill()
@@ -900,10 +920,10 @@ if HAS_VIM == "yes":
   S=DEFAULT_S
 
 if sys.version_info[0:2] >= (2,6):
-  #On some systems, "emacs -nw" launches dbus-daemon processes in
+  #On some systems, "emacs -nw" runs dbus-daemon processes in
   #background throwing off the number of processes in the computation. The
-  #test thus fails. The fix is to launch emacs-nox, if found. emacs-nox
-  #doesn't launch any background processes.
+  #test thus fails. The fix is to run emacs-nox, if found. emacs-nox
+  #doesn't run any background processes.
   S=15*DEFAULT_S
   if HAS_EMACS_NOX == "yes":
     # Wait to checkpoint until emacs finishes reading its initialization files


test/autotest.py
```diff
@@ -556,8 +556,10 @@ def testRestart():
       printFixed("ckpt:")
       # NOTE:  If this faile, it will throw an exception to CheckFailed
       #  of this function:  testRestart
+      sleep(S*SLOW)
       testCheckpoint()
       printFixed("PASSED ")
+      sleep(S*SLOW)
```
karya0: If `testCheckpoint()` has indeed succeeded (evident from "PASSED" above), then what's the point of doing a sleep here?

test/autotest.py
```diff
@@ -526,6 +526,7 @@ def testRestart():
         cmd+= " "+ckptDir+"/"+i
     #run restart and test if it worked
     procs.append(launch(cmd))
+    sleep(S*SLOW)
```
karya0: We don't need this sleep, because the `WAITFOR` below already takes care of waiting for 10 seconds. If we indeed want to wait longer, then I would suggest that we change `10` seconds to some higher value if --slow was specified.

karya0:
I am not sure if I understand the reasoning behind (b) and (c). Could you elaborate more about the need for a sleep in these cases?

gc00:
@karya0: One of the `sleep(S*SLOW)` commands was indeed misplaced.  Thanks for catching that.

In this revision of my pull request, I've also now changed the Python function `launch()` to `runCmd()`.  It was okay in the days of `dmtcp_checkpoint`.  But now that it's called `dmtcp_launch', the name of the distinct Python function `launch()' was confusing.

@karya0:  You also asked my items (b) and (c).  I admit that I phrased it poorly.
For the difference between (b) and (c) in my original comment, I intended to indicate:
`dmtcp_launch`
sleep to see if the new process succeeds
`dmtcp_checkpoint`
sleep to see if the checkpoint file was created, _and also_ if resume succeeds
`dmtcp_restart`

Admittedly, my original  description, _after checkpointing and killing the old process_, was bad phrasing.
sleep to see if it succeeds

test/autotest.py
```diff
@@ -506,6 +506,8 @@ def testKill():
   def testCheckpoint():
     #start checkpoint
     coordinatorCmd(CKPT_CMD)
+    #wait and see if process will die shortly after checkpoint
+    sleep(S*SLOW)
```
karya0: I am not sure if this 'sleep' really serves the purpose. It's quite possible that this sleep will overlap with the time it takes to perform ckpt. I think you intended to 'sleep' to be _after_ the ckpt image has been written. In which case, this sleep statement should go after the ckpt image verification has taken place. In fact, this should come after between calls to `testCheckpoint()` and `testKill()`. To be precise, the code should look like the following:
```
testCheckpoint()
sleep(S*SLOW)
CHECK(doesStatusSatisfy(getStatus(), status),
printFixed("PASSED ")
testKill()
```

test/autotest.py
```diff
@@ -525,7 +527,9 @@ def testRestart():
       if i.endswith(".dmtcp"):
         cmd+= " "+ckptDir+"/"+i
     #run restart and test if it worked
-    procs.append(launch(cmd))
+    procs.append(runCmd(cmd))
+    #wait and see if process will die shortly after restart
+    sleep(S*SLOW)
```
karya0: Same story here. This `sleep` should go _after_ the following `WAITFOR` statement.

test/autotest.py
```diff
@@ -543,8 +547,9 @@ def testRestart():
 
     #start user programs
     for cmd in cmds:
-      procs.append(launch(BIN+"dmtcp_launch "+cmd))
-      sleep(S*SLOW)
+      procs.append(runCmd(BIN+"dmtcp_launch "+cmd))
+    #wait and see if process will die shortly after checkpoint
```
karya0: There seems to be a cut-paste error in the comment :-).

test/autotest.py
```diff
@@ -543,8 +547,9 @@ def testRestart():
...
+    sleep(S*SLOW)
```
karya0: Again, as with other `sleep` statements, this should go _after_ the following `WAITFOR` statement.

karya0:
Hi Gene: It looks like either I have still not understood the intended behavior, or my comments are really valid :-). In either case, I think it's good to clarify.

Here is my understanding: at various points in the ckpt-resume/restart sequence, we want to verify that the process continues to run _after_ we have verified the "unit-test" like functionality of individual actions (ckpt, restart, kill, etc.). If that is the case, then I would explicitly want to perform `sleep` _after_ the individual unit-test has been  verified to succeed and the computation is in the `RUNNING` state. Once we perform the `sleep`, we would then verify that the computation is still in `RUNNING` state.

Please let me know if I understood it incorrectly :-).

karya0:
PS: I would have liked if the function rename were in a separate commit ... but let's not worry about it for now :-).

test/autotest.py
gc00: The pseudo-code you propose would require moving CHECK(...) outside of testCheckpoint().  Currently, CHECK(...) is part of doing a testCheckpoint(), which fits better with the name testCheckpoint().

It's true that the WAITFOR() inside testCheckpoint() will timeout if it doesn't see the correct number of checkpoint files create.  So how about if I move sleep() until after the WAITFOR():
```
def testCheckpoint():
    coordinatorCmd(CKPT_CMD)
    WAITFOR(...getNumCkptFiles(ckptDir)>0 and...)
    sleep(S*SLOW)
    CHECK(doesStatusSatisfy(getStatus(), status),
```
This keeps close to the original logic, and preserves the meaning behind the name `testCheckpoint()`.

test/autotest.py
gc00: Agreed.  Note that in the previous line note above, I moved the `sleep()' to after the `WAITFOR()`.  I agree that we should do the same here.  Thanks for the correction.

test/autotest.py
gc00: Actually, the comment goes with the `sleep()` statement on the following line.  But here also, I should move the comment and the sleep statement lower, to just after the `WAITFOR()` statement.  But in that case, the `sleep` duplicates the purpose of the `sleep()` stmt found insides the `for i in range(CYCLES):` loop a little bit lower.

So, in this cae, I propose that we delete this particular `sleep()` stmt.

gc00:
@karya0: Prviously, i wasn't modifying the `WAITFOR()` statements.  The `WAITFOR()` has a timeout that is not affected by the `SLOW` variable.  I agree that the logic that you are porposing is cleaner.  But then to make that logic work, we also need to lengthen the timeout by using `SLOW` in the timeout interval.  In that case, it then works to move each `sleep()` statement to after the `WAITFOR()` statement.

To summarize, previously, my logic was:
```
DO_ACTION
sleep()
WAITFOR()
CHECK()
```

You are proposing:
```
DO_ACTION
WAITFOR()
sleep()
CHECK()
```
This works, but then we need to multiply the timeout of `WAITFOR()` by `SLOW`.  All of my inline comments above are made under the assumption that we're going to move the `SLOW` to after the `WAITFOR()`.  I agree that the newer logic is somewhat cleaner.

Let me know if you prefer the second form above, and I'll then change the logic to correspond, based on my comments in the inline notes, and based on multiplying the `WAITFOR()` timeout by `SLOW`.

gc00:
@karya0:  github does not allow us to place inline notes in the function definition of `WAITFOR()` because it is too many lines away from the other lines that were changed.  Nevertheless, despite the opinions of github, the change to `WAITFOR()` actually is needed or else the logic of a `SLOW` multiplier will not be correct.  So, I'm copying our `WAITFOR()` code here, and adding my `sleep()` statement, to get around the restrictions of github.
```
 #wait TIMEOUT for test() to be true, or throw error
 def WAITFOR(test, msg):
   left=TIMEOUT*SLOW*(S/DEFAULT_S)/INTERVAL
```
Note that the `SLOW` multiplier in the above line was added by me.  I have no way of marking this as an inline note.

test/autotest.py
karya0: I should have been clearer. What I meant was that we don't touch testCheckpoint()'s definition at all. The CHECK from my pseudocode was a new one.

The pseudocode that you pasted works as well. However, purely from an aesthetic point of view, I would prefer sleep/check outside the testCheckpoint() routine since this sleep/check are not there to test checkpoint, but to test whether the computation runs normally after a successful resume.

karya0:
@gc00 I didn't mean for WAITFOR to go away. Instead, I want it to remain untouched, i.e. we shouldn't need to modify the timeout interval.

Also, it looks like TIMEOUT itself is a function of SLOW, so I am not sure if we should do something about it.

In any case, looks like we are converging. Please feel free to commit it after the changes are there. I won't be able to look at it before next week.

gc00:
@karya0: First, WAITFOR will not be going away.  I only meant to indicate that I will modify WAITFOR to change:
    TIMEOUT -> TIMEOUT*SLOW
But alternatively, I don't need to modify WAITFOR.  (See below.)

You wrote that "it looks like TIMEOUT itself is a function of SLOW".  I assume that you're referring to the code:
```
  if i=="--slow":
    SLOW=5
    TIMEOUT *= 2
```
If you prefer, I could change this code to:
```
  if i=="--slow":
    SLOW=5
    TIMEOUT *= SLOW
```
and then leave the definition of WAITFOR() untouched.  Is this consistent with what you're proposing?
(I do want to make TIMEOUT be multiplied by SLOW, now that we're also working with slow CPUs like __arm__.)

gc00:
@karya0: I've now made the changes we discussed:
1.
The 'sllep()' functions now follows this pattern, as requested:
```DO_ACTION
WAITFOR()
sleep()
CHECK()
```
2.
No modification to WAITFOR().  Instead:
```
  if i=="--slow":
    SLOW *= 5
    TIMEOUT *= SLOW
```
3.
renaming fnc:  launchCmd -> runCmd (and launch -> run, in the comments)

test/autotest.py
```diff
@@ -509,8 +509,10 @@ def testCheckpoint():
 
     #wait for files to appear and status to return to original
     WAITFOR(lambda: getNumCkptFiles(ckptDir)>0 and \
-                    (CKPT_CMD == 'xc' or doesStatusSatisfy(getStatus(), status)),
+                   (CKPT_CMD == 'xc' or doesStatusSatisfy(getStatus(), status)),
             wfMsg("checkpoint error"))
+    #wait and see if process will die shortly after checkpoint
+    sleep(S*SLOW)
```
karya0: I hate to do this to you, but this sleep statement doesn't serve the intended purpose. (See my comment below for the updated fix.)

In it's current form, after sleeping for "S*SLOW" seconds, we do not check if the processes are still alive. The `CHECK(doesStatusSatisfy((numFiles,True),status) ...` only checks for a valid number of checkpoint files.

Therefore, we need to insert a `CHECK(doesStatusSatisfy(getStatus(), status))` right after this sleep to ensure that the processes are still alive after ckpt.

Further, as I suggested in one of my previous comments (I can't find it now), this newly added sleep can be removed from here and placed right after the call to testCheckpoint() later in the file.

test/autotest.py
```diff
@@ -556,6 +560,7 @@ def testRestart():
       printFixed("ckpt:")
       # NOTE:  If this faile, it will throw an exception to CheckFailed
       #  of this function:  testRestart
+      sleep(S*SLOW)
       testCheckpoint()
```
karya0: Let's put the following two statements right after this call to `testCheckpoint`:
```
sleep(S*SLOW)
CHECK(doesStatusSatisfy(getStatus(), status))
```

test/autotest.py
```diff
@@ -525,9 +527,11 @@ def testRestart():
       if i.endswith(".dmtcp"):
         cmd+= " "+ckptDir+"/"+i
     #run restart and test if it worked
-    procs.append(launch(cmd))
+    procs.append(runCmd(cmd))
     WAITFOR(lambda: doesStatusSatisfy(getStatus(), status),
             wfMsg("restart error"))
+    #wait and see if process will die shortly after restart
+    sleep(S*SLOW)
```
karya0: We need to insert `CHECK(doesStatusSatisfy(getStatus(), status))` after this sleep to make it effective.

gc00:
@karya0:  I've now pushed in a revision that addresses your comments.  In the case of your comment https://github.com/dmtcp/dmtcp/pull/107#discussion_r30597006, I chose to place the two extra statements that you propose at the end of `testCheckpoint()' itself, instead of after the return from that call, since it seemed to be part of the checkpoint test.

In this version, I made one other change.  For each of the newly introduced `sleep(S*SLOW)` statements, I put it inside an `if SLOW > 1:`.  This way, the normal autotest will continue to execute 'make check' as fast as usual, and only the `--slow` flag will cause extra pauses to verify that no part of checkpoint-restart/resume was omitted.  (Otherwise, we've inserted extra `sleep(S*SLOW)` statements which will cause `make check` to go slower even though `SLOW==1`.)

If you feel strongly about  `if SLOW > 1:`, let me know, and I can remove the "if" statement again.

Also, thanks for your careful reading of the logic, @karya0.  I was doing too many things at once, and I had started programming by template instead of thinking carefully about the logic.  You were right to slow me down.  Thanks.

karya0:
Let's merge it.

gc00:
Done.

Issue #102, opened 2015-05-09
__xstat/__xstat64 crashes when the second argument is an invalid address


rohgarg:
This was reported by Ankit Garg.

The __xstat() call returns -1 (with EFAULT), when the second argument is an invalid memory address. This can lead to a crash when running with DMTCP. In the current wrapper implementation, updateStatPath() is called on the second argument without checking for its validity. Here's a patch as suggested by Ankit:

```patch
diff --git a/src/plugin/ipc/file/filewrappers.cpp b/src/plugin/ipc/file/filewrappers.cpp
index aefc6e4..75bdaac 100644
--- a/src/plugin/ipc/file/filewrappers.cpp
+++ b/src/plugin/ipc/file/filewrappers.cpp
@@ -583,8 +583,14 @@ extern "C" int __xstat(int vers, const char *path, struct stat *buf)
  char tmpbuf [ PATH_MAX ] = {0} ;
  char *newpath = tmpbuf;
  DMTCP_PLUGIN_DISABLE_CKPT();
-  updateStatPath(path, &newpath);
-  int retval = _real_xstat(vers, newpath, buf);
+
+  int retval = _real_xstat(vers, path, buf);
+  if(retval == 0)
+  {
+    updateStatPath(path, &newpath);
+    if(newpath != path)
+      retval = _real_xstat(vers, newpath, buf);
+  }
  DMTCP_PLUGIN_ENABLE_CKPT();
  return retval;
}
@@ -594,8 +600,14 @@ extern "C" int __xstat64(int vers, const char *path, struct stat64 *buf)
  char tmpbuf [ PATH_MAX ] = {0};
  char *newpath = tmpbuf;
  DMTCP_PLUGIN_DISABLE_CKPT();
-  updateStatPath(path, &newpath);
-  int retval = _real_xstat64(vers, newpath, buf);
+
+  int retval = _real_xstat64(vers, path, buf);
+  if(retval == 0)
+  {
+    updateStatPath(path, &newpath);
+    if(newpath != path)
+      retval = _real_xstat64(vers, newpath, buf);
+  }
  DMTCP_PLUGIN_ENABLE_CKPT();
  return retval;
}
diff --git a/src/plugin/pid/pid_filewrappers.cpp b/src/plugin/pid/pid_filewrappers.cpp
index 7795e38..965ec55 100644
--- a/src/plugin/pid/pid_filewrappers.cpp
+++ b/src/plugin/pid/pid_filewrappers.cpp
[...]

PR #110, opened 2015-05-11
__xstat() survives bad mem addr

gc00:
In a user's program, stat() was being called with a bad memory address for pathname.  The combination of libc and the kernel then returned -1 with `errno == EFAULT`.
&nbsp;&nbsp;&nbsp;&nbsp;Under DMTCP, stat() went through a wrapper, which tried to read the memory at the address of the pathname.  Because of this it segfaulted.  Now, our utilities updateStatPath() and updateProcPathVirtualToReal() check if the memory address is readable before reading it.
&nbsp;&nbsp;&nbsp;&nbsp;A test program, stat.c, has been added that used to fail under DMTCP.  It now works.

(This pull request address issue #102.)

### DIFF
--- a/src/plugin/pid/pid_filewrappers.cpp
+++ b/src/plugin/pid/pid_filewrappers.cpp
@@ -156,8 +156,16 @@ extern "C" int __xstat(int vers, const char *path, struct stat *buf)
 {
   char tmpbuf[PATH_MAX];
   char *newpath = tmpbuf;
-  updateProcPathVirtualToReal(path, &newpath);
-  int retval = _real_xstat( vers, newpath, buf );
+  // See filewrapper.cpp:__xstat() for comments on this code.
+  int retval = _real_xstat(vers, path, buf);
+  if (retval == -1 && errno == EFAULT) {
+    // We're done.  Return.
+  } else {
+    updateProcPathVirtualToReal(path, &newpath);
+    if (newpath != path) {
+      retval = _real_xstat(vers, newpath, buf); 
+    } 
+  }
   return retval;
 }
 
@@ -165,8 +173,16 @@ extern "C" int __xstat64(int vers, const char *path, struct stat64 *buf)
 {
   char tmpbuf[PATH_MAX];
   char *newpath = tmpbuf;
-  updateProcPathVirtualToReal(path, &newpath);
-  int retval = _real_xstat64( vers, newpath, buf );
+  // See filewrapper.cpp:__xstat() for comments on this code.
+  int retval = _real_xstat64(vers, path, buf);
+  if (retval == -1 && errno == EFAULT) {
+    // We're done.  Return.
+  } else {
+    updateProcPathVirtualToReal(path, &newpath);
+    if (newpath != path) {
+      retval = _real_xstat64(vers, newpath, buf); 
+    } 
+  }
   return retval;
 }
 
@@ -188,8 +204,16 @@ extern "C" int __lxstat(int vers, const char *path, struct stat *buf)
 {
   char tmpbuf[PATH_MAX];
   char *newpath = tmpbuf;
-  updateProcPathVirtualToReal(path, &newpath);
-  int retval = _real_lxstat( vers, newpath, buf );
+  // See filewrapper.cpp:__xstat() for comments on this code.
+  int retval = _real_lxstat(vers, path, buf);
+  if (retval == -1 && errno == EFAULT) {
+    // We're done.  Return.
+  } else {
+    updateProcPathVirtualToReal(path, &newpath);
+    if (newpath != path) {
+      retval = _real_lxstat(vers, newpath, buf); 
+    } 
+  }
   return retval;
 }
 
@@ -197,8 +221,16 @@ extern "C" int __lxstat64(int vers, const char *path, struct stat64 *buf)
 {
   char tmpbuf[PATH_MAX];
   char *newpath = tmpbuf;
-  updateProcPathVirtualToReal(path, &newpath);
-  int retval = _real_lxstat64( vers, newpath, buf );
+  // See filewrapper.cpp:__xstat() for comments on this code.
+  int retval = _real_lxstat64(vers, path, buf);
+  if (retval == -1 && errno == EFAULT) {
+    // We're done.  Return.
+  } else {
+    updateProcPathVirtualToReal(path, &newpath);
+    if (newpath != path) {
+      retval = _real_lxstat64(vers, newpath, buf); 
+    } 
+  }
   return retval;
 }
 
--- a/test/stat.c
+++ b/test/stat.c
@@ -1,0 +1,47 @@
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <errno.h>
+
+#if defined(__x86_64__) || defined(__aarch64__)
+static const char * badAddress = (char *)0xffffffffff000000;
+#else
+static const char * badAddress = (char *)0xff000000;
+#endif
+
+int main() {
+  struct stat buf;
+  int rc;
+  while (1) {
+    char procFd[100];
+    snprintf(procFd, sizeof(procFd), "/proc/%d/fd", getpid());
+    procFd[sizeof(procFd)-1] = '\0';
+
+    rc = stat("/dev/tty", &buf);
+    if (rc == -1) return 1;
+    rc = stat("/etc/passwd", &buf);
+    if (rc == -1) return 2;
+    rc = stat(procFd, &buf);
+    if (rc == -1) return 3;
+    rc = stat("/etc/there_is_no_such_file_and_so_rc_is_error", &buf);
+    if (rc != -1 || errno != ENOENT) return 4;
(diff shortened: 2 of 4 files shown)


src/util_gen.cpp
```diff
@@ -470,6 +477,14 @@ bool Util::isValidFd(int fd)
   return _real_fcntl(fd, F_GETFL, 0) != -1;
 }
 
+// On failure, returns 0, with errno == EFAULT
+bool Util::isReadableMemoryAddress(const char *addr) {
+  // FIXME: We wanted to call libc:open(), but _real_open goes to second wrapper
+  int fd = _real_syscall(SYS_open, "/dev/null", O_WRONLY);
+  int rc = write(fd, addr, 1);
```
karya0: Instead of doing an explicit open followed by write, you can do the following:
```int rc = write(PROTECTED_STDERR_FD, addr, 0);
```

It would return `0` if successful, and `-1` with `errno` set to `EFAULT`. This would reduce the unnecessary overhead of performing an open (performing `open` has all sorts of side-effects and let's avoid if we can).

src/plugin/pid/pid_filewrappers.cpp
```diff
@@ -63,6 +63,10 @@ using namespace dmtcp;
 // FIXME:  This function needs third argument newpathsize, or assume PATH_MAX
 static void updateProcPathVirtualToReal(const char *path, char **newpath)
 {
+  if (! Util::isReadableMemoryAddress(path)) {
+    *newpath = const_cast<char *>(path);
+    return;
+  }
   if (Util::strStartsWith(path, PROC_PREFIX)) {
```
karya0: Wouldn't it be cleaner if we did the following:
```
 if (! Util::isReadableMemoryAddress(path) && 
    Util::strStartsWith(path, PROC_PREFIX)) {
```

src/util_gen.cpp
gc00: I did forget to close the fd.  Thanks for pointing that out.  I'll fix that in the next iteration.

As for a write of length 0, I'm worried that some kernels will notice that the write is length 0, and will return success (rc == 0) without testing whether addr is a valid memory address for reading.  Given that we have eventual goals of supporting Android, MacOSX, and just maybe Windows, this seems dangerous.

karya0: Doing a write of "0" bytes is not dangerous. It's a well known trick and is used quite generously.  The POSIX standard documents this behavior too. I would strongly suggest to use this instead of doing an open and close.

src/plugin/pid/pid_filewrappers.cpp
gc00: I'm worried about keeping the intention of the code clear.  Right now, it's
```if (! Util::isReadableMemoryAddress(path)) { ...}
    else if (CaseA) {...}
    else {...}```
The logic of how to add CaseB is clear.  But if we write:
```if (CaseA && ! Util::isReadableMemoryAddress(path)) {...}
    else {...}```
will it be obvious for a naive programmer that
``` ! Util::isReadableMemoryAddress(path)```
must be added to every new case?  (I'm being a little paranoid, because much of the DMTCP code will continue to be written by new students, and so I want to keep the code "student-safe".  :-)

File: src/util_gen.cpp (hunk 1 above)
Comment (gc00) [inline src/util_gen.cpp]: I looked it up, and I'm getting a different reading of the POSIX standard.  I'm seeing:
    _Implementations are allowed, but not required, to perform error checking for write() requests of zero bytes._
at:  http://pubs.opengroup.org/onlinepubs/009695399/functions/write.html
This seems to say that a write of zero bytes from a bad memory address can succeed.

Comment (gc00) [inline src/util_gen.cpp]: I looked it up, and I'm getting a different reading of the POSIX standard.  I'm seeing:
    _Implementations are allowed, but not required, to perform error checking for write() requests of zero bytes._
at:  http://pubs.opengroup.org/onlinepubs/009695399/functions/write.html
This seems to say that a write of zero bytes from a bad memory address can succeed.

File: test/stat.c [hunk 3]
```diff
@@ -0,0 +1,33 @@
...
+    if (rc == -1) return 2;
+    rc = stat(procFd, &buf);
+    if (rc == -1) return 3;
+    rc = stat("/etc/there_is_no_such_file_and_so_rc_is_error", &buf);
+    if (rc != -1 || errno != ENOENT) {
+      printf("Expected 'rc == -1 && errno == ENOENT'; got rc: %d, errno: %d\n",
+             rc, errno);
+      return 4;
+    }
+    char * badAddress = (char *)0xffffffffff000000;
```
gc00: I made a mistake here.  The definition of `badAddress` should be:

    #if defined(__x86_64__) || defined(__aarch64__)
    const char * badAddress = (char *)0xffffffffff000000;
    #else
    const char * badAddress = (char *)0xff000000;
    #endif```
I'll fix this in the next iteration.

src/util_misc.cpp
```diff
@@ -470,6 +477,15 @@ bool Util::isValidFd(int fd)
   return _real_fcntl(fd, F_GETFL, 0) != -1;
 }
 
+// On failure, returns 0, with errno == EFAULT
+bool Util::isReadableMemoryAddress(const char *addr) {
+  // FIXME: We wanted to call libc:open(), but _real_open goes to second wrapper
+  int fd = _real_syscall(SYS_open, "/dev/null", O_WRONLY);
+  int rc = write(fd, addr, 1);
```
gc00: I'm copying the history for this line.  It was lost in the new revision to the code.  @karya0 had suggested a shorter version of the code, that avoids the call to open():
`int rc = write(PROTECTED_STDERR_FD, addr, 0)`

@karya0 wrote:
Doing a write of "0" bytes is not dangerous. It's a well known trick and is used quite generously. The POSIX standard documents this behavior too. I would strongly suggest to use this instead of doing an open and close.

@gc00 replied:
I looked it up, and I'm getting a different reading of the POSIX standard. I'm seeing:
Implementations are allowed, but not required, to perform error checking for write() requests of zero bytes.
at: http://pubs.opengroup.org/onlinepubs/009695399/functions/write.html
This seems to say that a write of zero bytes from a bad memory address can succeed.

@gc00 also added the statement `close(fd);', (below) which he had forgotten in the earlier revision.

karya0:
Looks like there is some disagreement between the posix manpages installed on my machine and the link you posted. My installed version could be outdated.

From the webpage, it's clear that an OS can decide to not do the error check for zero byte write. However, for all practical purposes (i.e. Linux/BSD), that assumption seems valid.  Also, I am still uncomfortable with doing `open` and `close` within a plugin for the sake of error check and would like to avoid it if possible. My general understanding is that these things cause more trouble than they are worth. Having said that, if you feel strongly about this, let's keep it as is but let's also put a comment near the calls to `open` and `close` to find a better way for doing address check.

gc00:
@karya0 wrote:
     _I am still uncomfortable with doing open and close within a plugin for the sake of error check ..._
This is understandable.  On the other hand,
    _I (@gc00) am uncomfortable about diverging from the POSIX standard and making narrower assumptions based on Linux/BSD.  :-(_
Is there something we can do to make the alternative design choices clear for future maintainers?  For example, one could imagine a `#ifdef WRAPPER_USES_OPEN ... #endif` around the code in question, and somehow conditionalize out the part of `test/stat.c` that tests this.  Possibly, we could even put
    `#define WRAPPER_USES_OPEN 1`
inside config.h.in, to make the design choice really stand out.
&nbsp;&nbsp;&nbsp;&nbsp;Any thoughts on this?

karya0:
I still don't like the open/close calls within the xstat wrapper. Specially since I think that we would probably need to extend this solution to a lot of wrappers. How about calling xstat with the given arguments instead and checking for errno. If it's EFAULT, then we return EFAULT, or else we continue with the remaining logic?

We have been using double-calling in a few wrappers. This is not a perfect solution but a preferred one.

Ideally, I would very much prefer the write() trick, even if it's non-POSIX because this is lightweight and works on Linux/BSD (our primary targets for now). Some other OSes (Windows) provide a better way to detect invalid pointers, so we can put it undef #ifdef along with other non-POSIX stuff when the time comes :).

gc00:
@karya0, I've now changed the pull request for the xstat wrappers.  The new version of the pull request now follows your suggestion.

> How about calling xstat with the given arguments instead and checking for errno. If it's EFAULT,
> then we return EFAULT, or else we continue with the remaining logic?

(Sorry it took me a while to get back to this pull request.)

Anyway, because essentially the same code has been added to the bodies _xstat, _xstat64, _lxstat, _lxstat64, we could consider a macro in the body.  This occurs both inside `filewrappers.cpp` and `pid_filewrappers.cpp`.  This would avoid having to duplicate code.  I'm not sure if the code is better as is, or with the use of macros.

gc00:
@karya0 and I talked.  We agreed to ship this.

ankitcse07:
When will this fix be part of  master branch

gc00:
@ankitcse07, the fix for xstat has been pushed in now.  I've also pushed in a test to check for this case in the future.  Hopefully, that will avoid regressions.

Thanks for your patience.  Sorry it took so long.

PR #111, opened 2015-05-11
Rename: util_gen.cpp -> util_misc.cpp

gc00:
In the past, I had always wondered what `util_gen.cpp` stood for.  (E.g., util_generate.cpp ?).  I realized today that it was util_general.cpp.  Probably I should have figured it out earlier.  But anyway, `util_misc.cpp` seems like a more precise name for the intended theme.


karya0:
LGTM.

PR #114, opened 2015-05-13
JALIB: Generate coredumps on JASSERT

rohgarg:
JASSERT doesn't generate a coredump by default. This patch
introduces a new DMTCP environment variable:
  DMTCP_ABORT_ON_FAILED_ASSERT.
Exporting this variable will cause JASSERT to call abort()
to generate a coredump. This can be useful when debugging
distributed applications.

### DIFF
--- a/doc/debugging-dmtcp.txt
+++ b/doc/debugging-dmtcp.txt
@@ -23,6 +23,20 @@ the cause, try:
 
 ===
 
+JASSERT doesn't generate a core dump by default (because it calls _exit()).
+Exporting the environment variable DMTCP_ABORT_ON_FAILED_ASSERT will cause
+it to call abort(). This can be useful when debugging distributed processes.
+Example:
+  DMTCP_ABORT_ON_FAILED_ASSERT=1 bin/dmtcp_launch a.out
+  gdb a.out core
+
+When used with dmtcp_restart, it is important to embed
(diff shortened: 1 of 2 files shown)


doc/debugging-dmtcp.txt
```diff
@@ -23,6 +23,15 @@ the cause, try:
 
 ===
 
+JASSERT doesn't generate a coredump by default (because it calls _exit()).
```
gc00: Perhaps:  "coredump" -> "core dump"

doc/debugging-dmtcp.txt
```diff
@@ -23,6 +23,15 @@ the cause, try:
...
+Exporting the environment variable DMTCP_ABORT_ON_FAILED_ASSERT will cause
+it to call abort(). This can be useful when debugging distributed processes.
+Example:
+  DMTCP_ABORT_ON_FAILED_ASSERT=1 bin/dmtcp_launch a.out
+  gdb -c core
```
gc00: 1. Let's change "gdb -c core" to "gdb a.out core".  It's safer, especially if this was a restarted process.
2. Then, let's add a further statement -- something like:

    When used with dmtcp_restart, it is important to embed DMTCP_ABORT_ON_FAILED_ASSERT
    in the checkpoint image by first using it with dmtcp_launch, as above.  After that, a
    restarted process will also generate a core dump.```

karya0:
LGTM. Please squash the commits and push.

PR #118, opened 2015-05-16
Simplify dmtcp_restart (no need for wrapper fncs or _real_XXX())

gc00:
This pull request includes two commits.  The first and larger one moves four functions from ckptserializer.cpp to dmtcp_restart.cpp:
    first_char(), open_ckpt_to_read(), CkptSerializer::openCkptFileToRead((), CkptSerializer::readCkptHeader()
They all become static functions in dmtcp_restart.cpp, since no other file refers to them.  (And of course, they are no longer methods of the CkptSerializer class.  This also makes logical sense, since the name ckpserializer.cpp refers to serializing files.  The files that were moved are used to inspec the already created checkpoint image file on restart, and so they are conceptually different.
&nbsp;&nbsp;&nbsp;&nbsp;And finally, the biggest win is that the four functions included about 12 occurrences of `_real_XXX()`.  The dmtcp_restart executable has no need for wrapper functions, and can now just call `XXX()` instead of `_real_XXX()`.

As for the second commit, this is a one-line change in ckptserializer.cpp:
```
-    sigsuspend(&suspend_sigset);
+    _real_sigsuspend(&suspend_sigset);
```
Our `sigsuspend()` wrapper ensured that SIGUSR2 was never blocked.  But in this usage, we are waiting for the gzip child process to finish during a checkpoint.  So, we really did want to block SIGUSR2 until the gzip from the previous checkpoint is done.  Using `_real_siguspend()` allows that to happen now.
&nbsp;&nbsp;&nbsp;&nbsp;The reason that this commit is bundled with the earlier one is that this commit becomes possible only once we moved the ckptserializer functions into dmtcp_restart.  Otherwise, there was a linker error in linking dmtcp_restart.o with ckptserializer.o, since neither of those defined _real_sigsuspend().

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -83,7 +83,7 @@ __targetdir__bin_dmtcp_coordinator_SOURCES = dmtcp_coordinator.cpp lookup_servic
 
 __targetdir__bin_dmtcp_nocheckpoint_SOURCES = dmtcp_nocheckpoint.c
 
-__targetdir__bin_dmtcp_restart_SOURCES = dmtcp_restart.cpp util_exec.cpp ckptserializer.cpp
+__targetdir__bin_dmtcp_restart_SOURCES = dmtcp_restart.cpp util_exec.cpp
 
 __targetdir__bin_dmtcp_command_SOURCES = dmtcp_command.cpp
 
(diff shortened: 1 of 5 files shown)


karya0:
LGTM modulo one comments on the sigsuspend commit.

gc00:
What was the scenario where the 2nd ckpt was getting triggered?

I don't thing there was a 2nd ckpt involved in sigsuspend.  The issue is described by the comment from the code above the place where the call to sigsuspend is modified:

    *  So, we use sigsuspend so that we can clean up the child zombie
    *  with wait4 whether it already terminated or not yet.

The comment shows that the intention was that sigsuspend should suspend until the SIGCHLD comes through.  But this was not actually happening, because we were using `sigsuspend` instead of `_real_sigsuspend`.

karya0:
LGTM.

PR #120, opened 2015-05-18
fileconnection: use flock around createDirectoryStructure

rohgarg:
This patch is a workaround for what is arguably a bug in the Lustre
FS. See: https://jira.hpdd.intel.com/browse/LU-6528.  When multiple
independent processes call mkdir() on an existing directory,
a process can sometimes get an EACCES error, which causes our
createDirectoryStructure function to assert. This patch adds flocks
around calls to createDirectoryStructure to ensure that only one
process calls mkdir().

This is still a WIP; putting it up for review/feedback.

### DIFF
--- a/src/ckptserializer.cpp
+++ b/src/ckptserializer.cpp
@@ -545,7 +545,7 @@ void CkptSerializer::writeCkptImage(void *mtcpHdr, size_t mtcpHdrLen)
   string tempCkptFilename = ckptFilename;
   tempCkptFilename += ".temp";
 
-  JTRACE("Thread performing checkpoint.") (gettid());
+  JTRACE("Thread performing checkpoint.") (dmtcp_gettid());
   createCkptDir();
   forked_ckpt_status = test_and_prepare_for_forked_ckpt();
   if (forked_ckpt_status == FORKED_CKPT_PARENT) {
--- a/src/nosyscallsreal.c
+++ b/src/nosyscallsreal.c
@@ -375,24 +375,14 @@ long _real_syscall(long sys_num, ...) {
                                                arg[5], arg[6]);
 }
 
-LIB_PRIVATE pid_t gettid() {
+LIB_PRIVATE pid_t dmtcp_gettid() {
   return syscall(SYS_gettid);
 }
-LIB_PRIVATE int tkill(int tid, int sig) {
+LIB_PRIVATE int dmtcp_tkill(int tid, int sig) {
   return syscall(SYS_tkill, tid, sig);
 }
-LIB_PRIVATE int tgkill(int tgid, int tid, int sig) {
+LIB_PRIVATE int dmtcp_tgkill(int tgid, int tid, int sig) {
   return syscall(SYS_tgkill, tgid, tid, sig);
-}
-
-// gettid / tkill / tgkill are not defined in libc.
-pid_t _real_gettid(void) {
-  REAL_FUNC_PASSTHROUGH_PID_T (syscall(SYS_gettid));
-}
-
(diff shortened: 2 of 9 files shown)


gc00:
1.  I see that DMTCP provides:  `Util::lockFile` and `Util::unlockFile` in util_misc.cpp.  It's worth testing if we can use  this instead of directly using flock().
In my next comment (point 2), I want to address this also to:  @karya0.

rohgarg:
The downside of using `fcntl(F_SET_LK)` is that it is process-specific and isn't thread-safe. `flock()`, on the other hand, is bound to a file-descriptor, but isn't as portable.

gc00:
@karya0:
    In order to work around a bug in Lustre, we need to use a file from the shared filesystem.  Here's a proposal.  You probably know more about this.  Could you improve on this?

2.  Second, as I'm sure @rohgarg already know, we can't hardwire in the directory `/work/...` in the code.  The issue is that we need to know a directory on a shared filesystem.  In HPC batch jobs, there will almost always be such a shared directory.     I assume that we cannot assume that $DMTCP_TMPDIR points to a shared directory, for example in the SLURM/ibrun environment of 

@karya0:  In the design of DMTCP, is there a place where we can guarantee that there is a shared filesystem?  For example, can we use the directory into which we write checkpoint images?  (Note that if the directory name changes between ckpt and restart, then this code must re-compute the directory name each time, while preventing checkpoints in this critical section.

@rohgarg:  Alternatively, notice that the routine `createDirectoryStructure(savedFilePath);` is already using a shared file, `savedFilePath`.  A routine could open each component subdirectory, beginning with `/`, until it finds the first directory for which it has permission to set locks.  Can each process discover this unique prefix of sharedFilePath, and then open it with a fd, and then use `flock()` (or `Util::lockFile()`)?  In this case, if the processes are not sharing the directory, then there is only one process doing an `flock()` on that file, and everything works.  If the processes _are_ sharing the directory, then the processes execute `flock()` to get an exclusive lock on the file, and then everything works again.

I'm just brainstorming here.  But would something like this work?

rohgarg:
@gc00: An alternative to file-locking is to try `mkdir()` or `open(O_RDONLY)` again on a directory for which a call to `mkdir()` returned `EACCES` earlier. `open()` is probably the safer option of the two.
```
 ret = mkdir(dir);
 if ret == EACCES:
   ASSERT (open(dir, O_RDONLY) > 0);
   close(dir);
   continue;
 else if ret == EEXIST:
   continue;
 else:
   ASSERT(False);
```

gc00:
@rohgarg:  I like your solution a lot.  The bug report on Lustre seems to apply only to `mkdir`.  That makes sense, since `mkdir` requires some kind of two-phase commit.  So, the call to `open` ought to be safe.
Let's try out the code that you displayed above.  Let's also add a comment just before the `ASSERT (open(dir, O_RDONLY) > 0);`, which refers to the URL of the Lustre mkdir bug, and explains that this is a workaround to avoid the bug in mkdir that will sometimes erroneously return EACCESS.

rohgarg:
@gc00: I'll try it today and check if it works. @karya0 and I were discussing this earlier when this came up as an alternative.

(And yes, the implementation will use `errno`. :-) )

gc00:
@rohgarg:  Great!  If this architecture works, this will be a big win over any other possible design.  It depends on Lustre doing the right thing with `open()`, but almost surely Lustre will do the right thing.

karya0:
As discussed earlier, I would rather have the open()-after-EACCES approach than using file locks.

PR #123, opened 2015-06-02
fileconnection: add check to verify the existence of a dir

rohgarg:
This patch is a workaround for what is arguably a bug in the Luster
FS. See: https://jira.hpdd.intel.com/browse/LU-6528.  When multiple
independent processes call mkdir() on an existing directory,
a process can sometimes get an EACCES error, which causes our
createDirectoryStructure function to assert. This patch adds an
extra check (using the stat() call) in createDirectoryStructure in
case the call to mkdir() returned EACCES.

### DIFF
--- a/src/plugin/ipc/file/fileconnection.cpp
+++ b/src/plugin/ipc/file/fileconnection.cpp
@@ -886,9 +886,26 @@ static void createDirectoryStructure(const string& path)
     if (index > 1) {
       string dirName = path.substr(0, index);
 
+      errno = 0;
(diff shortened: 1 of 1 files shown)


gc00:
@rohgarg: Does this replace pull request #120, or is it in addiont?

rohgarg:
Sorry I forgot to mention, this replaces pull request #120.

rohgarg:
@gc00, @karya0: Ping. :-)

karya0:
LGTM.

PR #124, opened 2015-06-02
Rsh execp commit 02ccbb8

ankitcse07:
fix #124  is trying to avoid to do anything with /bin/rsh executable. Currently as rsh executable has setid bit set so dmtcp infrastructure copies the executable locally and then execute the same from local copy, but this copy of rsh will not work if rsh is provided some command to execute on remote host. We have lots of such usages in our code, in which rsh command is invoked with some process on remote host but these process are short lived and are not required to be checkpointed and hence not required at restart time.  I can understand the rsh is not supported by dmtcp but this fix is just trying to avoid the above situations for which dmtcp_launch currently results in error for such rsh commands. 

In order to reproduce simply run following test under dmtcp_launch  and you will see rsh to fail

int main(int argc, char *argv[])
{
    pid_t pid;

    pid = fork();

    if(pid == 0)
    {
       char* argv[10];
       argv[0] = "/usr/bin/rsh";
       argv[1] =  "<remote host name";
       argv[2] =   "date";
       argv[3] = NULL;

       execvp("rsh", argv);
    }
    else
    {
        int i = 0;
        while(1)
        {
           
            sleep(1);
            printf("Waiting %d\n", i++);
        }
        waitpid(pid, NULL,  WUNTRACED | WCONTINUED);
    }
    printf("Hello\n");
}


gc00:
@rohgarg: I'm just curious.  What was the use case?  What was the type of software that still uses 'rsh'?

ankitcse07:
Our software still uses rsh internally to launch certain processes on remote hosts and moving to ssh is really not possible.  dmtcp_launch cannot successfully execute rsh commands invoked internally by the process being tracked by dmtcp so I did a small fix in my repository and was able to run my software using above fix. Being short-lived these won't be present at the time of checkpoint

ankitcse07:
I am closing this request as I had added support for rsh in my development sandbox.  I can create another pull request for support of rsh if you guys can review it

Issue #109, opened 2015-05-11
dmtcp_launch silently joins old coordinator if present


gc00:
In one window, type:  `dmtcp_coordinator`
In another window, type: `dmtcp_launch a.out`
The command `dmtcp_launch` (and similarly `dmtcp_restart`) will silently join the existing coordinator.  In the early days of DMTCP, this seemed like a nice convenience feature.  Now that DMTCP is more mature, this behavior seems dangerous.  At the very least, we should issue a JNOTE to warn when this happens.  But a still better policy would be to fail with an error message, unless the user had called it as: `dmtcp_launch --join a.out`

Any thoughts about this?

karya0:
I would prefer the following scheme:

1. By default, do not join an existing computation -- if the coordinator is currently serving an existing computation, dmtcp_launch should specify --join to explicitly join an existing computation.

2. If the coordinator is currently idle, connect to it even if no --join was specified.

3. With (1) and (2), nothing needs to be done for dmtcp_restart. Currently, dmtcp_restart can join an existing (restarting) computation only if they belong to the same computation group.

If we decide to go with this scheme (or any other scheme), we need to perform a deprecation cycle to avoid unnecessary hiccups for our current users.

gc00:
@karya0: I'd like to propose a modified rule 2.b instead of rule&nbsp;2 in the above schema:

2.b. If the coordinator is currently idle, _and if it is on the local host_, _and if the coordinator process is owned by the same user_, then connect to it even if no --join was specified.

What do you think?

PR #125, opened 2015-06-03
documentation: differentiate between ^C and kill

rohgarg:


gc00:
A userr (on dmtcp-forum?) had used ^C instead of 'dmtcp_command -k' to kill a job with more than one process.  I believe the story is that some of the processes continued to live, and so the DMTCP coordinator did not exit.  After that, connecting to the default DMTCP coordinator did not work as expected.  We need to document that ^C is not as reliable as `dmtcp_command --kill`.

gc00:
@karya0 and @rohgarg, I was the one who asked to make this a pull request, rather than an issue.  I think it's very important for us to support less experienced users who are confused by things like this.

In my view, this is an example of issue #109.  Hence, I propose to write code in the DMTCP coordinator that will notify inexperienced users when they accidentally launch a new process that silently joins the coordinator.

I propose to write the fairly trivial code to solve this.  But I'd like to get my code reviewed before I write it.  :-)

**PROPOSED CODE FOR REVIEW:**  Suppose that dmtcp_coordinator is already running with one process.  I propose that if a new request comes to the coordinator via a second `dmtcp_launch` (not via `fork()` or exec()`), then we will print a second header.  We will respect the `--quiet` flag for users who don't like this.  Note that when using MPI, it can happen for some resource manager, that many separate processes call `dmtcp_launch` separately.  But in these cases, the HPC environment should already be siphoning off the output of dmtcp_coordinator or using `--quiet`.  So, the proposed new functionality should not interfere with MPI situation.

Can I get a pre-review before I write this code?  Thanks.

gc00:
@karya0 and I talked.  We'll do a more ambitious change.  The user must type ``--join`` in order to join an existing computation.  If there is no current computation, ``--join`` will fail.  If the user typed ``--join``, then we don't need to re-print the launch header.  This will also fix the problem of a user joining an existing computation without realizing it.  I'll change this pull request later.

karya0:
Let's merge it.

karya0:
Merged.

Issue #126, opened 2015-06-03
restart fails for `stress`


rohgarg:
Here are the steps to reproduce:
```Bash
$ dmtcp_launch stress --vm 2 --vm-bytes 128M --timeout 30s
$ dmtcp_command -cb
$ dmtcp_command -k
$ ./dmtcp_restart_script.sh
```
Restart gets stuck and the program never finishes.

gc00:
@rohgarg: Could you give us a pointer to this program `stress`?

rohgarg:
@gc00: I looked at the `stress` program's source. It's basically an infinite loop and the only way out of it is a time-out. Unfortunately, it uses `alarm()` to set up time-out. The program is restarted correctly but appears to be "stuck" on restart because DMTCP fails to restore the alarm. This is a known issue: #106.

rohgarg:
Fixed via https://github.com/dmtcp/dmtcp/commit/29d7e3c473724ea4aebe2c0a284225cf2ac0b0a4

Issue #106, opened 2015-05-11
Call to alarm() is lost on restart


gc00:
For the following program:

    #include <unistd.h>
    int main() {
      alarm(15);
      while (1);
      return 0;}
Try:

    dmtcp_launch -i3 ./a.out
    dmtcp_restart ckpt_a.out_*.dmtcp
and wait 15 seconds to see if the alarm will kill the process.  It doesn't.

The timer plugin works.  So, why not this?

rohgarg:
Closed via https://github.com/dmtcp/dmtcp/commit/29d7e3c473724ea4aebe2c0a284225cf2ac0b0a4

PR #127, opened 2015-06-04
alarm-plugin: add plugin to checkpoint/restart alarm()

rohgarg:
This addresses issues: #126 and #106.


rohgarg:
@gc00, @karya0: Any comments?

contrib/alarm/alarm.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <unistd.h>
+#include <stdio.h>
+
+#include "dmtcp.h"
+#include "jassert.h"
+
+#ifdef ALARM_PLUGIN_DEBUG
+# undef JTRACE
+# define JTRACE JNOTE
+#endif
```
karya0: Instead of this block here, you could just do `#define DEBUG` before including jassert.h. That would allow you to see JTRACE output on console (there won't be any jassertlogs anyways).

contrib/alarm/alarm.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#define _next_alarm NEXT_FNC(alarm)
```
karya0: Let's use _real_XXX notion. The reason for not doing a _next_ is that from the plugin's point of view, it's calling the "real" function. Also, we still use _real_ else where in the repo, so if we ever decide to replace _real_ with _next_, we can do a bulk update.

contrib/alarm/alarm.cpp
```diff
@@ -0,0 +1,88 @@
...
+/* File local functions  */
+static unsigned
+start_stop_alarm(unsigned int timeout)
+{
+  return alarm(timeout);
+}
+
+/* Wrapper to determine if the application is using alarm */
+extern "C" unsigned int
+alarm(unsigned int seconds)
```
karya0: Do we really care about the alarm wrapper? Why not just do a `timeleft = alarm(0)` during pre-ckpt and then `alarm(timeleft)` during restart if `timeleft` is greater than zero?

karya0:
A couple of high level comments:
1. The license needs to be changed to LGPL. GPL is contagious :).
2. Why not convert it into an in-built plugin (src/alarm.cpp) so that it is enabled by default. I think this is one of those things that should be available to ordinary users. I understand that there will be issue related to the "timeleft" notion, but for the majority of our users, if we put the alarm without worrying about the time lapse between ckpt and restart, it would be okay. Any comments?

contrib/alarm/alarm.cpp
rohgarg: Sure. Will do.

contrib/alarm/alarm.cpp
rohgarg: Sounds good.

contrib/alarm/alarm.cpp
rohgarg: I agree. The only reason I added this was to prevent a spurious alarm on resume/restart.

rohgarg:
> The license needs to be changed to LGPL

Strange! I didn't realize this. I just copied it from some file in a plugin or DMTCP src. I'll fix this.

> Why not convert it into an in-built plugin

I agree. This was just a first draft. I wanted to put something together without polluting the DMTCP src.

> there will be issue related to the "timeleft" notion

I'm not sure I understand. Could you please elaborate?

karya0:
I just meant that the notion of "remaining time" gets complicated after a restart. Along the same lines as if you had 5 seconds remaining for the sleep at ckpt, and you restart after a day, do you sleep for 5 seconds or 0 seconds.

In any case, we don't have to worry about it for now.

rohgarg:
@karya0: Do you think it should be integrated with the timer plugin? Or libdmtcp?

karya0:
I would create a new file src/alarm.cpp and model it after procname.cpp or terminal.cpp. Basically, you want to rename `dmtcp_event_hook` to `dmtcp_Alarm_EventHook` and remove the `NEXT_FNC` definition from it. That should mostly do it.

karya0:
PS: Also, you'll have to modify `void DmtcpWorker::eventHook()` to install the hook.

rohgarg:
@karya0: I have updated the patch based on your inputs. Please let me know how this looks now.

test/alarm.c
```diff
@@ -0,0 +1,19 @@
...
+{
+  exit(EXIT_SUCCESS);
+}
+
+int
+main()
+{
+  signal(SIGALRM, handle_alarm);
+  alarm(15);
+  while (1);
```
karya0: Sorry for not noticing it sooner. Can you change it to `while (1) sleep(1);`? This would avoid cpu spinning.

src/alarm.cpp
```diff
@@ -0,0 +1,58 @@
...
+        if (l_timeleft > 0) {
+          JTRACE("*** Resuming alarm. ***") (l_timeleft);
+          l_timeleft = start_stop_alarm(l_timeleft);
+        }
+        break;
+      }
+    default:
+      break;
+  }
+}
```
karya0: Now that this file is in src/, may be we should start using the camelCase style and drop the `l_` prefix from variable names :-). Also, is there a reason for a separate `start_stop_alarm` now? Seems unnecessary to me.

rohgarg:
Ping. :-)

karya0:
Ship it!

rohgarg:
Closed via https://github.com/dmtcp/dmtcp/commit/29d7e3c473724ea4aebe2c0a284225cf2ac0b0a4

PR #128, opened 2015-06-11
modify-env-plugin: fix Makefile and test

rohgarg:
This patch fixes the broken `make check` for the modify-env plugin.


karya0:
ship it

PR #132, opened 2015-06-23
socketconnection: add PMI socket to blacklisted sockets

rohgarg:
This is a workaround for the random ckpt bug at Stampede.

### DIFF
--- a/src/plugin/ipc/socket/socketconnection.cpp
+++ b/src/plugin/ipc/socket/socketconnection.cpp
@@ -202,6 +202,20 @@ TcpConnection& TcpConnection::asTcp()
   return *this;
 }
 
+#ifdef MPISPAWN_WORKAROUND
+static int
+getMPISpawnPortNum(const char* envVar)
+{
+  /* PMI_PORT is of the form: "hostname:port" */
+  char *temp = getenv(envVar);
+  if (temp) {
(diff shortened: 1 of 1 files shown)


src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -221,6 +235,15 @@ bool TcpConnection::isBlacklistedTcp(const sockaddr* saddr, socklen_t len)
     int blacklistedRemotePorts[] = {53,                 // DNS Server
                                     389, 636,           // LDAP
                                     -1};
+#ifdef MPISPAWN_WORKAROUND
```
karya0: May be a more readable design here would be to replace the int array with a vector and then just do a push_back to the vector inside this #ifdef. What do you think?

karya0:
LGTM modulo one small style comment. Feel free to ignore it :).

src/plugin/ipc/socket/socketconnection.cpp
rohgarg: Sure. That sounds better.

rohgarg: Thinking again, it'll be a performance overhead unless we can move it to a global constructor. The cost is minor but can quickly add up for an application that uses many sockets.

I'll keep it as is for now.

PR #138, opened 2015-07-09
Fixes #45: Bug fix for sem_wait and glibc >= 2.21.

### LINKED ISSUE #45: make check fails for pthread2 and pthread3 with newer libc (2.21)
The failure happens after the process has been successfully restarted.  It eventually segfaults right after resuming from a restart. Here is the backtrace for the segfaulting thread:

```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00007f37ba7a1112 in unwind_stop (version=<optimized out>, actions=<optimized out>, 
    exc_class=<optimized out>, exc_obj=<optimized out>, context=<optimized out>, 
    stop_parameter=<optimized out>) at unwind.c:82
#2  0x00007f37b9966512 in ?? () from /lib64/libgcc_s.so.1
#3  0x00007f37b9966894 in _Unwind_ForcedUnwind () from /lib64/libgcc_s.so.1
#4  0x00007f37ba7a1180 in __GI___pthread_unwind (buf=<optimized out>) at unwind.c:126
#5  0x00007f37ba79b565 in __do_cancel () at pthreadP.h:283
#6  __pthread_exit (value=<optimized out>) at pthread_exit.c:28
#7  0x00007f37bac26ffe in _real_pthread_exit (retval=0x0) at syscallsreal.c:967
#8  0x00007f37babf43f3 in pthread_exit (retval=0x0) at threadwrappers.cpp:235
#9  0x0000000000400cee in threadMain (data=0x1b43d40) at pthread2.c:69
#10 0x00007f37babf40cc in pthread_start (arg=<optimized out>) at threadwrappers.cpp:159
#11 0x00007f37ba79a4b4 in start_thread (arg=0x7f37b7e12700) at pthread_create.c:333
#12 0x00007f37babf3f29 in clone_start (arg=0x7f37bb8de808) at threadwrappers.cpp:68
#13 0x00007f37ba4d8a4d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
(gdb) 
```

karya0:
In glibc >= 2.21, sem_wait registers a cleanup handler before blocking
for the semaphore. The handler is popped before returning from it.
However, if we use it in stopthisthread(), the cleanup handler is not
popped up on restart because, we do a longjmp. Thus if a thread, that was
created before ckpt, were to do pthread_exit, the pthread_exit routine
will pop all pending cleanup handlers and try to execute them. At this
point the stale sem_wait() handler results in SEGFAULT (since the
semaphore is invalid now).

This was most noticeable on pthread2/3 tests since they have threads that
are created before ckpt and removed after restart.

Here is the glibc commit: https://sourceware.org/git/?p=glibc.git;a=commit;h=042e1521c794a945edc43b5bfa7e69ad70420524

### DIFF
--- a/src/threadlist.cpp
+++ b/src/threadlist.cpp
@@ -54,6 +54,8 @@ static Thread *threads_freelist = NULL;
 static pthread_mutex_t threadlistLock = PTHREAD_MUTEX_INITIALIZER;
 static pthread_mutex_t threadStateLock = PTHREAD_MUTEX_INITIALIZER;
 
+static pthread_rwlock_t *threadResumeLock = NULL;
+
 static __thread Thread *curThread = NULL;
 static Thread *ckptThread = NULL;
 static int numUserThreads = 0;
@@ -65,7 +67,6 @@ static sem_t semWaitForCkptThreadSignal;
 
 static void *checkpointhread (void *dummy);
 static void suspendThreads();
-static void resumeThreads();
 static void stopthisthread(int sig);
 static int restarthread(void *threadv);
 static int Thread_UpdateState(Thread *th,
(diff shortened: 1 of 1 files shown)


jiajuncao:
LGTM.

Issue #139, opened 2015-07-09
Do not store return value of string::c_str().


karya0:
In the source code, sometimes we are taking a pointer to the char* via string::c_str(). The string might move (if it were to grow, etc.) and would cause the pointer to be invalid.

PR #140, opened 2015-07-09
Do not store return value of string::c_str().

karya0:
The string might move (if it were to grow, etc.) and would cause the
pointer to be invalid.

There might be other such incidences in the source code that we must find and fix. I have created #139 to track it.


karya0:
@rohgarg: Can you take another look at this PR. I added another couple of commits to handle the environ variable. The commit messages provide an explanation of the issue/fix.

karya0:
Merged.

PR #141, opened 2015-07-09
Better handling of deleted files.

karya0:
Some systems _prepend_ " (deleted)" instead of doing an _append_. This was observed on travis ci builds (https://travis-ci.org/dmtcp/dmtcp/builds/70006651).

### DIFF
--- a/src/plugin/ipc/file/fileconnection.cpp
+++ b/src/plugin/ipc/file/fileconnection.cpp
@@ -518,38 +518,6 @@ void FileConnection::doLocking()
   _ckpted_file = false;
 }
 
-void FileConnection::handleUnlinkedFile()
-{
-  if ((!jalib::Filesystem::FileExists(_path) && !_isBlacklistedFile(_path)) ||
-      _type == FILE_DELETED ||
-      Util::strStartsWith(jalib::Filesystem::BaseName(_path), ".nfs")) {
-    /* File not present in Filesystem.
-     * /proc/self/fd lists filename of unlink()ed files as:
-     *   "<original_file_name>(deleted)"
-     */
-    string currPath = jalib::Filesystem::GetDeviceName(_fds[0]);
-
-    if (Util::strEndsWith(currPath, DELETED_FILE_SUFFIX)) {
-      JTRACE("Deleted file") (_path);
-      _type = FILE_DELETED;
-    } else if (Util::strStartsWith(jalib::Filesystem::BaseName(currPath),
-                                   ".nfs")) {
-      JTRACE("Deleted NFS file.") (_path) (currPath);
-      _type = FILE_DELETED;
-      _path = currPath;
-    } else {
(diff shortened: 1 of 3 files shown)


src/util_misc.cpp
```diff
@@ -386,7 +386,10 @@ int Util::readProcMapsLine(int mapsfd, ProcMapsArea *area)
   c = readDec (mapsfd, (VA *)&inodenum);
   area -> name[0] = '\0';
   while (c == ' ') c = readChar (mapsfd);
-  if (c == '/' || c == '[') { /* absolute pathname, or [stack], [vdso], etc. */
+  if (c == '/' || c == '[' || c == '(') {
+    // absolute pathname, or [stack], [vdso], etc.
+    // On some machines, deleted files have a " (deleted)" prefix to the
+    // filename.
```
gc00: If `c == '('` is checking for the pattern "(deleted)" as a prefix, then let's have it really check for that entire string.  The kernel developers could later introduce something else like "(optimized)" or some other weird string.  Since there is no standard operating here, we need to protect ourselves, to be future-proof.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -518,38 +518,6 @@ void FileConnection::doLocking()
...
-      Util::strStartsWith(jalib::Filesystem::BaseName(_path), ".nfs")) {
-    /* File not present in Filesystem.
-     * /proc/self/fd lists filename of unlink()ed files as:
-     *   "<original_file_name>(deleted)"
-     */
-    string currPath = jalib::Filesystem::GetDeviceName(_fds[0]);
-
-    if (Util::strEndsWith(currPath, DELETED_FILE_SUFFIX)) {
-      JTRACE("Deleted file") (_path);
-      _type = FILE_DELETED;
```
gc00: If I understand this code, we're going to stop checking for a suffix string " (deleted)".  Or do we check for it somewhere else when it's a suffix?  Could you explain further why this code can be deleted?  Thanks.

src/util_misc.cpp
karya0: We don't have to check for the entire string. Even if something else comes up, the caller will handle the name appropriately. This particular piece of code just ensures that we don't simply fail on seeing a '(' at the beginning of a file name.

Having said that, if they were to introduce some other character, such as '{', we would be in trouble, but that seems quite unlikely and we should think about that only after we see something like that.

src/plugin/ipc/file/fileconnection.cpp
karya0: We just use a better way (`fstat()`) to detect "unlinked" files (see https://github.com/dmtcp/dmtcp/pull/141/files#diff-5d0e9b6251ae20cbd6e6a83177a3d586R550).

rohgarg:
Looks good to me.

PR #142, opened 2015-07-10
Temporary fix for deleted shm file at Stampede.

jiajuncao:
At Stampede, when more than one process on one node share the same deleted shared-memory file, the unlink in mtcp_restart() would cause a race condition. This is just a temporary fix to make it work.


karya0:
LGTM

gc00:
The comments above this patch say:  

 `Otherwise, Upon
    * the second checkpoint, the FILE plugin will treat them as undeleted files.
    * This would cause a problem on the second restart`

Assuming that this fix will sitll make a second restart impossible, can we change the macro name from `STAMPEDE_FIX` to `STAMPEDE_PARTIAL_FIX`?  After that, it looks good to me, and we can commit.

jiajuncao:
Actually this comment only applies to Buffalo, so it won't affect Stampede anyway.

PR #143, opened 2015-07-13
Fix for signal wrappers.

karya0:
Some other misc changes.


karya0:
I saw the comment only now. Yeah, `checkpointSignalBlockedForThread` is a global variable, but we are using some non-ideal generic functions, so let's keep it the way it is for now. Ideally, we would have a separate handle for both Sys V and Posix signals.

PR #148, opened 2015-07-16
PID: Fixed a race when parent exits right after fork.

karya0:
The parent will set the virtual-pid environment variable with
"<virtual-PID>:<virtual-PPID>". The child then records a mapping between
_real_getpid() and virtual-PID as well as _real_getppid() and
virtual-PPID. However, if the parent dies before _real_getppid() is
called, the latter would return the real-pid of the new parent (init
process). The child process then records a mapping between virtual-PPID
of the original parent and real-pid of "init" process. Later on, the
getppid() wrapper fails to detect the exited parent and keeps returning
the virtual-pid of the original parent. This further prevents
ProcessInfo from marking the process as the root of process tree and
thus we see errors on restart.

The bug has been present for a while but didn't show up on faster
machines. We observed it on travis-ci's containerized setup where the
bug showed up quite frequently after a series of recent commits.

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -165,7 +165,8 @@ void CoordinatorAPI::setupVirtualCoordinator(CoordinatorInfo *coordInfo,
   _coordinatorSocket.changeFd(PROTECTED_COORD_FD);
   Util::setCoordPort(_coordinatorSocket.port());
 
-  Util::setVirtualPidEnvVar(INITIAL_VIRTUAL_PID, getppid());
+  pid_t ppid = getppid();
+  Util::setVirtualPidEnvVar(INITIAL_VIRTUAL_PID, ppid, ppid);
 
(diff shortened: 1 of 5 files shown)


src/coordinatorapi.cpp
```diff
@@ -568,7 +571,9 @@ void CoordinatorAPI::createNewConnectionBeforeFork(string& progname)
 
   if (dmtcp_virtual_to_real_pid) {
     JTRACE("Got virtual pid from coordinator") (hello_remote.virtualPid);
-    Util::setVirtualPidEnvVar(hello_remote.virtualPid, getpid());
+    pid_t pid = getpid();
+    pid_t realPid = dmtcp_virtual_to_real_pid(getpid());
```
rohgarg: Change this to `pid_t realPid = dmtcp_virtual_to_real_pid(pid);`?

karya0: fixed.

Issue #149, opened 2015-07-20
dmtcp_get_restart_env() w/ --modify-env fails if no such env var


gc00:
A DMTCP user reported the following bug.  I've reproduced it, but won't have time to do a pull request fo tix it over the next few days.  So, in case someone else can get to it first.  ...  It should be easy to fix.

The issue is that a call to dmtcp_get_restart_env() will make reference to a file `dmtcp_env.txt` upon restart.  If the environment variables inside `dmtcp_env.txt` exist, there's no problem.  But when an environment variable doesn't exist, our documentation says that we should return `-1`.  Instead, we crash with the following `JASSERT`. 
    
    % bin/dmtcp_restart ckpt_a.out_12ec77b5d69-40000-55acb4dd.dmtcp
    Host: login  (129.10.117.100)
    Port: 7779
    Checkpoint Interval: disabled (checkpoint manually instead)
    Exit on last client: 1
    Backgrounding...
    [40000] ERROR at dmtcpplugin.cpp:403 in dmtcp_get_restart_env;     REASON='JASSERT(env_end_ptr > env_buf || env_buf[0] == '\0') failed'
         (char *)env_buf = USER=gene
    a.out (40000): Terminating...`

rohgarg:
Fixed by #151

PR #151, opened 2015-07-20
Fixes issue #149

rohgarg:
@gc00, @karya0: I decided to take a stab at fixing `dmtcp_get_restart_env()`. This is my best guess of what the original code in the function was doing. :-)

### DIFF
--- a/src/dmtcpplugin.cpp
+++ b/src/dmtcpplugin.cpp
@@ -28,6 +28,7 @@
 #include "shareddata.h"
 #include "threadsync.h"
 #include "mtcpinterface.h"
+#include "util.h"
 
 #undef dmtcp_is_enabled
 #undef dmtcp_checkpoint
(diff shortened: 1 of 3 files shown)


rohgarg:
Also, I'm not sure why the function is called `dmtcp_get_restart_env()`. Looking at its functionality `dmtcp_get_env_var()` seems more appropriate to me. Any thoughts?

rohgarg:
Ping :-)

mossberg:
@rohgarg: as i understand it, the function has `restart` in the name because it is used to retrieve new environment variables that are available after a process has been restarted. this is because simply using `getenv()` after restart will retrieve the old environment variables, which were simply restored from saved memory

gc00:
@rohgarg:  The explanation from @mossberg is exactly right!

src/dmtcpplugin.cpp
```diff
@@ -352,89 +353,68 @@ EXTERNC void dmtcp_close_protected_fd(int fd)
...
+     rc = DMTCP_BUF_TOO_SMALL;
+   } else {
+     char *start_ptr = env_buf;
+     // iterate over the flattened list of name-value pairs
+     while (start_ptr - env_buf < sizeof(env_buf)) {
+       pos = NULL;
+       if (strncmp(start_ptr, name, namelen) == 0) {
+         if ((pos = strchr(start_ptr, '='))) {
+           strncpy(value, pos + 1, maxvaluelen);
+           if (strlen(pos+1) >= maxvaluelen) {
```
gc00: Nice catch!  maxvaluelen was counting the buffer size, while strlen() was counting char's, _excluting_ the final null char.  Your code fixes the bug in this boundary case.  Formerly, I was comparing with just `>`.

src/util_misc.cpp
```diff
@@ -274,23 +274,40 @@ void Util::dupFds(int oldfd, const vector<int>& newfds)
...
+ *
+ * Returns num of chars read on success;
+ *         -1 on read failure or invalid args; and
+ *         -2 if the buffer is too small
+ */
 int Util::readLine(int fd, char *buf, int count)
 {
   int i = 0;
   char c;
+  if (fd < 0 || buf == NULL) return -1;
```
gc00: It's a nice catch to guard against `fd < 0`!  However, I wonder if we should do a JASSERT in this case, instead of leaving it up to the caller of `Util::readLine()` to check the return value.  If `fd < 0`, then the programmer should have checked for this _before_ calling `Util::readLine()`, rather than looking at the return value, and trying to fix it after the fact.

gc00:
@rohgarg : I have one significant in-line comment, concerning `src/util_misc.cpp`.  After we address that, it's good to go.  (My other in-line comment, is just a compliment on catching a boundary case bug.)

Thanks very much for cleaning up my original code!  I assume that you've also tested it, and that it also fixes the original bug report.  Nice job.

src/util_misc.cpp
rohgarg: Yes, I agree. I'll change this to a `JASSERT()`.

PR #152, opened 2015-07-25
Use automake for building internal plugins.

karya0:
This PR contains several smaller commits. Feel free to add comments to individual commits to make it easier to review.

The goal here is to use a single Makefile.am to build all the internal plugins and another one to build all plugins in top-level plugin/ directory. While automake is considered complex, our project already depends heavily on it and mixing plain Makefile.in files isn't scalable and is harder to maintain. The new Makefile.am files are fairly simple and we don't have any complex rules for them. With the growing list of internal plugins, we might at some point want to bundle several of them in single .so library and having a single Makefile.am makes it much easier to do.


rohgarg:
Looks good.

gc00:
@karya0, @rohgarg : Can we revert this for now, and discuss this?  I have some strong opinions here.

There is a tension between capturing two goals in DMTCP.  On the one hand, we would like to use state-of-the-art tools that hackers are familiar with, but CS undergraduate, and the majority of CS Ph.D. students are not familiar with.  This makes our code development for experienced programmers.

On the other hand, we want to make DMTCP easy to understand so that competent programmers will feel confident about tracing the logic of DMTCP, and then contributing to it.  I see a large part of this other end of the stick. With this change, I have to start explaining to students something like:

"And it's really easy to get started with DMTCP, because plugins are really modular.  You just need to understand the code within the given plugin directory.  But if you want to add a new file, or merge two files, or understand how a plugin is made, then please read up on automake.  To help get you started, all you need to know is that there are suffixes like _PROGRAMS, _SCRIPTS, _LIBRARIES, _SOURCES, _CPPFLAGS, _LDFLAGS, _LDADD, and more.  The names should be intuitive, so that you can quickly be up and running.  When you do configure in DMTCP, it will then generate a Makefile that is not intended to be human-readable.  So, the build process will be a "black box".  But if you really need to see some detail of the build process, you can still grep through Makefile.  (The Linux build commands being used are purposely omitted from Makefile.am to make things simple.)  Also, if you want to change anything, you'll need to learn about the automake command.  And if you need to use one of the autoconf variables in your Makefile, then ask one of us to do it."

Anyway, I hope you see my point.  this commit will make Kapil's job easier, because Kapil knows automake quite well.  I also know automake quite well.  But it will make my job extremely hard, because I'll have a lot more difficulty introducing people to the DMTCP project.

I can accept that the DMTCP core (src directory) is for experts only.  But I do want to keep the plugins simple and easy for entry-level people.  This will also help as we move on, and port DMTCP to new areas:  Android, Windows, embedded Linux, etc.

Can we revert this commit, and discuss the issues together?  Thanks.

gc00:
@karya0 , @rohgarg : One further point.  Kapil wrote:
    The new Makefile.am files are fairly simple and we don't have any complex rules for them.

The reason that the Makefile.am files are simple is because they are derived from a Makefile build.  The discipline of a Makefile build was good for us, because it kepts things simple.  Once we move to Makefile.am, I guarantee that we'll move down the slippery slope, since a single new line in our Makefile.am can replace some longer, more explicit build code.

karya0:
One of the things that kept coming up with multiple Makefile.ins was the amount of copy-paste we had to do when changing some rules globally. It was more error prone and also made it much harder for packaging and even tracking dependencies.

Having said that, I do agree with @gc00 in that the automake business is a complex one and we should try to avoid new contributors from getting hit by it. to that end, I have deliberately left the contrib and test directories as is (without automake). The idea is that the novice DMTCP developer wouldn't need worry about the internal plugins anyways and can easily take a look at contrib or test/plugins to find their way.

karya0:
PS: At some point, we might just want to throw all this automake stuff and shift to CMake which is easier to write/update and much more intuitive :-). But until then, I think we should keep the automake "magic" for stuff inside src/ since it's an experts only zone :-).

gc00:
@karya0 :  How about this intermediate proposal:  We could use the `include` command of GNU make to greatly reduce the need for copy-past.  Especially if our long-term goal is CMake, then staying with standard make for a little longer will make migration easier.

karya0:
I think include won't quite solve all the problems with recursive targets. I have tried that in past and it wasn't pretty. The CMake change would require a bit of work and I am not sure if any of us would have any free cycles to implement it in the medium term. And when we do have that time, it would be trivial to convert these new Makefile.ams into CMake.

So I would suggest that we keep the automake stuff for our internal plugins and for contrib plugins and test plugins, we keep maintaining traditional Makefiles. BTW, not sure if you had a chance to look at the new automake stuff, but to give an example, the ptrace plugin is now built using these following lines:

```
#
# Ptrace plugin
#
libdmtcp_PROGRAMS += $(d_libdir)/libdmtcp_ptrace.so
__d_libdir__libdmtcp_ptrace_so_SOURCES =                               \
	ptrace/ptrace.cpp                                              \
	ptrace/ptrace.h                                                \
	ptrace/ptraceinfo.cpp                                          \
	ptrace/ptraceinfo.h                                            \
	ptrace/ptracewrappers.cpp
__d_libdir__libdmtcp_ptrace_so_LDFLAGS = $(dmtcp_ldflags)
__d_libdir__libdmtcp_ptrace_so_LDADD = $(LDADD)
```

gc00:
@karya0 : Shall we talk on the phone?

I could accept the _ptrace_ plugin using automake, since few entry-level DMTCP contributors will be looking at it.  For most of the plugins inside src, I'd like to keep GNU make.  My general worries are:

1. Entry-level DMTCP contributors will get scared off by experts-only automake stuff.  (We're now starting to get entry-level DMTCP contributors from several directions, and I've been working to encourage them.)

2. If we have a medium-term goal of using Cmake (and I agree that Cmake is more intuitive), then a roadmap like "GNU make -> automake -> Cmake" seems to add complication.

karya0:
@gc00: Sure, phone sounds good.

gc00:
For others, Kapil and I talked on the phone.  The longer story is that Kapil has some really nice ideas for cleaning up and improving the plugin architecture of DMTCP.  The existing Makefile structures were getting in the way of the potential improvements from an improved architecture.  I agree that in this case, the benefits outweigh the costs, and so we should keep the automake for now.

For the medium-term (over the next year?), we should definitely think about replacing the automake tools with Cmake.  But for now, we can live with automake as the "lesser evil".

heroxbd:
From a Linux distribution maintainer point of view (I am maintaining dmtcp package in Gentoo), autoconf/automake/libtool is easier and more predictable.  CMake has many hidden assumptions and is less transparent and introduces glitches on old systems (RHEL/CentOS 5 for example) on every new release.

PR #153, opened 2015-07-27
valloc: replace memalign with valloc

rohgarg:
This patch replaces the call to _real_memalign() in the
valloc wrapper (in the alloc plugin) with a call to
_real_valloc(). The call to memalign() imposes a runtime
overhead of up to 600 percent.

### DIFF
--- a/src/plugin/alloc/alloc.h
+++ b/src/plugin/alloc/alloc.h
@@ -30,6 +30,7 @@ extern "C" void *__libc_memalign(size_t boundary, size_t size);
 
 #define _real_malloc  NEXT_FNC(malloc)
 #define _real_calloc  NEXT_FNC(calloc)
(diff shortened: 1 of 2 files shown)


karya0:
LGTM.

rohgarg:
Okay, so we can no longer checkpoint the `realpath` test and the `java1` test with this patch. Let me investigate.

rohgarg:
@kapil: so, your previous commit: 95a601c is the culprit :-)

It breaks all the tests that rely on `dlsym_default`. Should we pull that out?

karya0:
That was stupid of me. Just pushed the fix.

gc00:
@rohgarg: The java1 and realpath tests are failing on this branch.  Probably you just need to pull in the fix that @karya0 pushed in, and the tests will work again.  Could you update this branch?

PR #154, opened 2015-07-28
MTCP: Fixed a race w/ unlinked shared-memory files.

karya0:
For shared memory areas backed by unlinked files, MTCP used to create
the file with RW permissions and populate it with the contents from the
ckpt image. Next, it would close and reopen the file descriptor to
reflect the appropriate pre-ckpt flags. Finally, the memory area would
be mapped and the file would be unlinked.

When there are two or more processes involved, they all follow the same
steps. However, sometimes, one of the processes will unlink the file
right when a different process is in the middle of closing and reopening
the file descriptor. This would cause the reopening of the file to fail
and the process to exit.

The fix is to remove the closing/reopening logic since the flags on the
file descriptors don't matter. We are anyways creating the shared-memory
area with appropriate flags, so it's okay to have RW flags sets on the
file descriptor even if the shared memory area needs only the R flag.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -1140,15 +1140,6 @@ static void read_shared_memory_area_from_file(int fd, Area* area, int flags)
     if (area->prot & PROT_EXEC)  fileprot |= S_IXUSR;
     mtcp_sys_fchmod(imagefd, fileprot);
 
-    //close the file
-    mtcp_sys_close(imagefd);
-
-    // now open the file again, this time with appropriate flags
-    imagefd = mtcp_sys_open (area_name, flags, 0);
-    if (imagefd < 0){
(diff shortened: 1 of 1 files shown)


gc00:
@karya0 : Very nice catch!  I'm curious in what context you saw this bug.  Does it occur frequently?

If I understand the rationale, it's that we have two types of permissions for this region:  permissions on the fd, and permissions on the shared memory area to which it refers.  Since almost all use cases by users will have the permissions of the shared memory the same or more restrictive than the permissions of the fd, it's okay for DMTCP to make the fd permissions even more permissive (i.e., RW).

I agree with the logic.  Could we have some comment in the code that briefly summarizes this logic?

After that, it's good to go.  Thanks.

karya0:
I saw the failures with shared-area2 in our test suite. The permissions aren't really an issue since we are going to close the file descriptor anyways. I think this logic might have been a left over from a different approach that we used several years back.

PR #155, opened 2015-07-28
Fix dmtcp_get_generation()

gc00:
The function `dmtcp_get_generation()` in `include/dmtcp.h`.  Sometimes gives the wrong answer.  This fixes the bug.  In addition, a function `checkpoint_is_pending()` was added in the last of the commits.

I can optionally remove the function `checkpoint_is_pending()`, but the implementation is tiny, and it seems like it might be useful in the future.

The reason for the original bug is that the DMTCP generation (number of checkpoints so far) is maintained by the coordinator for the whole computation.  This is as it should be.  But on each host, the sharedDataHeader keeps one copy of the generation in shared memory for all processes on the same host.  This means that when a checkpoint request from the coordinator arrives at one process, it will update the generation even before the other processes have seen the checkpoint request.  Hence, other processes on that host will report a generation that is one too high (at least until they receive their own checkpoint request from the coordinator).

The bug also exists even for a single user process.  The checkpoint thread receives a checkpoint request and updates the shared memory variable for the generation.  The user thread then sees this and uses it, even before the checkpoint thread has had time to quiesce the process.

The motivation for this was that dmtcp_get_generation() was needed in test/plugin/applic-delayed-ckpt (and applic-initiated-ckpt), to fix a different plugin.  In those examples, the suer process would request a checkpoint, and it would then exit before the checkpoint had actually been invoked.  By using dmtcp_get_generation(), these examples can now test the generation count, and refuse to exit until the generation count has been updated (which happens now, just before the preCkpt event).  The improved code in applic-delayed-ckpt (provided as one of these commits) would still fail on slow computers with the old `dmtcp_get_generation()`.  (In fact, it seems to fail for still a different reason.  `dmtcp_get_generation()` continues to report the generation as 0.  It doesn't see the latest value in shared memory.)  The new version allows the improved code in applic-delayed-ckpt to succeed.

The solution was to rename the DmtcpUniquePid field, `_generation`, to now be `_coord_generation` (which is part of sharedDataHeader in shared memory); and then create a new field, `_generation`, as part of `ProcessInfo`, and that field is now a per-process value for the generation.  It is updated at the time of preCkpt, by coping the value of `_coord_generation` during preCkpt.

I observed the bug on a 32-bit dual-core Pentium 4 machine with 1 GB RAM, running at 3.2 GHz, and running Ubuntu 12.04.  To make the bug reproducible, I did:

    cd test/plugin/applic-delayed-ckpt
    make clean
    make
    make check
The original code of that example (with no call to  `dmtcp_get_generation()`) would exit before the ckpt, and so restart would fail.

### DIFF
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -208,6 +208,7 @@ EXTERNC const char* dmtcp_get_computation_id_str(void);
 EXTERNC uint64_t dmtcp_get_coordinator_timestamp(void);
 // Generation is 0 before first checkpoint, and then successively incremented.
 EXTERNC uint32_t dmtcp_get_generation(void) __attribute__((weak));
+EXTERNC int checkpoint_is_pending(void) __attribute__((weak));
 
 /**
  * Gets the coordinator-specific status of DMTCP.
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -207,7 +207,7 @@ EXTERNC const char* dmtcp_get_uniquepid_str(void) __attribute__((weak));
 EXTERNC const char* dmtcp_get_computation_id_str(void);
 EXTERNC uint64_t dmtcp_get_coordinator_timestamp(void);
 // Generation is 0 before first checkpoint, and then successively incremented.
-EXTERNC uint32_t dmtcp_get_generation(void) __attribute((weak));
+EXTERNC uint32_t dmtcp_get_generation(void) __attribute__((weak));
 
 /**
  * Gets the coordinator-specific status of DMTCP.
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -99,7 +99,7 @@ typedef struct DmtcpUniqueProcessId {
   uint64_t  _hostid; //gethostid()
   uint64_t _time; //time()
   pid_t _pid; //getpid()
-  uint32_t _generation; //generation()
+  uint32_t _computation_generation; //computationGeneration()
 } DmtcpUniqueProcessId;
 
 EXTERNC int dmtcp_unique_pids_equal(DmtcpUniqueProcessId a,
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -1243,7 +1243,7 @@ bool DmtcpCoordinator::validateNewWorkerProcess(
       // Connection of new computation.
       compId = UniquePid(hello_remote.from.hostid(), client->virtualPid(),
                          hello_remote.from.time(),
-                         hello_remote.from.generation());
+                         hello_remote.from.computationGeneration());
 
       JASSERT(gettimeofday(&tv, NULL) == 0);
       // Get the resolution down to 100 mili seconds.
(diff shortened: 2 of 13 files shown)


karya0:
Having two generations is a bit confusing. Specifically, for the per-process generation, should we reset the generation after an fork and/or exec?

An alternate proposal is to increment the generation during resume _after_ checkpoint image has been created (for restarting processes, it can be incremented right after restoring memory). Would this solve the problem that you are seeing with the test plugins?

gc00:
There will always be two generations.  There is the current generation, as known by the coordinator (updated when the coorrdinator requests a new checkpoint), and the current generation, as known by the client (updated when the client checkpoints).  I had found the original code very confusing, because I initially assumed that the `generation` variable on the client would reflect what is the current generation of the client.  This was not the case previously.

As for resetting the generation after fork/exec, this is currently handled correctly.  After a fork, the client `generation` stays the same.  During an exec, the `generation` is saved, note that `ProcessInfo::ProcessInfo()` saves `_generation` along with the other per-process variables, and that is used to preserve per-process variables across an exec.

Intuitively, I had always assumed that when we begin a new checkpoint, this is the _beginning_ of a new generation.  So, incrementing the generation during resume/restart would be counter-intuitive.  Then, a plugin using the preExec event would see the generation number from the previous generation, while the postExec event would display the current generation number.

gc00:
Hmm.  I tried testing against fork and exec, and something is still not right with the generation.  I'll work on that, and add a fix for that.

karya0:
>
> Intuitively, I had always assumed that when we begin a new checkpoint,
> this is the *beginning* of a new generation. So, incrementing the
> generation during resume/restart would be counter-intuitive. Then, a plugin
> using the preExec event would see the generation number from the previous
> generation, while the postExec event would display the current generation
> number.
>
Now I am confused :). What is the relationship of incrementing generation
during resume/restart and pre/postExec? These should be orthogonal.

gc00:
Let's talk on the phone if the bandwidth of text is not enough.

Anyway, the current logic of the commits is:

1.  The coordinator increments the _coord_generation when it calls for a new checkpoint.
2.  The coordinator includes the current _coord_generation in the message, when it sends a request to each client to checkpoint.
3.  Each client copies the _coord_generation into the sharedMemoryHeader when it receives the message.  So, the _coord_generation becomes visible to all clients when the first client receives the message.
4.  When a client processes a checkpoint request, it quiesces all user threads.  It then copies the _coord_generation into _generation (the per-process generation).  It then invokes the preCkpt event and any plugins responding to the event will be able to see _generation, by calling dmtcp_get_generation.

Because of this, a client first sees the new generation in responding to a preCkpt event, even though some other client may have copied _coord_generation into the sharedMemoryHeader at an earlier time.

Does this make the logic clear?  Because a client _never_ increments a generation, this logic should be robust.  A client copies from _coord_generation, and only the coordinator will increment a generation.

gc00:
@gc00 wrote:

    Hmm. I tried testing against fork and exec, and something is still not
    right with the generation.  I'll work on that, and add a fix for that.

I've now decided that this is an independent bug.  Let's continue to consider this pull request, while I track down the new bug.

====
**If you want to see the new bug (Surprise, a bug!  :-) ), then compile and run:**

    int main(int argc, char *argv[]) {
      while (1) {
        int interval = 5;
        while( (interval = sleep(interval)) > 0 ) ;
        printf("will exec.\n");
        execvp(argv[0], argv);
      }
    }

Then invoke this with `dmtcp_launch -i1 ./a.out`.

Apparently, even calling exec once every five seconds is enough to trigger the bug.  Apparently, when the coordinator is asked to checkpoint every second (`-i1`), then `workersRunningAndSuspendMsgSent` in `dmtcp_coordinator.cpp` becomes stuck at 1.  `DmtcpCoordinator::startCheckpoint()` is called every second, as it should be, but the coordinator refuses to ask for further checkpoints in this situation.

PR #157, opened 2015-07-29
Enhance test/plugin/*/Makefile; kill old coordinator

gc00:
In the test/plugin examples, kill any old coordinator using `${DEMO_PORT}` as first step of `make check`.


karya0:
Looks good.

PR #158, opened 2015-07-29
Polish comments (esp. dmtcp.h)

gc00:
There are two commits.  The first commit should be standard, and not need any significant review.

The second commit changes include/dmtcp.h.  It adds comments, and re-orders some declarations.  But it does not make changes to the code.  The goal is to better document the use of plugins for end users.  Comments and improvements to the dmtcp.h comments are welcome.

### DIFF
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -38,6 +38,13 @@
 # endif
 #endif
 
+// C++ takes null arg, while C takes void arg.
+#ifdef __cplusplus
+# define VOID
+#else
+# define VOID void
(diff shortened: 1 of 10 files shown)


gc00:
The review from @karya0 was somehow placed here: https://github.com/gc00/dmtcp/commit/aefae986fdefdb2fcec88a965e2f606cf5af3961
Anyway, the decision was to reformat one of the comments, and then push it in.

gc00:
Pushed in (and conflicts were fixed manually).

Issue #165, opened 2015-08-05
Coordinator is CPU hog during checkpointing when using interval timer


gc00:
In `DmtcpCoordinator::eventLoop()`, the current logic is that if there is an interval timer, then `epoll_wait()` returns when there is a timeout.  The variable `timeout` is then adjusted based on the return value of `DmtcpCoordinator::getRemainingTimeoutMS()`.  That function computes the timeout based on a start time, `startTime.tv_sec` (a file-local, global variable).  The variable `startTime` is updated to the current time in `DmtcpCoordinator::resetCkptTimer()`.  But that does not happen until just after the call to `broadcastMessage ( DMT_DO_RESUME )` at the end of the checkpointing sequence.

So, the net effect is that the coordinator is doing busy waiting inside the `eventLoop()` from the time that the interval timer elapses until the time of `DMT_DO_RESUME`.  If the coordinator is on the same host as one of the clients, it can penalize the performance of that client during checkpointing.

Note that this issue does not arise for manual checkpointing (`theCheckpointInterval == 0`).  That's because in this case, `DmtcpCoordinator::getRemainingTimeoutMS()` returns&nbsp;-1.  A timeout value of&nbsp;-1 causes the eventLoop to never time out.  This is the desired behavior when we're in the middle of the checkpointing sequence.  Hence, an easy fix for this issue would be to change the statement `timeout = 0;` to `timeout = -1;` in `DmtcpCoordinator::getRemainingTimeoutMS()`.

I've tested this fix for some examples (dmtcp1, dmtcp5, forkexec, shared-memory1), and it seems to work.  Furthermore, I used to see an annoying hesitation during the interval checkpoint when executing `dmtcp_launch -i5 test/dmtcp1`.  This no longer happens now.  Also, `dmtcp_restart -i5` seems to work.  I'll put up a corresponding pull request shortly.

gc00:
This was closed by pull request #166.

PR #166, opened 2015-08-05
Coordinator is CPU hog during checkpointing when using interval timer

gc00:
`dmtcp_launch -i5 test/dmtcp1` causes the coordinator to do busy waiting during the checkpointing sequence.  This slows down a client on the same host, which is trying to use the CPU to do the checkpoint.  See issue #165 for a detailed analysis of the cause, and why this one-line patch fixes it.


rohgarg:
Looks good to me.

PR #170, opened 2015-08-11
Enabled File plugin to handle shared memory areas.

### LINKED ISSUE #69: shared-memory1 is flaky
It sometimes fail with the following error (observed with `./test/autotest.py -v shared-memory1`):

```
[14211] mtcp_restart.c:1085 read_shared_memory_area_from_file:
  error 2 opening mmap file /home/kapil/dmtcp/dmtcp-shared-memory.QExEDL
```

### LINKED ISSUE #156: Handle shared memory areas in the File plugin
There is a race if two processes, that have a shared memory area backed by an unlinked file, are restarting at the same time.  It's possible that the first process creates, mmaps, and unlinks the file before the second process has a chance to use that file. In that case, the second process will create a new file, mmap it and then unlink it. Even though both of them have mapped the file as shared, the two memory regions refer to two different inodes and thus the file is not truly shared.  A proper fix would require synchronization between all restarting processes and thus the file plugin is a better place to handle shared memory areas.

This issue is somewhat related to #68.

### LINKED ISSUE #167: Bug in shared-memory1
__[ NOTE: This issue is probably related to these  other open issues, which also concern shared memory: #65, #68 and #156. ]__

I'll dig further later to analyze this more, but the shared-memory1 test is hanging on the second restart.  Because it hangs, autotest.py doesn't catch this test failure.  I test with an interval timer (`-i 5`) in all cases, but I'm guessing the bug occurs in the general case.

There are at least two problems:
1.  It hangs on restart.  It should either succeed or fail hard.  When it hangs, it happens at `ipc/connectionlist.cpp:547`.  The line of code is:<br> `int ret = _real_select(restoreFd+1, &rfds, &wfds, NULL, NULL);`<br>  Unfortunately, the NULL parameter means no timeout.  In the initial stages of restart, we should timeout and fail with a `JASSERT()` if there is a timeout.  Maybe we can directly detect failure in some other way.
2.  On an NFS filesystem, when executing the older `shared-memory.c` from DMTCP-2.3.1, I see:

    [40000] ERROR at fileconnection.cpp:686 in refill; 
    REASON='JASSERT(jalib::Filesystem::FileExists(_path)) failed'
         _path = /home/gene/dmtcp-vanilla/.nfs000000000234f48100000155
    Message: File not found.

However, after that JASSERT, both processes are still hanging as `[mtcp_restart]`.  Probably something related is happening on a non-NFS filesystem.

Finally, DMTCP-2.3.1 seems to work fine for both `shared-memory.c` and `shared-memory1.c` (from DMTCP-2.4.0), at least on the non-NFS filesystem.

So, there seems to have been a regression in our handling of shared memory.

karya0:
The capability is removed from MTCP and moved completely into File plugin. I tested it with our test suite and everything seemed fine. It would be better if others can take it for a test ride.

For reviewing, I would suggest that you look at individual commits as some of the commits are somewhat unrelated.

This Closes #167, closes #156 and closes #69.


gc00:
Hi @karya0.  Nice patch.  We've needed this for a while!  Here are some smaller comments:
1. Let's rebase  _Removed deprecated install-hooks api._ to be the first of the commits (or the last of the commits).  I expect that most of these commits will go into DMTCP 2.4.1, and We should do the deprecation for DMTCP 2.5, rather than for 2.4.1.  So, we don't want this commit in the middle of the others.
2.  See inline comment for _Better NSCD handling for writeckpt_, at: https://github.com/karya0/dmtcp/commit/91acdf9210f3c98001706592eb98f0b43d8f30ad.
3.  See inline comment for _Sys V SHM: Don't release mem region during ckpt._, at: https://github.com/karya0/dmtcp/commit/d53e389aa465ee0f07118dcd7a77eb9244d07dd2.
4. See inline comments for _writeckpt.cpp: Use ProcSelfMaps._, at: https://github.com/karya0/dmtcp/commit/26291ade3148b13e7e380cd43669ae907c441900.
5. And tiny inline comment for _	File: Moved msync from writeckpt.cpp to File plugin._, at: https://github.com/karya0/dmtcp/commit/ab735a9a2a16082fb773bf4aec35f9d978d24e73

karya0:
I have addressed all the concerns. Here is the summary for the issues:

> Issue 1: install-hooks
It is the first commit, but somehow github UI is not showing it like that. However, if you track the commits through the parent link, you can see it. 

> Issue 2: mtcp_restart.c:fix_parent_directories()
This is already handled by the new code: https://github.com/dmtcp/dmtcp/pull/170/files#diff-e9f539d1f306bad2a4a3658de2364d59R305

gc00:
It seems that @karya0 pushed this in, without the fix pointed out by @rohgarg above.  I've now added pull request #177, which fixes this, and also addresses a second issue that prevents us from building when configured with `--enable-debug`.  We can continue the discussion there.

PR #171, opened 2015-08-12
ckptfiles-plugin: add functionality

rohgarg:
This patch adds basic ability to read a list of files to be checkpointed and their restart locations (which may be different on restart) from a user-specified database (simple text file). A sample database file is also provided.


karya0:
I wonder if we should instead use a more verbose format like YAML for these kinds of things? Perhaps, we can find a small (preferably header-only) library that we can use to parse/generate this stuff.

rohgarg:
That was my initial thought but I just wanted a quick prototype.

gc00:
This pull request has been waiting quite long.  It seems like we're in a situation where "the perfect is the enemy of the good".  I volunteer to help finish this review now.  I think I saw that one of the DMTCP users was interested in using this.  If there is a better design for the future, we can add that as a github issue, and label it as an "enhancement".

1. The `ckptifile/README` for this still has only the text: *"This plugin serves as a template.  Please modify the given functions according to your needs."*  Can we update this for the new functionality?

2.  Without the README, it's hard for me to tell for sure, but I suspect that `ckptfiles.dat` contains ASCII (human-readable) text rather than binary.  If that's true, can we change `.dat` to `.txt`?

After that, I plan to read the README, and try it out myself.  That will help me understand the code better.  I'm looking forward to it.  This is the type of functionality that I've heard more than one user request.

rohgarg:
@gc00, @karya0: Could you please re-review this?

> The ckptifile/README for this still has only the text: "This plugin serves as a template. Please modify the given functions according to your needs." Can we update this for the new functionality?

I have updated the README.

> Without the README, it's hard for me to tell for sure, but I suspect that ckptfiles.dat contains ASCII (human-readable) text rather than binary. If that's true, can we change .dat to .txt?

The current one is just a sample file. A user can use whatever they want by pointing the `DMTCP_CKPTFILES_DATA` environment variable to right text file.

rohgarg:
Ping. :-)

contrib/ckptfile/README
```diff
@@ -1,2 +1,47 @@
...
+    being used by an application. In this case, DMTCP does not save a
+    copy of the file at checkpoint time.
+
+    On restart, the file that the application was using at checkpoint
+    time is expected to be at its original location, and only the file
+    pointers are restored.
+
+## Usage
+
+    $ DMTCP_CKPTFILES_DATA=/path/to/ckptfiles.dat dmtcp_launch --with-plugin /path/to/libdmtcp_ckptfile.so ./a.out
```
gc00: Let's put the `## Usage` comment and the following two lines at the beginning of this README file.

contrib/ckptfile/README
```diff
@@ -1,2 +1,47 @@
...
+
+ 1) DMTCP checkpoints an open file:
+
+    This happens when a user asks DMTCP to checkpoint a file being used
+    by an application. In this case, DMTCP saves a copy of the file as
+    it existed at checkpoint time in a separate directory.
+
+    On restart, the saved copy of the file is restored at its original
+    location, the file pointers are restored, and the user application
+    can continue to read from there.
```
gc00: SUGGESTED RE-WORDING:
```
   location, the corresponding file descriptors are restored, and the user
   application can continue to read from the offset saved at checkpoint time.
```

contrib/ckptfile/README
```diff
@@ -1,2 +1,47 @@
...
+
+ 2) DMTCP does not checkpoint an open file:
+
+    This happens when a user does not ask DMTCP to checkpoint a file
+    being used by an application. In this case, DMTCP does not save a
+    copy of the file at checkpoint time.
+
+    On restart, the file that the application was using at checkpoint
+    time is expected to be at its original location, and only the file
+    pointers are restored.
```
gc00: SUGGESTED RE-WORDING:
```
   descriptors are restored (with the offset saved at checkpoint time).
```

contrib/ckptfile/README
```diff
@@ -1,2 +1,47 @@
...
+
+The plugin reads a user-specified database at checkpoint time. The
+database is a simple text file that specifies the files that should
+be checkpointed, and the restart paths of files that shouldn't be
+checkpointed.
+
+The database file's format is defined below.
+
+   file:ckpt={1|0};rstrt=file
+   file-pattern:ckpt={1|0};rstrt=file-pattern
```
gc00: I have several comments and questions here:

1. Shouldn't we be writing FILE and FILE_PATTERN to emphasize that these are meta-names that should be replaced with an actual filename?
2. Would it be easy to modify the code to use true and false, and change the README to:  FILE:ckpt={true|false};... (and similarly for FILE_PATTERN?
3. You never say what is a FILE_PATTERN.  Is it a regular expression?
4. In `file:ckpt={1|0};rstrt=file`, do the two occurences of the keyword, `file`, refer to the same filename?  If so, why repeat `file`?  If not, then why do you later write `the saved copy of the file is restored at its original`?
5. How does the user know if something like `a.txt`will be treated as a raw name, or as a regular expression with `.` matching any single character?  (This assumes that `file-pattern` means a regular expression.)

Have you considered an alternative syntax, such as:
```
SAVE_AND_RESTORE: {FILE|FILE_PATTERN}
FILE_WILL_BE_UPDATED_BEFORE_RESTART: {FILE|FILE_PATTERN}
```

contrib/ckptfile/README
```diff
@@ -1,2 +1,47 @@
...
+    by an application. In this case, DMTCP saves a copy of the file as
+    it existed at checkpoint time in a separate directory.
+
+    On restart, the saved copy of the file is restored at its original
+    location, the file pointers are restored, and the user application
+    can continue to read from there.
+
+ 2) DMTCP does not checkpoint an open file:
+
+    This happens when a user does not ask DMTCP to checkpoint a file
```
gc00: SUGGESTED RE-WORDING:
```
   This happens when a user asks DMTCP to _not_ checkpoint a file
```

contrib/ckptfile/README
```diff
@@ -1,2 +1,47 @@
...
+The database file's format is defined below.
+
+   file:ckpt={1|0};rstrt=file
+   file-pattern:ckpt={1|0};rstrt=file-pattern
+
+Note that for a file that's checkpointed (ckpt=1), the restart path
+(rstrt=XXX) is ignored.
+
+## Background
+
```
gc00: SUGGESTED ADDITION:
```
Recall that the default behavior of DMTCP in checkpointing
open files is:
a) if the file is open as read-only, do not assume write permission,
   and do not checkpoint it.
b) if the file is open with write permission, and the offset is at the
   end of the file, assume that the file is a log file.  Do not save,
   and set the file descriptor back to the offset saved at checkpoint time.
c) if the file is open with write permission, and the offset is _not_ at the
   end of the file, assume that the file is a database file.  Save the file
   contents and restore at restart time.
d)  If invoked as:  dmtcp_launch --checkpoint-open-files
    then save the contents of all open files, and restore at restart time.

```

gc00:
@rohgarg, Sorry for the delay in reviewing this.  We can also discuss it offline.

I really like this plugin.  Please see my in-line notes on suggested re-wording for the README file.
Also, I noticed that you have empty files, `contrib/ckptfile/test/test1` and `.../test2`.  Is that intentional?

rohgarg:
> @rohgarg, Sorry for the delay in reviewing this. We can also discuss it offline.

Thanks for reviewing this @gc00. This was mostly a first attempt at trying to make the existing plugin a little more useful. :-)

I'll get to all the comments, eventually, but to start with the first one here.

> Also, I noticed that you have empty files, contrib/ckptfile/test/test1 and .../test2. Is that intentional?

The files aren't empty, they do have some text. Please see my latest commit: 03f72538f4bb9898a6c062ca241810ab838ede54.

gc00:
> The files aren't empty, they do have some text. Please see my latest commit: 03f7253.

Is the text in contrib/ckptfile/test/test1 and .../test2 important?  If not, can we put in a header line at the top saying something like `This file is for testing.`

ankitcse07:
When this change be available in master branch ?

gc00:
Hi @ankitcse07,  @rohgarg tells me that he is hoping to finalize and commit this along with some related commits during this week (before the winter break).

ankitcse07:
Thanks @gc00 for the information.

rohgarg:
@gc00, @karya0: Could you please re-review this?

gc00:
@rohgarg, please see my in-line notes that I wrote for the commit:
https://github.com/rohgarg/dmtcp-1/commit/fc5ea28389784473c1df678d357938515d2efd2a
concerning polishing the wording for the README file.

gc00:
@rohgarg:  LGTM 

When we're together again, let's interactively go over the README to see if we can further improve the clarity.  But this is about as good as it will get through the online code review.  Let's push this in so that all developers can begin using this.

PR #173, opened 2015-08-13
pthread-mutex: allow checkpoints between locks and unlocks

rohgarg:
If a checkpoint is taken after a lock is acquired by an application but before
the application had a chance to release it, it can result in a deadlock on
restart. This patch fixes this race condition, and will allow checkpoints
between a lock and an unlock in the future. Detailed explanation follows.

libc caches the tid of the owner of a mutex, which is determined when acquiring
the lock, in its internal structure, and only the owner of the mutex can unlock
it.  As a result, on restart, the call to `pthread_mutex_unlock()` fails with an
EPERM.  If an application fails to check the result of the unlock, trying to
acquire the lock again will cause a deadlock. This patch fixes this problem by:
a) keeping track of the mutexes that were still locked at checkpoint time, and
b) patching the libc's internal mutex struct for all such mutexes.

### DIFF
--- a/include/virtualidtable.h
+++ b/include/virtualidtable.h
@@ -107,6 +107,8 @@ namespace dmtcp
           resetNextVirtualId();
         }
 
+        map<IdType, IdType> getMap() const { return _idMapTable; }
+
         bool getNewVirtualId(IdType *id) {
           bool res = false;
           _do_lock_tbl();
--- a/src/plugin/Makefile.am
+++ b/src/plugin/Makefile.am
@@ -129,6 +129,7 @@ __d_libdir__libdmtcp_pid_so_SOURCES =                                  \
 	pid/pidwrappers.cpp                                            \
 	pid/pidwrappers.h                                              \
 	pid/virtualpidtable.cpp                                        \
+	pid/pthread_mutex_wrappers.cpp                                 \
 	pid/virtualpidtable.h
 __d_libdir__libdmtcp_pid_so_LDFLAGS = $(dmtcp_ldflags)
 
--- a/src/plugin/Makefile.in
+++ b/src/plugin/Makefile.in
@@ -170,7 +170,8 @@ __d_libdir__libdmtcp_ipc_so_LINK = $(CXXLD) $(AM_CXXFLAGS) $(CXXFLAGS) \
 am___d_libdir__libdmtcp_pid_so_OBJECTS = pid/pid.$(OBJEXT) \
 	pid/pid_filewrappers.$(OBJEXT) pid/pid_miscwrappers.$(OBJEXT) \
 	pid/pid_syscallsreal.$(OBJEXT) pid/pidwrappers.$(OBJEXT) \
-	pid/virtualpidtable.$(OBJEXT)
+	pid/virtualpidtable.$(OBJEXT) \
+	pid/pthread_mutex_wrappers.$(OBJEXT)
 __d_libdir__libdmtcp_pid_so_OBJECTS =  \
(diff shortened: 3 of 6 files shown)


rohgarg:
@ankitcse07: Could you please check if this works for you?

rohgarg:
Okay, some tests are failing. I'm investigating; will update the patch when I have a fix.

ankitcse07:
Let me know when patch is ready

ankitcse07:
Hi Rohan Any updates on this fix.  This might become blocking issue for us in near future

gc00:
@rohgarg, @karya0:  @jiajuncao and I were discussing this issue.  We think we have a simpler solution that still uses most of the code here.

Right now, this code tries to patch each of the user locks at the time of restart.  To do that, the code needs to capture and remember all locks created by the end user.  It does this with a wrapper around pthread_mutex_lock().

Instead, let's patch the owner only during runtime (after a restart), and not at the time of restart itself.

The idea is to keep a map for mapMutexVirtOwner and wrappers for pthread_mutex_lock and pthread_mutex_unlock.  We don't need anything else.  (We don't do anything special during ckpt and restart.

Next, here is the pseudo-code:
```
pthread_mutex_lock(mutex):
  rc = _real_pthread_mutex_lock(mutex)
  mapMutexVirtOwner[mutex] = dmtcp_gettid() // virtual tid
  return rc

pthread_mutex_unlock(mutex):
  rc =  _real_pthread_mutex_unlock(mutex)
  if error(rc) == EPERM:
      if mapMutexVirtOwner[mutex] == dmtcp_gettid(): // same virtual tid as prior to ckpt
        // then it's a DMTCP issue; patch the owner
        patch_mutex_owner(mutex, kernel_gettid()) // set owner to kernel tid (same as what libc saved)
        rc =  _real_pthread_mutex_unlock(mutex)  // This second time, there will be no EPERM
        mapMutexVirtOwner[mutex] = 0 // not strictly necessary, but helpful for debugging
  return rc
```

We'd like to get comments on this.  If it looks good, whoever gets to this first can implement it.  I would argue that there are fewer possible sources of error in this pseudo-code, and it will be easier to debug if there is a new bug in the future.

ankitcse07:
@gc00 This seems to solve the problem in more cleaner way.  Also gene I guess wrappers are also required for pthread_cond_wait/pthread_cond_signal right ?

gc00:
@ankitcse07, @jiajuncao:  It's a good point about `pthread_cond_wait`/`pthread_cond_signal`.  I;ll check the glibc source code to see it these must be handled sepcially or if they go through pthread_mutex_lock/unlock.

Separately, I've been looking up more on this issue, and there is a second issue:  robust mutexes.  At the time that a thread exits, _if_ the thread is a robust mutex, then it will be located on a per-thread list:  I'm still looking at whether this affects us.  Here are some links to robust mutexes:
https://www.kernel.org/doc/Documentation/robust-futexes.txt and http://man7.org/linux/man-pages/man2/get_robust_list.2.html .

I suspect that you're not seeing this issue of robust mutexes.  If you are, we'll start a second issue and pull request for that.
@jiajuncao and @ankitcse07:  Are you able to get the attributes of the thread for which the mutex lock/unlock is falling?  If so, is the "robust" attribute set?  I have to go somewhere now, but I'll come back to this discussion later.

gc00:
@rohgarg:  This PR seems to now be accomplished by PR #282 using the alternative algorithm descrribed in the pseudo-code above.  If you agree with this, could you close this PR, and withdraw this branch without committing?  Thanks.

PR #176, opened 2015-08-19
Use alarm() for interval checkpointing.

karya0:
This replaces the existing somewhat complex code with (hopefully) a much simpler and less error prone one.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -438,8 +438,10 @@ static bool uniqueCkptFilenames = false;
  */
 static uint32_t theCheckpointInterval = 0; /* Current checkpoint interval */
 static uint32_t theDefaultCheckpointInterval = 0; /* Reset to this on new comp. */
-static struct timespec startTime;
 static bool isRestarting = false;
+static bool timerExpired = false;
+
+static void resetCkptTimer();
 
 const int STDIN_FD = fileno ( stdin );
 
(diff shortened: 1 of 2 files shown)


karya0:
@rohgarg: @gc00: Can you take a look at this one?

src/dmtcp_coordinator.cpp
```diff
@@ -1554,18 +1556,26 @@ void DmtcpCoordinator::writeRestartScript()
...
 }
 
-static void SIGINTHandler(int signum)
+static void signalHandler(int signum)
 {
-  prog.handleUserCommand('q');
+  if (signum == SIGINT) {
+    prog.handleUserCommand('q');
+  } else if (signum == SIGALRM) {
+    timerExpired = 1;
```
gc00: Shouldn't this be:  `timerExpired = true;`?  (Too much thinking in C. :-) )

gc00:
I had one minor "line note".  Once that is fixed, go ahead and push it in.

(I like it.  It's definitely more straightforward.)

gc00:
@karya0, Before you push it in, could you do:  `grep onTimeoutInterval  src/*.cpp`.  There are some comments in `src/dmtcp_coordinator.cpp` that need to be updated.  (And please do: `git pull --rebase` before the `grep`.  I created an accidental conflict (long story), and you'll find another reference to `onTimeoutInterval` there.

PR #177, opened 2015-08-22
Make --enable-debug work again

gc00:
@karya0 and @rohgarg, There is one issue to note here, in a fix to test/Makefile.in.  The `--enable-debug` flag was intended to set the `DEBUG` macro in the `src` directory.  But since `DEBUG` is also used in the test directory, and maybe elsewhere, we end up turning on `DEBUG` where we don't intend to.

This bug fix adds `-UDEBUG` to `CFLAGS` in `test/Makefile.in`, to turn off any previous `-DDEBUG` flag.

 A different and probably better bug fix would be to create a configure variable, `@DEBUG@`.  We would then write:

    CFLAGS = @DEBUG@ @CFLAGS@
in `src/Makefile.in` and all of its subdirectories.  But in `test/Makefile.in`, we would use only:

    CFLAGS = @DEBUG@ @CFLAGS@
Then, `test/Makefile.in` would set `@DEBUG@` instead of modifying `@CFLAGS@`.  This alternative bug fix actually seems like a better one to me.  But before I do that, I'd like to get a review, and agreement on this solution.  Thanks for reviewing this.

### DIFF
--- a/test/Makefile.in
+++ b/test/Makefile.in
@@ -4,7 +4,10 @@ OPENMP_CFLAGS=@OPENMP_CFLAGS@
 OPENMP_CXXFLAGS=@OPENMP_CXXFLAGS@
 CC = @CC@
 M32=@M32@
-CFLAGS = @CFLAGS@
+# Unfortunately, there is a name clash for 'DEBUG'.
+# --enable-debug intends to set 'DEBUG' in src, but sets it everywhere.
+# Here in 'test', we use 'DEBUG' for one-off tests.  So, we unset 'DEBUG'.
+CFLAGS = @CFLAGS@ -UDEBUG
 CPPFLAGS = @CPPFLAGS@
 CXX = @CXX@
 CXXFLAGS = @CXXFLAGS@


rohgarg:
Looks good to me.

karya0:
I am not sure if `--enable-debug` should apply only to the `src` directory. We should use it for `include`, `jalib`, `plugin`, and `contrib`directories as well. As for the `test` directory goes, there are only two tests -- `pty2` and `timer1` that use the `DEBUG` macro. This should be easy to fix. Unless, I am missing something here, I think it might be better to keep the DEBUG flag as is.

karya0:
If we want an equivalent flag, we can use `DEBUG_TEST` for the one-off tests.

gc00:
Then I propose that we go with the commit listed here.  (In the test directory, we undefine `DEBUG` in `test/ Makefile`.)

We could go with the alternative that you propose:  replace `DEBUG` by `DEBUG_TEST` everywhere in the test subdirectory.  But then I predict that someone will forget, and will add `DEBUG` to another file in the `test` directory.

So, `-UDEBUG` in `test/ Makefile` creates a cutout in which `--enable-debug` will apply in all directories except the `test` directory.  This is the current pull request.

Shall I push in this pull request, as is?

gc00:
@karya0 and I talked.  I'll ship this.

PR #179, opened 2015-08-23
Minor polishing suggested by warnings from clang

gc00:
All of these changes were suggested by warnings when compiling
with the clang-3.4 compiler.  Most are self-explanatory.

Here are reasons for two of the more obscure changes:

1. shareddata.cpp :  size_t -> int
    size_t is an unsigned data type.  So, the i>=0 condition is always true.
    This is a bug, since we want to allow i-- to cause i to become negative, and return false in that case.

2. procselfmaps.cpp
    ProcSelfMaps::ProcSelfMaps() initializes the private field members in a certain order.  But in include/procselfmaps.h, those fields were declared in a different order.
    Clang warns about this.  Now the initialization respects the order in which the fields were declared.

### DIFF
--- a/src/mtcp/mtcp_check_vdso.ic
+++ b/src/mtcp/mtcp_check_vdso.ic
@@ -94,7 +94,7 @@ static void * get_at_sysinfo(char **environ) {
   stack = (void **)&my_environ[-1];
   if (*stack != NULL) {
     MTCP_PRINTF("This should be argv[argc] == NULL and it's not.\n"
-	"NO &argv[argc], stack: %p\n", stack);
+        "NO &argv[argc], stack: %p\n", stack);
     mtcp_sys_exit(1);
   }
 #endif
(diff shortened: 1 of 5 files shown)


src/mtcp/mtcp_check_vdso.ic
```diff
@@ -335,11 +335,12 @@ void mtcp_check_vdso(char **environ)
          */
 #define ARCH_STACK_DEFAULT_SIZE (256 * 1024 * 1024)
 
+        int rlimit_failed = 0;
```
rohgarg: nit: Can you please fix the indentation on this?

gc00:
@rohgarg, I fixed the indentation.  mtcp_check_vdso.ic was inconsistent about whether to use TAB or 8&nbsp;spaces before.  I added a commit to use 8&nbsp;spaces always.

rohgarg:
Looks good to me.

PR #181, opened 2015-08-23
Fix for bug, under NFS, exposed by shared-memory1

gc00:
This bug fix will address the first half of issue #167 (shared-memory1 on NFS machines).  It does *not* address the second half of the issue (from the 16-core test machine).  (The second half of the issue concerns a spurious JWARNING.) 

This is a patch to the logic of commit e03f2727ef8fdde45d79 in pull request #170.

@karya0, could you please ta look at this one-line fix, since it concerns your code?  Thanks.

### DIFF
--- a/src/plugin/ipc/file/fileconnection.cpp
+++ b/src/plugin/ipc/file/fileconnection.cpp
@@ -676,7 +676,7 @@ void FileConnection::refill(bool isRestart)
 
   if (!_ckpted_file) {
     int tempfd;
-    if (_type == FILE_DELETED && (_flags & O_WRONLY)) {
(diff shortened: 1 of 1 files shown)


gc00:
**QUESTION:**  Note that there is a separate unresolved issue.  If you manually do:

    bin/dmtcp_launch -i5 test/shared-memory1
    # ls -a ckpt_*_files
    # There is a stray .nfs file that got saved
    ./dmtcp_restart_script.sh
    # Now, there is a .nfs file in the current directory.
So, apparently, we are saving and recreating .nfs* files when we should not be doing that.  There are other places in fileconnection.cpp that refert to `O_WRONLY`, but not `O_RDWR'.  Is this a bug?

gc00:
LGTM (alongside @karya0).

PR #185, opened 2015-08-28
clang-3.4 failed on launch, & on restart; fixed

gc00:
This is joint work between Pranay Surana and myself.  There were two bugs, due to weaker support from clang.  This provides workarounds.  It was tested against clang-3.4.

1. dmtcp.h : clang-3.4.1 was converting some static local variables into weak variables when they were inside the function `dmtcp_event_hook()`, which is declared weak.  It appears that clang-3.4.2 fixes this.

2. mtcp_restart.c : This is compiled with -O0.  Apparently, clang uses the ebp/rbp register in the epilog to `estart_fast_path()`, and we were zeroing it out just before returning.

### DIFF
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -338,6 +338,63 @@ EXTERNC void dmtcp_plugin_enable_ckpt(void);
     }                                                                       \
   } while (0)
 
+#if defined(__clang__) && __clang_major__ < 3 || \
+    __clang_major__ == 3 && __clang_minor__ <= 4
+/***************************************************************************
+ * This workaround is required for what is arguably a bug in clang-3.4.1 
+ *   under Ubuntu 13.10.
+ * We don't see a problem with clang-3.4.2 under Ubuntu 14.04.  So, eventually
+ *   we can deprecate this patch, when most distros use a later clang.
+ * clang-3.4 declares fn and dmtcp_event_hook as weak symbols ("V")
+ *   when these variables are delcared inside the function dmtcp_event_hook().
+ *   This workaround declares them outside of dmtcp_event_hook().
+ *   If the bug in clang gets fixed, we should expand this macro inline
+ *     for the non-clang case.
+ ***************************************************************************/
+# define DECLARE_TYPEOF_FNC(fnc_type,fnc) \
+static __typeof__(&fnc_type) fnc          \
+      = (__typeof__(&fnc_type)) -1   
+
+// For clang, declare these at top level, instead of inside a function.
+DECLARE_TYPEOF_FNC(dmtcp_event_hook,fn);
+DECLARE_TYPEOF_FNC(dmtcp_event_hook,_real_dmtcp_event_hook);
+# undef DECLARE_TYPEOF_FNC
+// This removes the declarations from NEXT_FNC2() and DMTCP_NEXT_EVENT_HOOK()
+// Those macros are invoked from inside dmtcp_next_event_hook(), which
+//   is declare in dmtcp.h as a weak function.
+//   clang-3.4.1 seems to declare these inner fnc pointer types as weak
+//     because the outer function is weak.  Arguably, this is a bug.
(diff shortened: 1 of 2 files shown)


gc00:
For `DECLARE_TYPEOF_FNC()`, an alternative is to have it expand to something that uses the explicit type of dmtcp_event_hook instead of `__typeof(dmtcp_event_hook)`.  If this gets aound the clang bug, this will be a simpler fix.

Also, note:  delcaration -> declaration (in the comment).

gc00:
@karya0, I tried the alternative that you suggested (and that I described in the previous comment).  The alternative, hard-wiring `__typeof(dmtcp_event_hook)`, still results in hitting the clang bug.

Do you agree that we should push in the current pull request, as is?

*Here are the gory details:*<br>Since the clang bug is only for their version 3.4, I propose that we keep the current pull request as is.  Even if there might be a more elegant workaround, we don't need to spend time looking for a more elegant workaround for the clang bug, when in a few years we'll just drop this code entirely and recommend that people use a recent version of clang.

Here is the code for the hard-wired patch that you proposed.  It still results in clang producing things like `V _ZZ16dmtcp_event_hookE22_real_dmtcp_event_hook` (as output of `nm` on lib*.so). 
```C
#if defined(__clang__) && __clang_major__ < 3 || \
    __clang_major__ == 3 && __clang_minor__ <= 4
/***************************************************************************
 * This workaround is required for what is arguably a bug in clang-3.4.1 
 *   under Ubuntu 13.10.
 * We don't see a problem with clang-3.4.2 under Ubuntu 14.04.  So, eventually
 *   we can deprecate this patch, when most distros use a later clang.
 * clang-3.4 declares fn and dmtcp_event_hook as weak symbols ("V")
 *   when these variables are delcared inside the function dmtcp_event_hook().
 *   This workaround declares them outside of dmtcp_event_hook().
 *   If the bug in clang gets fixed, we should expand this macro inline
 *     for the non-clang case.
 ***************************************************************************/

# define NEXT_dmtcp_event_hook()                                            \
  ({                                                                        \
     /* static __typeof__(&func) _real_##func = (__typeof__(&func)) -1; */  \
     typedef void (*dmtcp_event_hook_t)(DmtcpEvent_t, DmtcpEventData_t *);  \
     static dmtcp_event_hook_t _real_dmtcp_event_hook =                     \
       (dmtcp_event_hook_t ) NULL;                                          \
     if (_real_dmtcp_event_hook == (dmtcp_event_hook_t) NULL) {             \
       if (dmtcp_prepare_wrappers) dmtcp_prepare_wrappers();                \
       __typeof__(&dlsym) dlsym_fnptr;                                      \
       dlsym_fnptr = (__typeof__(&dlsym)) dmtcp_get_libc_dlsym_addr();      \
       _real_dmtcp_event_hook = (dmtcp_event_hook_t) (*dlsym_fnptr) (RTLD_NEXT, 
"dmtcp_event_hook");\
     }                                                                      \
   _real_dmtcp_event_hook;})

# undef DMTCP_NEXT_EVENT_HOOK
# define DMTCP_NEXT_EVENT_HOOK(event, data)                                 \
  do {                                                                      \
    /* static __typeof__(&dmtcp_event_hook) fn                              \
        = (__typeof__(&dmtcp_event_hook)) -1; */                            \
    typedef void (*dmtcp_event_hook_t)(DmtcpEvent_t, DmtcpEventData_t *);   \
    static dmtcp_event_hook_t fn = (dmtcp_event_hook_t ) NULL;              \
    if ((void*) fn == (void*) NULL) {                                       \
      fn = NEXT_dmtcp_event_hook();                                         \
    }                                                                       \
    if (fn != NULL) {                                                       \
      (*fn) (event, data);                                                  \
    }                                                                       \
  } while (0)

// End of patches for clang-3.4.1
#endif
```

gc00:
@karya0 and I talked.  We agreed to ship this.

PR #186, opened 2015-09-01
Fix markdown files; ```...''' is github-specific

gc00:
The markdown files have been made generic:

``````
by converting the ```...``` syntax
into the more standard <BLANK_LINE>...<BLANK_LINE> syntax.
``````
- This has been checked against the markdown preprocessors in github and doxygen.

Note:

``````
   The github-specific ```bash<NEWLINE> <BASH CODE>``` syntax was retained.
``````

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -5,25 +5,22 @@
 ### Build instructions
 
 To build, use:
-```
-  ./configure
-  make
-  make check [Optional]
-```
+
+    ./configure
+    make
+    make check [Optional]
 
 As usual, you may prefer to use `make` with `'-j'` for faster, parallel builds.
 
 `make check2`, and `make check3` will exercise further examples in the
 test subdirectory:
-```
-  check2: readline.c
-  check3: shared-memory.c
-```
+    check2: readline.c
+    check3: shared-memory.c
 
 This software runs in the original directory, and
-```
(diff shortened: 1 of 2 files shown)


gc00:
Kapil and I talked.  We decided:  make the triple backquote/bash work for github, and then ship it.

gc00:
@karya0, Could you check again?  I put the following markdown code into this github comment (click on "edit" to see the source), and it does indeed show bash-style syntax highlighting under github.  Since it also displays reasonably under doxygen (with "bash" appearing explicity), and since github markdown is our higher priority, I propose that this is good to go.

2. Build from source (Excerpted markdown code from INSTALL.md):

       ```bash
       ./configure
       make
       # Do NOT do 'make clean' here.  You will lose bin/dmtcp_launch, etc.
       ./configure --enable-m32
       make clean   # This won't remove the 64-bit binary and library files.
       make
       # Optionally, change 'make' to 'make install'' if desired.
       ```

gc00:
In the bash quote above, github works fine with at most 6 spaces at the head of the line.  Doxygen wants at least 7 spaces to put it in a special display.  (Probably, as soon as github sees the 4-space indent, it quotes everything instead of applying the triple-backquote syntax.

PR #189, opened 2015-09-14
Allow dmtcp_restart to be invoked by root.

gc00:
- Previously, the --run-as-root flag was needed to allow this.
- The --run-as-root flag is still needed, if the uid of the ckpt image
  doesn't match the user running the process.

In this code, a warning is issued if the user runs dmtcp_restart as root, and if ``--run-as-root`` was not set.  Previously, it would abort.

The flag ``--run-as-root`` may not the right name now.  What should we change it to?

Right now, the warning gets printed even if ``--quiet`` was requested.  Should we print nothing if ``--quiet`` was requested, or only if ``--run-as-root`` was requested?

@karya0, this is the change to DMTCP that you had requested.  This is just a rough version.  If you tell me how you would prefer to change this, I'm happy to do so.


src/dmtcp_restart.cpp
```diff
@@ -695,9 +695,9 @@ int main(int argc, char** argv)
 
   if (!runAsRoot && (getuid() == 0 || geteuid() == 0)) {
     JASSERT_STDERR <<
-      "Running dmtcp_restart as root is dangerous.  Aborting.\n"
-      "If you still want to do this (at your own risk), then use\n" \
-      "    dmtcp_restart --run-as-root\n";
+      "WARNING:  Running dmtcp_restart as root can be dangerous.\n"
+      "  An unknown checkpoint image or bugs in DMTCP may lead to unforeseen\n"
+      "  consequences.  Continuing as root ....\n";
     exit(0);
```
karya0: We need to remove this exit(0) :-).

karya0:
How about simply dropping ``-run-as-root` flag. We already have a `--no-strict-uid-checking` flag for handling uids. We can probably make it a bit more generic or leave it as is.

src/dmtcp_restart.cpp
```diff
@@ -725,9 +725,9 @@ int main(int argc, char** argv)
       printf("\nProcess uid (%d) doesn't match uid (%d) of\n" \
              "checkpoint image (%s).\n" \
 	     "This is dangerous.  Aborting for security reasons.\n" \
-             "If you still want to do this (at your own risk),\n" \
-             "  then modify dmtcp/src/%s:%d and re-compile.\n",
-             getuid(), buf.st_uid, restorename.c_str(), __FILE__, __LINE__ - 7);
+             "If you still want to do this, then re-run dmtcp_restart\n" \
+             "  with the --run-as-root flag.\n",
```
karya0: replace `--run-as-root` with `--no-strict-uid-checking` in the message.

gc00: I've now replaced --run-as-root with --no-strict-checking in the second commit.  (This line note refers to the first commit.)

gc00:
@karya0, I've now addressed both your comments.  I removed the ``exit(0);``.  (Thanks for catching that.)  Second, I decided to generalize ``--no-strict-uid-checking`` to ``--no-strict-checking``, and ``--run-as-root`` has been replaced by ``--no-strict-checking`` in the second line note comment that you inserted.

rohgarg:
Looks good to me.

karya0:
LGTM

PR #192, opened 2015-09-17
fixup! Use alarm() for interval checkpointing.

### LINKED ISSUE #191: Interval checkpointing doesn't work when --daemon and --exit-on-last are both set for coordinator
In this case, only manual checkpoint can be used. To reproduce the bug:

```bin/dmtcp_coordinator --daemon --exit-on-last```
```bin/dmtcp_launch -i 5 test/dmtcp1```

rohgarg:
Fixes #191


karya0:
Thanks for the fix Rohan! That was clearly an oversight on my end :). Let's merge it.

PR #193, opened 2015-09-17
dmtcp_launch --quiet is now truly quiet

gc00:
- The problem was that launch called the coordinator, which  was not quiet.
- The first step was to revert most of commit 7bdf16b3, which had
    made the --quiet flag do nothing.
- The second step was to look for 'JASSERT_STDERR <<' in dmtcp_coordinator.cpp
    and wrap it in 'if (! quiet)'.

**QUESTION:**
In coordinatorapi.cpp::439, we always pass on the flat ``--quiet`` if ``dmtcp_launch`` is starting a new coordinator.  We do this because the coordinator will be in the same window as the target process, and we don't want the coordinator output interfering.  We did this even if `dmtcp_launch` was not called with ``--quiet``.  I like this policy, and this PR continues that policy.  Are we agreed on this?

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -120,6 +120,8 @@ static const char* theUsage =
   "  -i, --interval (environment variable DMTCP_CHECKPOINT_INTERVAL):\n"
   "      Time in seconds between automatic checkpoints\n"
   "      (default: 0, disabled)\n"
+  "  -q, --quiet \n"
+  "      Skip startup message.\n"
   "  --help:\n"
   "      Print this message and exit.\n"
   "  --version:\n"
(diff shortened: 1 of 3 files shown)


src/dmtcp_coordinator.cpp
```diff
@@ -120,6 +120,8 @@ static const char* theUsage =
   "  -i, --interval (environment variable DMTCP_CHECKPOINT_INTERVAL):\n"
   "      Time in seconds between automatic checkpoints\n"
   "      (default: 0, disabled)\n"
+  "  -q, --quiet \n"
+  "      Skip copyright notice.\n"
```
karya0: We don't print the copyright notice at all, so let's remove this statement. We can replace it with something else that mentions the quietness.

karya0:
LGTM module the comment about help message above.

gc00:
@karya, I fixed the help message of dmtcp_coordinator.  (Thanks for catching that.)

I then realized that if the coordinator starts up in a separate terminal, then ``dmtcp_coordinator --quiet`` was not suppressing JNOTE.  I added that logic now (and "--quiet --quiet" suppresses JWARNING).  This is a second commit.  Could you (or somebody else) now review the second commit.

rohgarg:
Looks good to me.

PR #194, opened 2015-09-18
Aarch64 no gzip

gc00:
We have a long-standing bug for DMTCP with --gzip under aarch64.  Until we find the time to look at it carefully, we should automatically turn off --gzip.  This patch does that for aarch64..

### DIFF
--- a/src/dmtcp_launch.cpp
+++ b/src/dmtcp_launch.cpp
@@ -231,6 +231,7 @@ static void processArgs(int *orig_argc, char ***orig_argv,
       allowedModes = COORD_JOIN;
       shift;
     } else if (s == "--gzip") {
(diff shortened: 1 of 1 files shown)


src/dmtcp_launch.cpp
```diff
@@ -330,6 +331,23 @@ static void processArgs(int *orig_argc, char ***orig_argv,
       break;
     }
   }
+#define __aarch64__ 1
```
karya0: That doesn't look correct (a leftover from testing?) :)

gc00:
Thanks, @karya0.  You're right.  I had a leftover line of code from testing.

I believe it's fixed now.  Could you re-review it?

rohgarg:
Looks good to me.

rohgarg:
Looks good to me.

PR #196, opened 2015-09-18
mtcp_restart: fix the logic for mmap-ing truncated files

rohgarg:
This patch fixes an issue reported by Ankit. The problem was that We
were hitting an mtcp_abort() in adjust_for_smaller_file_size() because
MTCP was failing to restore a memory region where the application had
mmap-ed /dev/zero, probably to emulate a fast calloc().

The original intent of the function was to figure out if the backing
file of an mmap-ed region has truncated since checkpoint time, compute
the difference, and then patch the memory region using MAP_ANONYMOUS.

The logic in the function is difficult to follow, but, more importantly,
it fails to correctly handle the situation when the size of the backing
file of an mmap-ed region is 0, which is the case with /dev/zero.

This patch simplifies the code and fixes this issue by: a) mmap-ing an
entire region as MAP_ANONYMOUS if its backing file has truncated since
checkpoint time, and b) avoiding calls to the function.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -113,7 +113,9 @@ static RestoreInfo rinfo;
 /* Internal routines */
 static void readmemoryareas(int fd);
 static int read_one_memory_area(int fd);
+#if 0
 static void adjust_for_smaller_file_size(Area *area, int fd);
+#endif
 static void restorememoryareas(RestoreInfo *rinfo_ptr);
 static void restore_brk(VA saved_brk, VA restore_begin, VA restore_end);
 static void restart_fast_path(void);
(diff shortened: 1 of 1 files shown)


karya0:
LGTM.

gc00:
@rohgarg, could you add a comment: /* deprecated */ near the function definition for ``adjust_for_smaller_file_size()``?  This will remind us to remove this dead code at the next major release, if there have been no problems.

Aside from that, I agree:  LGTM.

src/mtcp/mtcp_restart.c
```diff
@@ -113,7 +113,9 @@ static RestoreInfo rinfo;
 /* Internal routines */
 static void readmemoryareas(int fd);
 static int read_one_memory_area(int fd);
+#if 0
```
karya0: Why not simply remove this block? We can get it from git history if we want to.

src/mtcp/mtcp_restart.c
```diff
@@ -892,11 +908,13 @@ static int read_one_memory_area(int fd)
       mtcp_abort ();
     }
 
+#if 0
```
karya0: Remove this?

src/mtcp/mtcp_restart.c
```diff
@@ -928,6 +946,7 @@ static int read_one_memory_area(int fd)
   return 0;
 }
 
+#if 0
```
karya0: Remove this?

rohgarg:
@ankitcse07: Could you please check if this works for you?

gc00:
I discussed this with @rohgarg.  The ``#if 0`` branch is truer to maintaining the user's view of ``/proc/*/maps``, but it has some bugs/unfinished code.  Let's keep this, but explain in a comment what I mentioned in the first sentence.

ankitcse07:
This is working fine now. Thanks for the fix. When will this fix be part of master branch ?

PR #197, opened 2015-09-20
Default for epoll_wait ret shouldn't assume ckpt

gc00:
This is a two-line pull request.  There was a danger of a stray signal (signal 0, SIGWINCH) being interpreted by the DMTCP coordinator as a request to perform a checkpoint.

Could someone check my reasoning here?


karya0:
LGTM. Although, ideally, we should simply block all signals except SIGALRM and SIGKILL and not worry about interrupts at all.

gc00:
Done.

Issue #190, opened 2015-09-14
DMTCP doesn't work with tclsh:


karya0:
[As reported by Linh Phan]

   I just noticed that dmtcp_launch does not work with tclsh:

``` 
phan-virtual-machine:/usr/local/src/dmtcp-2.4.1> dmtcp_launch /opt/ActiveTcl-8.6/bin/tclsh
dmtcp_coordinator starting...
    Host: phan-virtual-machine (127.0.1.1)
    Port: 7779
    Checkpoint Interval: disabled (checkpoint manually instead)
    Exit on last client: 1
Backgrounding...
```
dmtcp_launch just hangs here forever ....
 
But dmtcp_launch does work with tclsh (ActiveTcl-8.4 version):
 
```
phan-virtual-machine:/usr/local/src/dmtcp-2.4.1> dmtcp_launch /opt/ActiveTcl-8.4/bin/tclsh
dmtcp_coordinator starting...
    Host: phan-virtual-machine (127.0.1.1)
    Port: 7779
    Checkpoint Interval: disabled (checkpoint manually instead)
    Exit on last client: 1
Backgrounding...
% puts "No hanging issues"
No hanging issues
%
```

gc00:
@karya0, I started to take a look at this with Tcl-8.6 on our 16-core test machine.  I show how to set up the gdb environment, what to execute, and then how to see the infinite loop, which manifests as a hang.  I did:
```
(gdb) b main
(gdb) run
(gdb) cont
(gdb) break fork
(gdb) cont
# and continue a few more times to get past all the wrappers for fork.
```

You'll then see:
```
(gdb) thr 4
[Switching to thread 4 (Thread 0x7fb7ff2000 (LWP 22602))]
#0  0x0000007fb7c2961c in do_futex_wait ()
   from /lib/aarch64-linux-gnu/libpthread.so.0
(gdb) where
#0  0x0000007fb7c2961c in do_futex_wait ()
   from /lib/aarch64-linux-gnu/libpthread.so.0
#1  0x0000007fb7c296cc in sem_wait@@GLIBC_2.17 ()
   from /lib/aarch64-linux-gnu/libpthread.so.0
#2  0x0000007fb7e90d38 in dmtcp::ThreadList::init () at threadlist.cpp:172
#3  0x0000007fb7e874f0 in dmtcp::initializeMtcpEngine ()
    at mtcpinterface.cpp:56
[...]

PR #198, opened 2015-09-20
Fix for initThread on fork (bug seen w/ tclsh-8.6)

gc00:
There is a long-standing bug in how DMTCP does fork().  It doesn't properly re-initialize the descriptor of the thread that called fork().  Presumably, the bug did not manifest earlier because most code calls exec() immediately after fork(), and thread descriptors are then re-initialized once more.

On fork(), the thread caller continues to live on, and so the thread-local variable ``threadlist.cpp:curThread`` of the first thread is using its old (stale) thread descriptor instead of a new one.  The memory of the old thread descriptor is also being re-allocated for the second thread (the checkpoint thread).

This bug manifests with tclsh version 8.6 (but not with version 8.5).
The bug report is described in issue #190.


gc00:
I've now also pushed in a second commit.  The second commit is a test for autotest.py that shows a hang on launch when this fix is not enabled.  In autotest.py, it fails the checkpoint test, and maybe that's okay for now.  But the real problem is that it hangs at the time of launch.  To see it hang, run:

    LD_LIBRARY_PATH=./test bin/dmtcp_launch test/pthread_atfork2

For comparison, `LD_LIBRARY_PATH=./text test/pthread_atfork2` will print numbers instead of hanging.

karya0:
LGTM. In my current work for barriers, this particular logic will get simplified a bit and so we should revisit this problem once I have the PR ready. In the meanwhile, let's ship it!

gc00:
@karya0, I'll push it in.  When the new PR is ready and we push it in, we should re-visit `test/pthread_atfork2.c`.  It has some comments about additional, related tests that are failing.

This closes issue #190.

Issue #205, opened 2015-10-07
Bug in SysvIPC plugin


rohgarg:
This is just a placeholder for now to track the bug in the `on_semget()` function; I'll fill in the details later today.

rohgarg:
Fixed via PR #207.

PR #207, opened 2015-10-07
sysvipc-plugin: Fix a bug in on_semget()

rohgarg:
The sysvipc class uses a STL map to map virtual semaphore
ids to semaphore objects. The on_semget() method, which is
called on every semget(), was using the the real semaphore
ids as keys to read the values, whereas the STL map maps
virtual sem. ids to semaphore objects.

This fixes issue #205.


karya0:
Can you also change the variable name in the header file? We can then merge it.

PR #208, opened 2015-10-08
Fix for bbb0332 (decl. of msgrcv for SLES10 lost)

gc00:
I added a small change to sysvipcwrappers.cpp.  This is a hack.  But for the 2.4.2 release, this is a minimal change, and we don't want to risk breaking DMTCP.

The story is that in the commit bbb0332, there was no chance to test for SLES10.  This commit
broke SLES10, which was using glibc version 2.4.  The declaration of msgrcv() changed after glibc-2.4.  Our wrapper for msgrcv() has to use either the SLES10 declaration, or the current POSIX declaration.

Prior to commit bbb0332, we had a scheme for doing this (`#define msgrcv msgrcv_glibc` before `#include <sys/msg.h>`, so that `msg.h` doesn't declare msgrcv().  Then our own wrapper can declare msgrcv() the same way for everybody.

This minimal fix brings back the logic prior to commit bbb0332.  After the 2.4.2 release, we can simplify the logic.  Currently, we have two DMTCP .h files that try to include almost everything available in glibc, and our .cpp files then include these "giant" .h file.  This is a bad idea.  The glibc include files should be included only where they are needed, and not almost everywhere.

### DIFF
--- a/src/plugin/pid/pidwrappers.h
+++ b/src/plugin/pid/pidwrappers.h
@@ -51,7 +51,16 @@ struct user_desc {int dummy;}; /* <asm/ldt.h> is missing in Ubuntu 14.04 */
 #include <sys/ipc.h>
 #include <sys/shm.h>
 #include <sys/sem.h>
-#include <sys/msg.h>
+// FIXME:   We define _real_msgctl() in terms of msgctl() here.  So,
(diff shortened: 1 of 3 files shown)


karya0:
LGTM. Can we also create an issue to do some configure-time checks and see if we mechanize it?

karya0:
Merged via 809b4f.

PR #212, opened 2015-10-15
Remove reference to deprecated install-hooks api in python example

kwharrigan:
823d096 removed the dmtcp_install_hooks function entirely, so this
should be gone as well.


karya0:
Thanks @kwharrigan. I have merged the commit.

PR #213, opened 2015-10-15
Clean up test/plugin/*/Makefile: make check

gc00:
We kill an old coordinator using dmtcp_command at the beginning of 'make check'.  But a JWARNING message goes to stderr.  This patch redirects:  2>/dev/null

An alternative would be to add a`--quiet` flag to `dmtcp_command`.  But that seems like overkill to me.


rohgarg:
LGTM

PR #214, opened 2015-10-15
Some preparation work for new barriers API

karya0:
@gc00, @rohgarg, @jiajuncao: Can you take a quick look at to see if there is something wrong.


gc00:
@karya0, my only important comment is the in-line comment on `shareddata.cpp`, where I write that a name change and/or a comment would clarify the meaning of `SharedData::suspended()`.  The other request is trivial.

After this is done, it all looks good to me.

Issue #215, opened 2015-10-21
Broken dmtcp_checkpoint() API


rohgarg:
This issue was reported separately by @kwharrigan and @ankitcse07.

As @kwharrigan has rightly noted:

The commit 823d096 removed all references and the definition of the
function `dmtcp::userHookTrampoline_postCkpt()`.  The removal of
`userHookTrampoline_postCkpt` also removed calls that incremented the
static global variables `numRestarts` and `numCheckpoints`. Consequently,
the calls to `dmtcp_checkpoint()` would get stuck in the following while
loop in `dmtcpplugin.cpp`.

```C
  if(dmtcpRunCommand('c')){ //request checkpoint
    //and wait for the checkpoint
    while(oldNumRestarts==numRestarts && oldNumCheckpoints==numCheckpoints){
      //nanosleep should get interrupted by checkpointing with an EINTR
error
      //though there is a race to get to nanosleep() before the checkpoint
      struct timespec t = {1,0};
      nanosleep(&t, NULL);
      memfence();  //make sure the loop condition doesn't get optimized
    }
    rv = (oldNumRestarts==numRestarts ? DMTCP_AFTER_CHECKPOINT :
DMTCP_AFTER_RESTART);
  }
```

@kwharrigan has suggested the following patch to fix the problem:

```diff
diff --git a/include/dmtcp.h b/include/dmtcp.h
index 39ca249..1a4b6ac 100644
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -417,4 +417,6 @@ DECLARE_TYPEOF_FNC(dmtcp_event_hook,_real_dmtcp_event_hook);
 /// Pointer to a "void foo();" function
 typedef void (*dmtcp_fnptr_t)(void);
 
+EXTERNC void increment_counters(int isRestart);
+
 #endif
diff --git a/src/dmtcpplugin.cpp b/src/dmtcpplugin.cpp
index 4c5167c..3b40fa1 100644
--- a/src/dmtcpplugin.cpp
+++ b/src/dmtcpplugin.cpp
@@ -500,3 +500,12 @@ EXTERNC int dmtcp_no_coordinator(void)
 {
   return CoordinatorAPI::noCoordinator();
 }
+
+EXTERNC void increment_counters(int isRestart)
+{
+  if (isRestart) {
+    numRestarts++;
+  } else {
+    numCheckpoints++;
+  }
+}
[...]

PR #217, opened 2015-10-27
Fix broken dmtcp_checkpoint() API

rohgarg:
The broken dmtcp_checkpoint() was reported separately by two DMTCP
users.

The commit 823d096 removed all references and the definition of the
function `dmtcp::userHookTrampoline_postCkpt()`.  The removal of
`userHookTrampoline_postCkpt` also removed calls that incremented the
static global variables `numRestarts` and `numCheckpoints`. Consequently,
the calls to `dmtcp_checkpoint()` would get stuck in a while loop in
`dmtcpplugin.cpp`.

This patch fixes the problem by adding a call to increment the global
counters post checkpoint.

Fixes issue #215.

### DIFF
--- a/src/dmtcpplugin.cpp
+++ b/src/dmtcpplugin.cpp
@@ -500,3 +500,12 @@ EXTERNC int dmtcp_no_coordinator(void)
 {
   return CoordinatorAPI::noCoordinator();
 }
+
+void dmtcp::increment_counters(int isRestart)
+{
+  if (isRestart) {
+    numRestarts++;
+  } else {
+    numCheckpoints++;
+  }
+}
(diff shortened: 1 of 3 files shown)


gc00:
Can we move the declaration of `increment_counters(int isRestart)` from `include/dmtcp.h` to `mtcpinterface.h`?  The file `include/dmtcp.h` is used by our end users for declaring DMTCP functions that they need to call.  The function `increment_counters(int isRestart)` is internal to DMTCP, and should not be exposed to the end user.

Second,  `increment_counters(int isRestart)` is declared as a C function.  But it manipulates the C++ variables numCheckpoints and numRestarts in `dmtcpplugin.cpp`, and it gets called from within the C++ function, `dmtcp::callbackPostCheckpoint()`.  So, why don't we make this a C++ function, and remove `EXTERNC`? (And of course, if it's C++, itshould be in the `dmtcp` namespace, just as the functions around it.

rohgarg:
Updated as per @gc00's comments.

gc00:
I assume you also tested the new version of the code.  Given all of that:
LGTM

PR #220, opened 2015-10-28
New barriers API

karya0:
It's a big PR, so take your time :-).


gc00:
I'll start with a first comment.  It's hard to read because the commits are listed out of order.  (More of the git/github complexity, I suppose.)  In order to make sense of this, I cloned your code, and did `git log --oneline`, and I get:
```
60bf4f4 Introduced the new barriers API.
7d32f62 Reverse the sequence of calls for post-ckpt events.
50d3cdf Call plugins event hooks using the new api.
b6458ac Preparing plugins for a newer interface.
3983a52 Added a new interface to the plugin API.
```
So, I'm going to review the commits as if they were checked in using that ordering (with the first line being the last commit).

gc00:
Second comment:  This depends on the earlier pull request:
```
Merge pull request #214 from karya0/prep-for-barriers
```
as given by `git log --oneline 98a67c6`. 

Let's agree to stop using github "Merge pull request", since github is replacing the original one-line commit message by github's preferred message about a merge request that refers to the original github repo branch name.  Note that the original github repo branch name is usually deleted by the user, so that the github commit message will be maximally cryptic.  The user can then decode the commit message by going to github's "pull request #214" in this example.

gc00:
Can we change `pluginmanager.{cpp,h}` to simply `plugin.{cpp,h}`?  Similarly, can we change the class `PluginManager` to a class `Plugin`?

It seems to me that the class concerns plugins in general.  When I think of a manager class, I assume that some object of type `Plugin` is talking to or being created by a class of type `PluginManager`.

gc00:
@karya0, I didn't try executing anything, since I assume that you've already done this.  So, this is a reading review of the code, only.

As we discussed, I'd like to see this code for barriers (and the previous code for pull request #214) go into a 3.x branch, which should be different from the 2.x branch.  There are many changes here, and we'll need more incubation time in the master branch.  Also, as we discussed, there may be some minor performance penalties associated with the initial implementation of barriers.  We can release that as 3.0, and then do 3.1 with the more efficient implementation.

I would suggest a 2.5 release by the end of 2015, and after that, all new commits can go into the 3.x branch.  If some users want the stability of 2.5, we can put some of the new commits both into 3.x and into 2.5.

Finally, we'll also need to revise the doc directory later to correspond to this.  We should also review the test/plugin directory, and add to it in order to illustrate how to use the new barriers.

karya0:
Duplicate of #291

Issue #218, opened 2015-10-27
Support CMA (Cross Memory Attach)


jiajuncao:
process_vm_readv() and process_vm_writev() were added in Linux 3.2 as a new method for data transfer between processes. Although they're nonstandard Linux extensions, some applications start to use them (e.g. MVAPICH2). We might need to support CMA eventually.

karya0:
It should just require us to add the two wrappers to the PID plugin, correct? In that case, it should be pretty straight forward.

gc00:
PR #221 fixes this issue.

PR #221, opened 2015-10-29
Added support for CMA (Cross Memory Attach)

jiajuncao:
Fix for #218


rohgarg:
@jiajuncao: Could you please add a test for this to the DMTCP test suite? Othewise, it looks good to me.

src/plugin/pid/pidwrappers.h
```diff
@@ -130,7 +135,9 @@ extern "C"
   MACRO(__xstat64)          \
   MACRO(__lxstat)           \
   MACRO(__lxstat64)         \
-  MACRO(readlink)
+  MACRO(readlink)           \
+  MACRO(process_vm_readv)   \
```
karya0: Can we put these MACROs in their own category and include them in PIDVIRT_GEN_ENUM only if the kernel version matches?

karya0:
LGTM modulo the comment about syscallwrapper.h

jiajuncao:
@karya0 Updated as suggested.

test/autotest.py
```diff
@@ -831,6 +831,7 @@ def saveResultsNMI():
 runTest("shared-memory2", 2, ["./test/shared-memory2"])
 S=DEFAULT_S
 
+runTest("cma",     2, ["./test/cma"])
```
gc00: Can we change the indentation of the second argument, the "2",
to correspond to the indentation of the later lines?  (I know this is picky.)

test/cma.c
```diff
@@ -0,0 +1,47 @@
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <linux/version.h>
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(3,2,0))
+#include <sys/uio.h>
+#endif
```
gc00: Can we indent the inner `#include`?

```C
#if (LINUX_VERSION_CODE >= KERNEL_VERSION(3,2,0))
# include <sys/uio.h>
#endif
```

gc00:
Please change the `#if` condition everywhere to also verify that the correct GLIBC is available.  The condition should be:
```C
#if  (LINUX_VERSION_CODE >= KERNEL_VERSION(3,2,0)) && __GLIBC_PREREQ(2,15)
```

gc00:
In addition to the conditoinals in cma.c, let's not build it if it can't run.  Otherwise, autotest would print that the cma test succeeds, when it's not really being executed.  To do this, add to `test/Makefile.in`:

After the logic with `DO_PTHREAD_ATFORK = yes`, add:
```Makefile
ifeq (${shell test `uname -r` \> 3.3.0 && echo true},true)
DO_CMA = yes
endif
```

And after the logic with `echo "FIXME: pthread_atfork is not being compiled for arm/aarch64"`, add:
```Makefile
# FIXME:  We should create a test in configure.ac to see if process_vm_XXX()
#   is available.
ifeq (${DO_CMA},yes)
cma: cma.c
        ${CC} ${CFLAGS} -o cma cma.c
else
cma: cma.c
        @ echo "#cma: Skipping test for cma.  CMA requires Linux 3.2 or later."
endif
```

And finally, we then need to modify autotest.py to verify that the cma executable exists:
```py
# Makefile compiles cma only for Linux 3.2 and higher.
if os.path.exists("cma"):
  runTest("cma",          2, ["./test/cma"])
```

jiajuncao:
```ifeq (${shell test `uname -r` \> 3.3.0 && echo true},true)
DO_CMA = yes
endif```

Not sure if this is the right way to do it.

**Comment by gc00** [comment]
@jiajuncao wrote:
`Not sure if this is the right way to do it.`

Let's talk (or else please write more here why it's not right :-) ).

**Comment by gc00** [comment]
Thanks for catching my bug.  (It does the wrong thing for version 3.11 in Linux.)  Next, could you test the following code?
```Makefile
HAS_PYTHON = @HAS_PYTHON@
ifeq ($(HAS_PYTHON),yes)
ifeq (${shell python -c "x='`uname -r`'; y='3.2.0'; print x.split('.') > y.split('.')"},True)
DO_CMA = yes
endif
endif
```

gc00:
@jiajuncao, As agreed, you can push in this pull request.  I'll then add a newer pull request in which I create HAS_CMA according to the pattern I describe above.

Issue #224, opened 2015-11-05
`dmtcp_get_local_status()` return values are not in sync with its documentation


rohgarg:
This looks like a documentation bug. The `dmtcp_get_local_status()` function returns 1 on success and 3 (`DMTCP_NOT_PRESENT`) on failure, but the documentation suggests that it should return 0 on success and -1 on failure.

gc00:
@rohgarg, Thanks for your quick response.  Could we make two more addition to your patch?  It's ugly to return `1` or `DMTCP_NOT_PRESENT`.  It should always return a macro.  (It's my fault.  I wrote the original code.)

Let's add after `#ifndef DMTCP_NOT_PRESENT` the following code.
```C
#ifndef DMTCP_IS_PRESENT
# define DMTCP_IS_PRESENT 4
#endif
```

Then, let's change the documentation of `dmtcp_get_local_status` to:
```C
 * + Returns DMTCP_IS_PRESENT on success and DMTCP_NOT_PRESENT on error.
```
and after that, let's change the logic to:
```C
  (dmtcp_get_local_status ? (dmtcp_get_local_status(c,r), DMTCP_IS_PRESENT) : \
                            DMTCP_NOT_PRESENT)
```

rohgarg:
Closed via PR #225

PR #225, opened 2015-11-05
dmtcp.h: Fix documentation of two *_get_status functions

rohgarg:
Fixes issue #224.

### DIFF
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -212,7 +212,9 @@ EXTERNC int checkpoint_is_pending(void) __attribute__((weak));
 
 /**
  * Gets the coordinator-specific status of DMTCP.
- * - Returns 0 on success and -1 on error.
+ * - Returns DMTCP_IS_PRESENT if running under DMTCP and DMTCP_NOT_PRESENT
+ *   otherwise.
+ * - Side effects: modifies the arguments
  *
  * Args:
  *   numPeers: Number of processes connected to dmtcp_coordinator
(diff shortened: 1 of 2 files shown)


gc00:
@rohgarg, Thanks for your quick response.  Could we make two more addition to your patch?  It's ugly to return `1` or `DMTCP_NOT_PRESENT`.  It should always return a macro.  (It's my fault.  I wrote the original code.)

Let's add after `#ifndef DMTCP_NOT_PRESENT` the following code.
```C
#ifndef DMTCP_IS_PRESENT
# define DMTCP_IS_PRESENT 4
#endif
```

Then, let's change the documentation of `dmtcp_get_local_status` to:
```C
 * + Returns DMTCP_IS_PRESENT on success and DMTCP_NOT_PRESENT on error.
```
and after that, let's change the logic to:
```C
  (dmtcp_get_local_status ? (dmtcp_get_local_status(c,r), DMTCP_IS_PRESENT) : \
                            DMTCP_NOT_PRESENT)
```

gc00:
And let's also change the documentation of `dmtcp_get_coordinator_status`in the same way:
```C
 * + Returns DMTCP_IS_PRESENT on success and DMTCP_NOT_PRESENT on error.
```
and after that, let's change the logic to:
```C
  (dmtcp_get_coordinator_status ? (dmtcp_get_coordinator_status(p,r), DMTCP_IS_PRESENT) : \
                            DMTCP_NOT_PRESENT)
```

gc00:
LGTM

PR #228, opened 2015-11-06
threadlist.cpp: Fix compilation error for prctl

rohgarg:
The issue arose due to a recent commit: 0523b46, in which
procname related code (from src/procname.cpp) was moved into the
stopthisthread signal handler (in src/threadlist.cpp), but the
relevant header file: `<sys/prctl.h>` that is required for prctl()
was never moved out of the conditional in threadlist.cpp. This
results in a compilation failure on RHEL-5.8.

### DIFF
--- a/src/threadlist.cpp
+++ b/src/threadlist.cpp
@@ -7,7 +7,7 @@
 #include <sys/resource.h>
 #include <linux/version.h>
 #include "config.h"
-#ifdef HAS_PR_SET_PTRACER
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(2,6,11) || defined(HAS_PR_SET_PTRACER)
 # include <sys/prctl.h>
 #endif
 #include "threadlist.h"


gc00:
@rohgarg, The conditional uses HAS_PR_SET_PTRACER, set in `include/config.h`.  Is it set to 1 or 0 for RHEL-5.8.  It's set based on a test of `prctl` in `configure.ac`.  If you perform that test manually, what do you see?  Thanks.

rohgarg:
@gc00: The header file `<sys/prctl.h>` needs to be included regardless.

All the previous references to `prctl()` in `threadlist.cpp` were under the `HAS_PR_SET_PTRACER`. The recent commit 0523b46 moved code in `src/procname.cpp` to `src/threadlist.cpp`. This introduced additional calls to `prctl()` that are not under the original conditional.

gc00:
@rohgarg wrote:

    The header file <sys/prctl.h> needs to be included regardless.
    All previous references to prctl() in threadlist.cpp were under the HAS_PR_SET_PTRACER conditional. The recent commit: 0523b46 moved code in src/procname.cpp to src/threadlist.cpp. This introduced additional calls to prctl() that are not under the said conditional.

I agree with this.  I guess what I'm asking is:  Shouldn't we protect all calls to `prctl` and `sys/prctl.h` with either:
`#ifdef HAS_PR_SET_PTRACER`
or with:
`#if LINUX_VERSION_CODE >= KERNEL_VERSION(2,6,11)`
but not with a mixture of the two.

I'm proposing to only use:
`#ifdef HAS_PR_SET_PTRACER`
everywhere, and to stop using the `KERNEL_VERSION(2,6,11)` check.  This should be more robust.  If you look at `configure.ac', you'll see the test to define `HAS_PR_SET_PTRACER`.  It runs an actual program with the header and with `prctl()`.  If the program exits successfully, then `HAS_PR_SET_PTRACER` is set to 1.

Isn't this more robust?  There's a weird history of when the kernel implements `prctl` and when glibc finally added an include file to support it.  So, a direct test with `HAS_PR_SET_PTRACER` seems to be the best.

I believe the reason that threadlist.cpp was using `KERNEL_VERSION(2,6,11)` is because it inherited it from some very old mtcp code.  The very old mtcp code had a Makefile, but no Makefile.in.  So, `#ifdef HAS_PR_SET_PTRACER` wasn't possible at that time.

rohgarg:
At this point, there are two users of `prctl()` in `threadlist.cpp`:

 `SET_PTRACER` -- this code has been around for a while

and

  `SET_NAME` -- this code was moved recently from `src/procname.cpp`

According to the man page, while `SET_PTRACER` was introduced in Linux-3.4, `SET_NAME` has been around since Linux-2.6.9.

Given this, I'd say that we keep the two users under separate conditionals, and include the header file (`sys/prctl.h`) regardless, or perhaps, under a third conditional that checks for its availablility.

gc00:
Given this, I'd say that we keep the two users under separate conditionals, and include the header file (sys/prctl.h) regardless, or perhaps, under a third conditional that checks for its availablility.

gc00:
Thanks for the clarification.  You're right.

Can we include `sys/prctl.h` in each of the two cases?  I'll propose the following:
```C
#if LINUX_VERSION_CODE >= KERNEL_VERSION(2,6,11) || defined(HAS_PR_SET_PTRACER)
// KERNEL_VERSION(2,6,11) checks for PR_GET_NAME; HAS_XXX for PR_SET_PTRACER
# include <sys/prctl.h>
#endif
```

gc00:
LGTM

PR #229, opened 2015-11-11
Add HAS_CMA to configure; fixes commit 0d8bf7150e3e

gc00:
Replaces
`#if (LINUX_VERSION_CODE >= KERNEL_VERSION(3,2,0)) && __GLIBC_PREREQ(2,15)`
with
`#ifdef HAS_CMA`

It also fixes a bug that crept into commit 0d8bf7150e3e.
<code>ifeq (${shell test \`uname -r\` > 3.2.0 && echo true},true)</code>
doesn't do the right thing for `uname -r` == 3.11.0


jiajuncao:
LGTM.

PR #230, opened 2015-11-11
Fix warnings in configure tests (include files)

gc00:
This pull request fixes some of the C programs being tested.  They were missing some include files.

The pull request is meant to be committed after PR #229.  For that reason, the configure.ac and configure file here also include changes made within PR #229.


configure.ac
```diff
@@ -339,6 +340,60 @@ if test "$has_pr_set_ptracer" == "yes"; then
 fi
 AC_MSG_RESULT([$has_pr_set_ptracer])
 
+dnl CMA brought in with  Linux 3.2.0 and glibc 2.15
```
karya0: Let's use the linkage with AC_CHECK_FUNC or similar.

gc00:
@karya0, the change you requested is in here now.  Could you review it?  (Should be fast.)

PR #231, opened 2015-11-11
Change size_t variable from %d to %zu (& %zd->%zu)

gc00:
size_t variable requires format of %zu.  One occurrence had `%zd` and the other occurrence had `%d`.

Note that while this is part of C99,  But C++ compilers tend to support this, since they need to read C-based include files.


rohgarg:
Looks good to me.

PR #233, opened 2015-11-18
This commit adds a new dmtcp API dmtcp_checkpoint_ckptDir which

ankitcse07:
takes checkpoint directory name as argument and stores all checkpoint
files for all processes at common place. This is necessary for
DMTCP Veloce Integration as while checkpointing Veloce Software
also takes HDL checkpoint which is spread across multiple sub-directories.
So in order to consolidate complete database, it is necessary that
all possible database related files get stored at one place. This
is already present in Veloce Specific checkpointing, however missing
in DMTCP application initiated checkpointing.

There is an API to set the checkpoint directory but this API
only sets checkpoint directory for the calling process. Its too
cumbersome to make all process call same API everytime when
new checkpoint activity is initiated. So we need a support which
globally changes the checkpoint directory for all processes.

### DIFF
--- a/plugin/unique-ckpt/unique-ckpt.cpp
+++ b/plugin/unique-ckpt/unique-ckpt.cpp
@@ -16,7 +16,13 @@ extern "C" int dmtcp_unique_ckpt_enabled(void)
 
 void updateCkptDir()
 {
-  const char *ckptDir = dmtcp_get_ckpt_dir();
+  const char *ckptDir = NULL; 
+  string globalCkptDir = dmtcp_get_global_ckpt_dir();
+  if(globalCkptDir.empty())
+    ckptDir = dmtcp_get_ckpt_dir();
+  else
+    ckptDir = globalCkptDir.c_str();
+
   string baseDir;
   if (strstr(ckptDir, dmtcp_get_computation_id_str()) != NULL) {
     baseDir = jalib::Filesystem::DirName(ckptDir);
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -369,6 +369,49 @@ void CoordinatorAPI::updateCoordCkptDir(const char *dir)
   _coordinatorSocket.writeAll(dir, strlen(dir) + 1);
 }
 
+void CoordinatorAPI::updateCkptDirGlobally(const char *dir)
+{
+  JASSERT(dir != NULL);
+
+  jalib::JSocket _coordinatorSocket1 = createNewSocketToCoordinator(COORD_ANY);
+  if (!_coordinatorSocket1.isValid()) {
+    return;
+  }
+
+  DmtcpMessage msg(DMT_UPDATE_CKPT_DIR_FOR_ALL_PROCESS);
+  msg.extraBytes = strlen(dir) + 1;
+  _coordinatorSocket1 << msg;
+  _coordinatorSocket1.writeAll(dir, strlen(dir) + 1);
+
+  msg.poison();
+  _coordinatorSocket1 >> msg;
+  msg.assertValid();
+  JASSERT(msg.type == DMT_UPDATE_CKPT_DIR_FOR_ALL_PROCESS_RESULT) (msg.type);
+
+  _coordinatorSocket1.close();
+}
+
+string CoordinatorAPI::getGlobalCkptDir(void)
+{
+  // FIXME: Add a test for make-check.
+  char buf[PATH_MAX];
+  if (noCoordinator()) return "";
+  DmtcpMessage msg(DMT_GET_GLOBAL_CKPT_DIR);
+  _coordinatorSocket << msg;
+
+  msg.poison();
+  _coordinatorSocket >> msg;
+  msg.assertValid();
+  JASSERT(msg.type == DMT_GET_GLOBAL_CKPT_DIR_RESULT) (msg.type);
+
+  buf[0] = '\0';
+
+  if(msg.extraBytes > 0) {
+    _coordinatorSocket.readAll(buf, msg.extraBytes);
+  }
+  return buf;
+}
+
 void CoordinatorAPI::sendMsgToCoordinator(const DmtcpMessage &msg,
                                           const void *extraData,
                                           size_t len)
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -369,7 +369,7 @@ void CoordinatorAPI::updateCoordCkptDir(const char *dir)
   _coordinatorSocket.writeAll(dir, strlen(dir) + 1);
 }
 
-void CoordinatorAPI::updateCkptDirGlobally(const char *dir)
+void CoordinatorAPI::setGlobalCkptDir(const char *dir)
 {
   JASSERT(dir != NULL);
 
@@ -391,10 +391,9 @@ void CoordinatorAPI::updateCkptDirGlobally(const char *dir)
   _coordinatorSocket1.close();
 }
 
-string CoordinatorAPI::getGlobalCkptDir(void)
+char* CoordinatorAPI::getGlobalCkptDir(void)
 {
-  // FIXME: Add a test for make-check.
-  char buf[PATH_MAX];
+  static char buf[PATH_MAX];
   if (noCoordinator()) return "";
   DmtcpMessage msg(DMT_GET_GLOBAL_CKPT_DIR);
   _coordinatorSocket << msg;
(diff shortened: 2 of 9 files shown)


ankitcse07:
I have a testcase for the same and can you provide you if required.

src/dmtcpplugin.cpp
```diff
@@ -130,6 +131,15 @@ EXTERNC int dmtcp_checkpoint()
   return rv;
 }
 
+EXTERNC int dmtcp_checkpoint_ckptDir(const char* ckptDir)
+{
+  if (ckptDir != NULL) {
+    CoordinatorAPI::instance().updateCkptDirGlobally(ckptDir);
+  }
+  int rv = dmtcp_checkpoint();
```
rohgarg: Unless I'm misunderstanding the purpose of the new API, I think we don't need to call `dmtcp_checkpoint()` here. A user can make an explicit call to `dmtcp_checkpoint()` when they want.

src/coordinatorapi.cpp
```diff
@@ -369,6 +369,49 @@ void CoordinatorAPI::updateCoordCkptDir(const char *dir)
...
+
+  msg.poison();
+  _coordinatorSocket1 >> msg;
+  msg.assertValid();
+  JASSERT(msg.type == DMT_UPDATE_CKPT_DIR_FOR_ALL_PROCESS_RESULT) (msg.type);
+
+  _coordinatorSocket1.close();
+}
+
+string CoordinatorAPI::getGlobalCkptDir(void)
```
rohgarg: I don't see a need for using a C++ string here. How about we change this to: `char* CoordinatorAPI::getGlobalCkptDir(void)`? This way we avoid having to make an extra function call (in `dmtcpworker.cpp`) to convert this back to a C-style string.

src/dmtcpplugin.cpp
ankitcse07: I think that's better idea we can simply create API to set checkpoint directory globally for all processes and then user can make explicit call to dmtcp_checkpoint whenever required.  I created single call for Veloce Integration but now it sounds logical to create one API which sets ckpt directory globally

src/coordinatorapi.cpp
ankitcse07: Yes I will incorporate this change

rohgarg: I don't have a strong opinion though. If it seems to be a better style and/or if you can anticipate complex string manipulations (for which a C++-style string is obviously better than having to manipulate raw char buffers) we can keep this. The only change I'd suggest in that case is to use the dmtcp alloc arena by switching to the dmtcp namespace, i.e., use: `dmtcp::string CoordinatorAPI::getGlobalCkptDir(void)` instead of: `string CoordinatorAPI::getGlobalCkptDir(void)`.

ankitcse07: I don't think to have any complex string manipulations for this function as this just acting as query to get the checkpoint directory name from coordinator. Although just curious do you have your own implementation of string in dmtcp namespace ? If yes what is the reason for the same ?

src/dmtcpplugin.cpp
rohgarg: Okay, and then rename the API to: `dmtcp_set_global_ckpt_dir()`?

ankitcse07: I will rename to above following name conventions to other APIs

src/coordinatorapi.cpp
```diff
@@ -369,6 +369,49 @@ void CoordinatorAPI::updateCoordCkptDir(const char *dir)
   _coordinatorSocket.writeAll(dir, strlen(dir) + 1);
 }
 
+void CoordinatorAPI::updateCkptDirGlobally(const char *dir)
```
rohgarg: nit: Change `updateCkptDirGlobally()` to `setGlobalCkptDir()`

ankitcse07:
Let me know if this looks good

rohgarg:
@ankitcse07: I must confess that I don't fully understand the usecase yet. In particular, I don't understand how is this different from the `--ckpt-dir` runtime option? Nevertheless, I have added some minor inline suggestions for the code per se. A testcase would definitely help clarify the usecase.

ankitcse07:
Using runtime option at the time of dmtcp_launch means that checkpoint directory is pre-decided right at the time of launch.  Consider a case in which user wants to store multiple checkpoint databases with different names in the same run and wants to control via API from its own source code.  For eg. in our case Emulation can run for days and user is asking for an API to integrate that in its own source code which periodically takes the checkpoint with different database names. This API will be helpful in those cases.

rohgarg:
I'm not sure if this fits your usecase but there's a `--enable-unique-checkpoint-filenames` that'll ensure that DMTCP creates unique files for successive checkpoints  (for long running applications).

Also, there's an existing `CoordinatorAPI` function: `updateCoordCkptDir()`. Why can't we just expose that through a new DMTCP API (`dmtcp_set_ckpt_dir()`) if it isn't already? In other words, call: `CoordinatorAPI::instance().updateCoordCkptDir(dir)` instead of: `CoordinatorAPI::instance().setGlobalCkptDir(dir)` from `dmtcp_set_ckpt_dir()`?

>Using runtime option at the time of dmtcp_launch means that checkpoint directory is pre-decided right at the time of launch. Consider a case in which user wants to store multiple checkpoint databases with different names in the same run and wants to control via API from its own source code. For eg. in our case Emulation can run for days and user is asking for an API to integrate that in its own source code which periodically takes the checkpoint with different database names. This API will be helpful in those cases.

ankitcse07:
I think updateCoordCkptDir is called from already existing API dmtcp_set_coord_ckpt_dir which just sets the directory for restart scripts and it doesn't change the place where ckpt files for each process are stored. Secondly using unique-checkpoint-filenames is not solving the purpose as it is just creating unique names for ckpt files, however I require them to consolidated in some different directory in which HDL checkpoint database would also be stored. 

Emulation checkpointing invokes following databases:
     1) HDL design checkpoint taken by our Veloce Software : Consist of multiple sub-directories
     2) C/C++ software layer checkpoint via DMTCP 

Now I have to associated (2) with (1), only can be to have same directory name for (1) and (2)

ankitcse07:
I have one more question on this does unique-file-names are unique across multiple runs of same application

rohgarg:
That sounds good. I have sent you a request separately.

> May be we can have small discussion on call to understand the scenario, time which suits you tomorrow. I can call you up, let me know your contact details.

ankitcse07:
Please see the last commit details #ec0c236

plugin/unique-ckpt/unique-ckpt.cpp
```diff
@@ -16,7 +16,13 @@ extern "C" int dmtcp_unique_ckpt_enabled(void)
 
 void updateCkptDir()
 {
-  const char *ckptDir = dmtcp_get_ckpt_dir();
+  const char *ckptDir = NULL; 
+  string globalCkptDir = dmtcp_get_global_ckpt_dir();
+  if(globalCkptDir.empty())
+    ckptDir = dmtcp_get_ckpt_dir();
```
rohgarg: What's the difference between `dmtcp_get_global_ckpt_dir()` and `dmtcp_get_ckpt_dir()`?

rohgarg: Nevermind, I figure it out. While the former queries the coordinator, the latter returns the directory that was set at launch time, if I'm reading this correctly.

rohgarg:
Thanks @ankitcse07. I'm beginning to see the usecase and the requirements now. I think this warrants a discussion over the phone because it affects some parts the DMTCP design. Hopefully, we'll converge to a clean and a general fix.

plugin/unique-ckpt/unique-ckpt.cpp
ankitcse07: Yes that's true

ankitcse07:
@rohgarg Ok I am setting a meeting for tomorrow. Let me send you details via mail

ankitcse07:
Thanks alot for meeting, I was able to create a plugin taking care of my requirements of common ckpt dir by storing information in coordinator via send key-value pair function. I will close this pull request as this no longer required and can create my custom plugin for the same.

gc00:
Excellent!

gc00:
@ankitcse07, I'm closing this pull request without merging it, based on your request:
>  I will close this pull request as this no longer required and can create my custom plugin for the same.

Thanks for the discussion.  It was helpful for us, too.

PR #236, opened 2015-11-20
execwrapper.cpp: Fix null pointer dereference when env is NULL

rohgarg:
This patch fixes a bug in the `copyEnv()` code where we
would dereference a NULL pointer if the environment
argument of an `execve()` call was NULL.

### DIFF
--- a/src/execwrappers.cpp
+++ b/src/execwrappers.cpp
@@ -407,8 +407,10 @@ static string getUpdatedLdPreload(const char* filename,
 static vector<string> copyEnv(char *const envp[])
 {
   vector<string> result;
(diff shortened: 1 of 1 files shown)


rohgarg:
Here's a simple testcase that demonstrates this bug:
```C
   int main()
   {
     int status;
     pid_t pid = fork();
     if (pid == 0) {
        char *myargv[] = {"/usr/bin/ls", NULL};
        int result = execve(myargv[0], myargv, NULL);
     } else if (pid > 0) {
       waitpid(pid, &status, 0);
     }
   }
```

gc00:
Looks good to me.

PR #237, opened 2015-11-21
Fix dmtcp_nocheckpoint

gc00:
dmtcp_nocheckpoint was unsetting LD-PRELOAD, even if the user had a value for LD_PRELOAD.
* This copies `dmtcpworker.cpp:restoreUserLDPRELOAD()` into `dmtcp_nocheckpoint.c` and uses it.
* In the second commit, this creates a test for dmtcp_nocheckpoint to verify that it is saving the user's LD_PRELOAD.


rohgarg:
LGTM.

PR #238, opened 2015-11-23
Improve coordinator logic

gc00:
There are three commits.  It will be easier to review by reading one at a time:
1.  Warn on restart if coord already running prev comp
2.  If --coord-port unspecified, use --new-coordinator 
   - Prev., if no coord mode specified, `--any-coordinator` was the default.
     This could cause `--join-coordinator` to be invoked.
     This caused unexpected behavior when the user was unaware of a curr. coord.
   - Changing -`-join` to `--join-coordinator`, but still recognizing `--join` for now.
3.  Use macro:  7779 -> `STRINGIFY(DEFAULT_PORT)`

### DIFF
--- a/src/constants.h
+++ b/src/constants.h
@@ -64,8 +64,12 @@
 #define INITIAL_VIRTUAL_PID         40000
 #define MAX_VIRTUAL_PID             400000000
 
-#define DEFAULT_HOST                "127.0.0.1"
-#define DEFAULT_PORT                7779
+// NEEDED FOR STRINGIFY(DEFAULT_PORT)
+#define QUOTE(arg) #arg
+#define STRINGIFY(arg) QUOTE(arg)
+
+#define DEFAULT_HOST "127.0.0.1"
+#define DEFAULT_PORT 7779
 #define UNINITIALIZED_PORT          -1 /* used with getCoordHostAndPort() */
 
 // Match up this definition with the one in plugin/ptrace/ptracewrappers.cpp
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -483,7 +483,7 @@ CoordinatorAPI::createNewConnToCoord(CoordinatorMode mode)
   if (mode & COORD_JOIN) {
     _coordinatorSocket = createNewSocketToCoordinator(mode);
     JASSERT(isValid()) (JASSERT_ERRNO)
-    .Text("Coordinator not found, but --join was specified. Exiting.");
+     .Text("Coordinator not found, but --join-coordinator specified. Exiting.");
   } else if (mode & COORD_NEW) {
     startNewCoordinator(mode);
     _coordinatorSocket = createNewSocketToCoordinator(mode);
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -435,12 +435,14 @@ CoordinatorAPI::startNewCoordinator(CoordinatorMode mode)
   jalib::JServerSocket coordinatorListenerSocket(jalib::JSockAddr::ANY,
                                                  port, 128);
   JASSERT(coordinatorListenerSocket.isValid())
-    (coordinatorListenerSocket.port()) (JASSERT_ERRNO)
-    .Text("Failed to create listen socket."
-          "If msg is \"Address already in use\", this may be an old"
-          "coordinator."
-          "Kill other coordinators and try again in a minute or so.");
-
+    (coordinatorListenerSocket.port()) (JASSERT_ERRNO) (host) (port)
+    .Text("Failed to create socket to coordinator port."
+          "\nIf msg is \"Address already in use\","
+             " this may be an old coordinator."
+          "\nEither try again a few seconds or a minute later,"
+          "\nOr kill other coordinators on this host and port:"
+          "\n    dmtcp_command ---coord-host XXX --coord-port XXX"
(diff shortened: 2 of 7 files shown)


gc00:
@karya0:  This has been ready for review for a while.  It should be easy to review.

src/constants.h
```diff
@@ -62,6 +62,10 @@
 #define INITIAL_VIRTUAL_PID  40000
 #define MAX_VIRTUAL_PID  400000000
 
+// NEEDED FOR STRINGIFY(DEFAULT_PORT)
+#define QUOTE(arg) #arg
+#define STRINGIFY(arg) QUOTE(arg)
```
karya0: Why not just delete `QUOTE` since we don't seem to be using it elsewhere in the code?

src/coordinatorapi.cpp
```diff
@@ -446,10 +446,14 @@ void CoordinatorAPI::startNewCoordinator(CoordinatorMode mode)
   jalib::JServerSocket coordinatorListenerSocket(jalib::JSockAddr::ANY,
                                                  port, 128);
   JASSERT(coordinatorListenerSocket.isValid())
-    (coordinatorListenerSocket.port()) (JASSERT_ERRNO)
-    .Text("Failed to create listen socket."
-          "\nIf msg is \"Address already in use\", this may be an old coordinator."
-          "\nKill other coordinators and try again in a minute or so.");
+    (coordinatorListenerSocket.port()) (JASSERT_ERRNO) (host) (port)
+    .Text("Failed to create socket to coordinator port."
```
karya0: Why not test ourselves for `EADDRINUSE` and print the message accordingly. This seems weird to tell the user to look at the errno string.

src/coordinatorapi.cpp
```diff
@@ -544,6 +548,17 @@ DmtcpMessage CoordinatorAPI::sendRecvHandshake(DmtcpMessage msg,
...
             " Reason: This process has a different computation group.");
   }
+  // Coordinator also prints this, but its stderr may go to /dev/null
+  if (msg.type == DMT_REJECT_NOT_RESTARTING) {
+    const char *coordinatorHost;
+    int coordinatorPort;
+    Util::getCoordHostAndPort(COORD_ANY, &coordinatorHost, &coordinatorPort);
+    JNOTE ("\n\n*** Computation not in RESTARTING or CHECKPOINTED state."
+        "\n***Reject incoming computation process requesting restart."
+        "\n***Could the coordinator be still running a previous computation?"
```
karya0: DMT_REJECT_NOT_RESTARTING does mean that the coordinator is running a different computation.  Let's also remove " or CHECKPOINTED" from the message since this is misleading.

src/dmtcp_command.cpp
```diff
@@ -38,7 +38,8 @@ static const char* theUsage =
   "  -h, --coord-host HOSTNAME (environment variable DMTCP_COORD_HOST)\n"
   "              Hostname where dmtcp_coordinator is run (default: localhost)\n"
   "  -p, --coord-port PORT_NUM (environment variable DMTCP_COORD_PORT)\n"
-  "              Port where dmtcp_coordinator is run (default: 7779)\n"
+  "              Port where dmtcp_coordinator is run (default: "
```
karya0: Should we change "run" to "running", here and above?

karya0:
I am not so sure about the "--any-coordinator" flag. We should make things explicit. Can we generate and finalize a policy before we make the changes?

1. Should we differentiate between "existing computation" and "existing coordinator", where the latter means the coordinator is running, but idle (without any computation).
2. For joining an "existing computation", should we enforce that one *must* specify the "--join-computation" flag? (I think we should remove the "--join-coordinator" flag since it's ambiguous.)
3. If no flag is specified, if the coordinator is "idle", we should connect to it and create a new computation. However, if the coordinator is *not* "idle", we should just fail to launch.
4. Whether or not a port has been specified, shouldn't affect that decision about joining/not-joining a coordinator/computation. This is not the standard behavior of most systems. Address/port are just to find coordinator or to launch a new one, not for decision making.

PS: I really think these changes should go in 3.0.0 branch, and not in the 2.x branch since they involve usability changes.

gc00:
@karya0, you've had this commit under review for a year now.  This commit does:

1.  replaces a hard-wired 7779 with DEFAULT_PORT
2. Adds more explanation to users.  For example: 
```
"\nEither try again a few seconds or a minute later,"
+          "\nOr kill other coordinators on this host and port:"
+          "\n    dmtcp_command ---coord-host XXX --coord-port XXX"
+          "\nOr specify --join-coordinator if joining existing computation.");
```

3. Replace --join with the more meaningful --join-coordinator.  And make --new-coordinator the default rather than have a default of joingin a cooridnator that the user knew nothing about.
4.  Warn the user if the computation is not in RESTARTING or CHECKPOINTING state.

In your review, you seemed to be saying that we should not add these improvements because you have some different improvements that you would prefer.  As a result, this PR has been orphaned.  I think this really gets in the way of forward progress.

I can propose several possibilities:

1.  You take responsibility for this merge of a year-old PR, and then you add on your favorite features.

2.  We abandon your favorite features, and I push this in.  And then you can start your own PR.

3.  You take full responsibillty for the user API, and we send you all of the enhancement requests that we'd like to see.

In general, we should avoid an ego thing about whose features are going to be pushed in.

PR #242, opened 2015-11-24
Fix for script test.

### LINKED ISSUE #241: No support for newer `poll` calls.
With newer libc, the poll function is defined as:
```
 __fortify_function int
poll (struct pollfd *__fds, nfds_t __nfds, int __timeout)
{
  if (__bos (__fds) != (__SIZE_TYPE__) -1)
    {
      if (! __builtin_constant_p (__nfds))
	return __poll_chk (__fds, __nfds, __timeout, __bos (__fds));
      else if (__bos (__fds) / sizeof (*__fds) < __nfds)
	return __poll_chk_warn (__fds, __nfds, __timeout, __bos (__fds));
    }

  return __poll_alias (__fds, __nfds, __timeout);
}
```

Thus it calls `__poll_chk` instead of `poll` and thus skips our `poll` wrapper altogether causing programs like script to fail if a checkpoint takes place in the middle of `__poll_chk` call.

karya0:
Added a wrapper for `__poll_chk` libc call.

Fixes #241.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -653,7 +653,7 @@ void DmtcpCoordinator::onDisconnect(CoordClient *client)
     }
   }
   client->sock().close();
-  JNOTE ( "client disconnected" ) ( client->identity() );
+  JNOTE ( "client disconnected" ) ( client->identity() ) (client->progname());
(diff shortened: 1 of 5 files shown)


rohgarg:
Looks good to me.

PR #243, opened 2015-11-25
Fixed a race with simultaneous calls to close.

karya0:
The locks were incorrectly acquired inside the if block. This fix ensures that
the locks are acquired for the whole duration of the function call.


gc00:
Looks good to me.  Nice catch!

Issue #245, opened 2015-11-25
dmtcp_launch --no-coordinator failing on certain applications


gc00:
A user reported (and I have confirmed) the following bug:
```
bash% bin/dmtcp_launch --no-coordinator --interval 5 tcsh
[40000] ERROR at coordinatorapi.cpp:562 in createNewConnectionBeforeFork; REASON='JASSERT(!noCoordinator()) failed'
tcsh (40000): Terminating...
```

Separately, while `--no-coordinator` works well with our favorite test (`test/dmtcp1`), it seems to be hanging on resume for `sleep`:
```
bash% bin/dmtcp_launch --no-coordinator --interval 5 sleep 10
```
The checkpoint image is produced correctly, and restarts correctly.  But the resume of the original process hangs after checkpointing.

gc00:
I built tcsh with -g and am seeing this:
```
(gdb) where
#0  __GI__exit (status=99) at ../sysdeps/unix/sysv/linux/_exit.c:29
#1  0x00007ffff710ef50 in jassert_internal::JAssert::~JAssert (this=0x7fffffffc8a0, __in_chrg=<optimized out>)
    at ../jalib/jassert.cpp:114
#2  0x00007ffff70da47f in dmtcp::CoordinatorAPI::createNewConnectionBeforeFork (this=0x7ffff7338320, progname=...)
    at coordinatorapi.cpp:562
#3  0x00007ffff70ded08 in fork () at execwrappers.cpp:181
#4  0x00000000004419d9 in remotehost () at tc.func.c:2002
#5  0x0000000000403c3a in main (argc=1, argv=0x7fffffffe4a8) at sh.c:684```
```

At frame 2, I'm seeing:
```C++
void CoordinatorAPI::createNewConnectionBeforeFork(string& progname)
{
JASSERT(!noCoordinator());
```

I suppose this mean that we don't support `--no-coordinator` if the target application uses fork().  This makes sense, since --no-coordinator means that the coordinator will be built into a _single_ process.  However, we could change this to assume that the child process is short-lived, and doesn't need to be checkpointed.  But clearly `tcsh` will want to run some long-lived applications in background.

So, the right solution here seems to be to improve the error message in the `JASSERT` to explain the issue to users, and to also add to the documentation of `--no-coordinator` in `dmtcp_launch -h` that this will fail on fork().

I'll create the appropriate pull request (and also investigate the issue for the `sleep` target application).

gc00:
[...]

PR #248, opened 2015-11-25
Improved error message/--help for --no-coordinator

gc00:
`dmtcp_launch --no-coordinator` should fail when target application calls fork(), since the coordinator is embedded in a single process.  But our error message did not make it clear why this happens.  This improves the JASSERT message and also the documentation for `dmtcp_launch -h`.

This issue is discussed as issue #245.


rohgarg:
Looks good to me.

Issue #240, opened 2015-11-23
misleading error message from attempt to set stack limit


loveshack:
There's a misleading error message if you run with a stack limit below 256MB.  Here's a correction (for 2.4.2, but the current source is unchanged):

```diff
--- dmtcp-dmtcp-172982f/src/mtcp/mtcp_check_vdso.ic~	2015-10-15 17:39:36.000000000 +0100
+++ dmtcp-dmtcp-172982f/src/mtcp/mtcp_check_vdso.ic	2015-11-12 16:23:35.024995359 +0000
@@ -342,7 +342,8 @@
                rlimit_failed |= mtcp_sys_getrlimit(RLIMIT_STACK, &rlim),
                rlimit_failed || rlim.rlim_max == RLIM_INFINITY )
            ) {
-          MTCP_PRINTF("Failed to reduce RLIMIT_STACK below RLIM_INFINITY\n");
+          MTCP_PRINTF("Failed to set RLIMIT_STACK to %d\n",
+		      ARCH_STACK_DEFAULT_SIZE);
           mtcp_sys_exit(1);
         }
         write_args(argv, "/proc/self/cmdline");

```

gc00:
Thanks, @loveshack.  I've now added pull request #249 to do your proposed change.

gc00:
This issue is now closed by PR#249.

PR #249, opened 2015-11-25
Improve stack limit error message in vdso.c

gc00:
This pull request was suggest by @loveshack.  It fixes a misleading error message in vdso.c when DMTCP tries to set the resource limit for the stack on restart.  This was raised originally as issue #240.


rohgarg:
Looks good to me.

Issue #247, opened 2015-11-25
User thread should wait until ckpt thread calls select on coord socket


gc00:
One of the principles of DMTCP is that either the checkpoint thread or the user threads can run, but normally not both at the same time.

This principle seems to be violated in the startup of the checkpoint thread during launch.  It leads to a real bug.  To illustrate, try:
```
bin/dmtcp_launch --no-coordinator tcsh
```
There is a race condition under which it will usually succeed, but sometimes fail with a SEGFAULT in the `malloc()` routine of tcsh.  (tcsh defines its own `malloc()`.)  I then observed tcsh to segfault here:
```
0x000000000044a4ab in malloc (nbytes=24) at tc.alloc.c:222
222	    nextf[bucket] = nextf[bucket]->ov_next;
```

The reason is that the malloc() of tcsh calls sbrk() and the malloc() of glibc (used by DMTCP) calls sbrk().  In both instances of malloc(), they follow the pattern:  `begin_buf = sbrk(0); ... end_buf = sbrk(size);`

There is an obvious race that heppens when both instances ask for `morecore()` at the same time.  In the case of tcsh, DMTCP calls select(), which calls dlsym() (since select() is a wrapper), and dlsym() calls malloc().

A final proof that we have a race condition is that in `src/threadlist.cpp`, one can modify the code to pause the user thread until the checkpoint thread reaches select:
```
  JASSERT(pthread_create(&checkpointhreadid, NULL, checkpointhread, NULL) == 0);
  sleep(2);
```

After that, the race condition goes away.

The solution is to set up a semaphore. initialized to 0.  After calling pthread_create to create the ckpt thread, we should call `sem_wait()`.  The ckpt thread can then call `select()` trivially (which invokes dlsym the first time, and only the first time), and then the ckpt thread can call `sem_post()`, followed by its call to `select()` to wait on the coordinator.  [We use a similar semaphore solution on restart so that the checkpoint thread and user thread don't interact.]

Are there comments on this solution?  One possible downside is that startup of the user thread will be slightly delayed, but not by much.

gc00:
Here is a little further explanation.  The checkpoint thread enters select() on the coordinator socket and then has this stack:
```
(gdb) where
#0  0x00007fbaf18fa3d3 in select () from /lib64/libc.so.6
#1  0x00007fbaf29f1228 in select (nfds=822, readfds=0x7fbaefa80d70, writefds=0x0, exceptfds=0x0, timeout=0x0)
[...]

PR #250, opened 2015-11-29
Fix race in: dmtcp_launch --no-coordinator tcsh

gc00:
More generally, sem_launch (formerly sem_start) was being posted too early,
while the checkpoint thread was still active

See github issue #247 for further background on this PR.


rohgarg:
Looks good to me.

PR #251, opened 2015-11-30
Build dlopen2 test correctly

gc00:
Fixes test/Makefile.in (It wasn't building libdlopen-lib3.so due to a typo).
It also polishes the dlopen1/dlopen2 tests.


gc00:
I've updated this.  It still only does some minor polishing.  It should be easy to review.

rohgarg:
Looks good to me.

PR #252, opened 2015-12-01
IB: Fix the memory leak when freeing device list.

jiajuncao:
The memory leak leads to a weird segfault during MPI_Finalize().

### DIFF
--- a/contrib/infiniband/ibv_internal.h
+++ b/contrib/infiniband/ibv_internal.h
@@ -12,6 +12,14 @@ struct dev_list_info {
   struct ibv_device ** real_dev_list;
 };
 
+/*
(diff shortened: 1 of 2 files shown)


gc00:
When the in-line changes are done, let's also open now the pull github issue that this code assumes that a single thread uses the IB API, and it will not work if the end user code has multiple threads using the IB API.  Then let's refer to this PR in the github issue, and refer to the github issue from this PR.  (If you do one, github should automatically do the other part.)

After that, it's fine to commit this.  The code is clean and elegant.  Nice!

jiajuncao:
@gc00 the thread safety issue isn't limited to this PR, and is assumed for the entire plugin. I already created a separate issue for that.

PR #253, opened 2015-12-02
dmtcp-api: Expose some existing functions

rohgarg:
Many functions declared in dmtcp.h are not really usable by
an end-user because they are not correctly declared.

This patch fixes the issue for functions related to setting
the checkpoint directory path and the coordinator scripts
directory path.


gc00:
It looks good to me.

PR #257, opened 2015-12-10
Don't require g++; AC_PROG_CXX already does the general case

gc00:
I remove a configuration section from configure.ac that forces g++ to be installed.  This is leftover from an early version of the file, and AC_PROG_CXX now checks for any available C++ compiler, instead of insisting on g++.

I didn't do autogen.sh and push in the corresponding configure file.  Since this is for version 2.5.0, someone else will inherit this configure.ac and do a new configure in time for version 2.5.0.  :-)


rohgarg:
Looks good to me.

PR #258, opened 2015-12-11
Cleaner display of glibc version in Makefile.in

gc00:
Tiny change in top level Makefile.in; Should be easy to review


rohgarg:
LGTM.

PR #259, opened 2015-12-11
Remove single host processing from restart script.

jiajuncao:
We need to use the RM plugin processing even it's a single-host computation.
Not 100% sure if multiple host processing can cover the single host case, but it looks good from my tests.

### DIFF
--- a/src/restartscript.cpp
+++ b/src/restartscript.cpp
@@ -197,34 +197,7 @@ static const char* cmdlineArgHandler =
   "fi\n\n"
 ;
 
-static const char* singleHostProcessing =
-  "ckpt_files=\"\"\n"
-  "if [ ! -z \"$DMTCP_RESTART_DIR\" ]; then\n"
-  "  for tmp in $given_ckpt_files; do\n"
-  "    ckpt_files=\"$DMTCP_RESTART_DIR/$(basename $tmp) $ckpt_files\"\n"
-  "  done\n"
-  "else\n"
-  "  ckpt_files=$given_ckpt_files\n"
-  "fi\n\n"
-
-  "coordinator_info=\"--coord-host $coord_host --coord-port $coord_port\"\n"
-
-  "tmpdir=\n"
-  "if [ ! -z \"$DMTCP_TMPDIR\" ]; then\n"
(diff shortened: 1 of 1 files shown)


src/restartscript.cpp
```diff
@@ -197,34 +197,7 @@ static const char* cmdlineArgHandler =
...
-  "  ckpt_dir=\"--ckptdir $DMTCP_CKPT_DIR\"\n"
-  "fi\n\n"
-
-  "exec $dmt_rstr_cmd $coordinator_info $ckpt_dir \\\n"
-  "  $maybejoin --interval \"$checkpoint_interval\" $tmpdir $noStrictChecking \\\n"
-  "  $ckpt_files\n"
-;
-
-static const char* multiHostProcessing =
+static const char* rstrCmdProcessing =
```
gc00: Let's change the name from `rstrCmdProcessing` to `rstrtCmdProcessing`.  (This distinguishes "restore" from "restart".)

gc00:
@karya0, you were probably the last one to touch this code.  In MPI with SLURM, when there are four processes on a single node, restart fails because the restart script does not use the additions provided by the rm plugin.  So, @jiajuncao is proposing to eliminate the singleHost case in the restart script, and send everything through the multi-host case.

Is there any reason to keep a singleHost case (e.g., for a single process)?  The pros are greater readability of the script by a human being.  The cons are more cases for us to maintain.  So, would tend to follow the proposed PR of @jiajuncao.  Thanks.

gc00:
@karya0, do you have a chance to look at this pull request?

@jiajuncao, Probably the planned UD pull request does not depend on this PR.  But if it does, feel free to push this in now, and we can ask @karya0 to do a second review at a later time.

gc00:
In the interest of not delaying the 2.4.4 point release, I propose that it's fine to delay this pull request until 2.4.5 (until after the upcoming 2.4.4 release).  In discussions with @karya0, it appears that we'll need a different strategy for the single-host case.  Merging it into the multi-host case will force calls to `ssh`, which @karya0 tells me could create problems.

gc00 (approved):
LGTM.  @karya0 thoguht there might be a problematic interaction with the ssh plugin, but I'm not seeing any.  And it's a small PR.  It's easy to back out later if we see a problem.

Issue #256, opened 2015-12-10
dmtcp_set_coord_ckpt_dir() sets up the dmtcp_restart_script.sh symbolic link incorrectly


rohgarg:
The `dmtcp_set_coord_ckpt_dir("dirname")` API is used to specify the location where the restart scripts are written.

A user reported (and I have verified) that the API sets up the `dmtcp_restart_script.sh` symbolic link incorrectly. If a relative path, say `./a/b/c`, is specified, the symbolic link we create (in the `./a/b/c` directory) ends up pointing to `./a/b/c/dmtcp_restart_script_XXX.sh` instead of `./dmtcp_restart_script_XXX.sh`.

PR #262, opened 2015-12-15
restartscript: Fix dmtcp_restart_script.sh symbolic link set up

rohgarg:
The dmtcp_set_coord_ckpt_dir("dirname") API is used to specify
the location where the restart scripts are written.

This patch fixes an issue with the API where specifying a
relative path would set up the dmtcp_restart_script.sh symbolic
link incorrectly. If a relative path, say ./a/b/c, was specified,
DMTCP would create the symbolic link (in the ./a/b/c directory)
that pointed to ./a/b/c/dmtcp_restart_script_XXX.sh instead
of ./dmtcp_restart_script_XXX.sh.

### DIFF
--- a/src/restartscript.cpp
+++ b/src/restartscript.cpp
@@ -524,7 +524,7 @@ void writeScript(const string& ckptDir,
     JTRACE("linking \"dmtcp_restart_script.sh\" filename to uniqueFilename")
       (filename) (dirname) (uniqueFilename);
     // FIXME:  Handle error case of symlink()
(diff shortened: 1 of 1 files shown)


rohgarg:
This fixes issue #256.

gc00:
I looked at `man basename`.  It does not use `malloc`.  This is good.

Separately, `basename` has two definitions:  POSIX and GNU.  They differ when the path ends in '/'.  I'm guessing that it's impossible for `uniqueFilename` to end in '/'.  If this is true then we're good to go.

LGTM (with the proviso in the previous paragraph).

PR #264, opened 2015-12-15
Fix the bug when dlopen() is used to open ib verbs

jiajuncao:
We link the plugin with the verbs library, and when the application
calls dlopen() to open libibverbs.so, we dlopen() the plugin in
the wrapper for dlopen(), so that all subsequent verbs calls will
fall in our verbs wrappers.

### DIFF
--- a/contrib/infiniband/infinibandwrappers.c
+++ b/contrib/infiniband/infinibandwrappers.c
@@ -15,7 +15,15 @@
 void *dlopen(const char *filename, int flag) {
   if (filename) {
     if (strstr(filename, "libibverbs.so")) {
-      return RTLD_DEFAULT;
+      void *handle = NEXT_FNC(dlopen)("libdmtcp_infiniband.so", flag);
+      if (handle == NULL) {
+        fprintf(stderr,
+                "\n*** Please either add $DMTCP_PATH$/lib/dmtcp "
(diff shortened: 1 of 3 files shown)


contrib/infiniband/infinibandwrappers.c
```diff
@@ -15,7 +15,12 @@
 void *dlopen(const char *filename, int flag) {
   if (filename) {
     if (strstr(filename, "libibverbs.so")) {
-      return RTLD_DEFAULT;
+      void *handle = NEXT_FNC(dlopen)("libdmtcp_infiniband.so", flag);
+      if (handle == NULL) {
+        fprintf(stderr,
+                "Please add $DMTCP_PATH$/lib/dmtcp to LD_LIBRARY_PATH");
```
gc00: Let's mention both user options here.  I'm also adding extra `\n` and '*'.  This message should stand out, since it's important for the user to see it.
```        fprintf(stderr,
               "\n*** Please either add $DMTCP_PATH$/lib/dmtcp to LD_LIBRARY_PATH,\n"
               "*** or else include an absolute pathname for libdmtcp_infiniband.so\n\n");
```

gc00:
The current branch apparently has a conflict with `master`.  If it's a trivial conflict, I don't need to see it.  After the in-line note above is addressed, it looks good to me.

PR #265, opened 2015-12-17
pathvirt plugin (Path Virtualization)

mossberg:
This plugin virtualizes path access for a program, which allows DMTCP to better accommodate to users who are running programs which use global paths under DMTCP.


plugin/pathvirt/test/pv.c
```diff
@@ -0,0 +1,308 @@
...
+
+    /* not found */
+    return -1;
+}
+
+/*
+ * clget - returns pointer to element in colonlist at index i
+ *         and NULL if not found
+ */
+char *clget(char *colonlist, unsigned int i)
```
gc00: Shouldn't `clget` be declared static?  (used only in this file)

plugin/pathvirt/test/pv.c
```diff
@@ -0,0 +1,308 @@
...
+    case DMTCP_EVENT_INIT:
+    {
+        /* On init, check if they've specified paths to virtualize via
+           DMTCP_PATH_PREFIX env */
+        char *old_env = getenv(ENV_DPP);
+        if (old_env) {
+
+            prefix_list_sz = strlen(old_env) + 1;
+            old_path_prefix_list = malloc(prefix_list_sz);
+            new_path_prefix_list = malloc(prefix_list_sz);
```
gc00: I'm a little worried about using `malloc`, since the end-user code could also be using `malloc`.  Some other possibilities are using jalib/jalloc.h, or else statically allocating 10 KB for each prefix_list, using `#define PREFIX_LIST_SIZE (10*1024)`, and then `JASSERT( prefix_list_sz >= PREFIX_LIST_SIZE).(prefix_list_sz).text("Too many environment variables.  Please increase PREFIX_LIST_SIZE")`

gc00:
I've added some inline comments.  I'll also suggest adding a github issue:  The test for the pathvirt plugin should eventually be moved into the DMTCP test directory for use with autotest.py.  After that, it looks good to me.

gc00:
One other comment.  Before you push this into master, could you combine some of the commits into a smaller number that are logically related (or even into one commit)?  It's easy using `git rebase -i HEAD~20` or something similar.  (Make sure you have a backup of your repo before you do that.)

plugin/pathvirt/pathvirt.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+    /* not found */
+    return -1;
+}
+
+/*
+ * dynamic_path_swap - given old path, return new path or NULL
+ *
+ * Returns NULL if no swap is to be done and the original path value should
+ * be used. Returns a malloc'd pointer to the new string if a swap should
```
gc00: @rohgarg and @mossberg, thanks for removing the malloc from the code.  Could you now remove the malloc reference from the comment above?  :-)

gc00:
We're almost ready.  The TODO list is:
1.  Add a github issue: The test for the pathvirt plugin in `test/slot5` should eventually be moved into the DMTCP test directory for use with autotest.py, and the comments referring to `slot5` should be updated to the new location.
2.  Use `git rebase -i HEAD~20` to squash the commits into one commit or else a very small number.

After that, it looks good to me.

mossberg:
@rohgarg @gc00: I'd like to do a little bit of cleanup

- `dynamic_path_swap` was named as such to hint to the caller that they needed to free the return value. now that is no longer the case, and since it is a publicly exported function, I think we should rename it. I think `pathvirt_get_physical_path` would be good
  - as Gene mentioned, need to update the comment for this function
- the style of this plugin is somewhat scattered with snake_case and camelCase mixed variables. It seems like camel case is preferred throughout the codebase, so I can change those

plugin/pathvirt/pathvirt.cpp
```diff
@@ -0,0 +1,240 @@
...
+        /* necessary since we don't know how many bytes dmtcp_get_restart_env
+           will write */
+        memset(new_path_prefix_list, 0, sizeof(new_path_prefix_list));
+
+        /* Try to get the value of ENV_DPP from new environment variables,
+         * passed in on restart */
+        int ret = dmtcp_get_restart_env(ENV_DPP, new_path_prefix_list,
+                                        sizeof(new_path_prefix_list) - 1);
+
+        JASSERT(ret == 0);
```
mossberg: @rohgarg @gc00 : this JASSERT will cause fail if they do not provide `DMTCP_PATH_PREFIX` on dmtcp restart and will enforce that users *must* provide `DMTCP_PATH_PREFIX` on *all* restarts. is this the desired behavior? or do we want to be more flexible and allow them to either provide it or not?

gc00: @mossberg, We should document what the behavior is clearly in a README or comments.

I would propose the following:  `DMTCP_PATH_PREFIX` is never required on restart.  If  `DMTCP_PATH_PREFIX` was not provided on the current restart, then there will _not_ be any path virtualization.

It seems to me that this is a feature rather than a bug.  :-)  A user can decide much later at restart time if they need path virtualization.  For example, maybe they need it if restarting on a new computer, and don't need it if restarting onthe original computer.

gc00:
@mossberg, I also like the two points that you bring up.  I agree that `pathvirt_get_physical_path` is a better name.  I also agree that uniform use of camel case is better.  In our C code, we tend to use the names with '_'.  But now that this code is C++, camel case reflects the dominant style better.

mossberg:
@rohgarg I think restarting multiple times is broken, have you tried using the `make restart2` test? it seems to checkpoint once, then not again?

plugin/pathvirt/pathvirt.cpp
mossberg: @gc00 I agree! Will correct this and add documentation

mossberg:
@gc00: when `dmtcp_get_restart_env()` fails with -2, that means the pathvirt buffer is not large enough for the user provided environment variable. Since this is a static buffer, it can't be reallocated, so I'd assume we should fail with an error message to the user.

What is the appropriate way to do this? `JASSERT` doesn't seem to accept a message to display, although I suppose I could do the old `assert(condition && "error message here")` trick.

mossberg:
Todo:

- [ ] handle case when environment variable is too large
- [x] submit issue for pathvirt test to be moved (done: https://github.com/dmtcp/dmtcp/issues/267)

gc00:
@mossberg wrote:
```
So I'd assume we should fail with an error message to the user.
What is the appropriate way to do this? JASSERT doesn't seem to accept a message to display,
```
In fact, `JASSERT` does accept a message to display.  The full format is:
```
JASSERT(CONDITION)(EXPR1)(EXPR2)...(EXPRN).Text("The message goes here,"
                    " and JASSERT will automatically add a newline at the end.");
```
You can find some examples if you grep through the source code.  Note also the trick
in my example, showing that when C and C++ see two adjacent string constants, they are concatenated
at compile time.

plugin/pathvirt/pathvirt.cpp
```diff
@@ -0,0 +1,250 @@
...
+               MAX_ENV_VAR_SIZE (to get past the first assert) we will then
+               trigger this assert */
+            JASSERT(ret != -2).Text("pathvirt: DMTCP_PATH_PREFIX exceeds "
+                    "maximum size (10kb). Use a shorter environment variable "
+                    "or increase MAX_ENV_VAR_SIZE and recompile.");
+
+            /* all other errors */
+            JASSERT(ret >= 0).Text("Fatal error retrieving DMTCP_PATH_PREFIX "
+                    "environment variable.");
+        }
```
mossberg: @gc00 : I'd appreciate a review of this code. Previously there was no handling for dmtcp_get_restart_env() failures, but this should be fairly robust.

plugin/pathvirt/pathvirt.cpp
```diff
@@ -0,0 +1,250 @@
...
+        /* necessary since we don't know how many bytes dmtcp_get_restart_env
+           will write */
+        memset(newPathPrefixList, 0, sizeof(newPathPrefixList));
+
+        /* Try to get the value of ENV_DPP from new environment variables,
+         * passed in on restart */
+        int ret = dmtcp_get_restart_env(ENV_DPP, newPathPrefixList,
+                                        sizeof(newPathPrefixList) - 1);
+
+        /* ret == -1 is fine everything else is not */
```
gc00: `is fine everything` -> `is fine; everything`

plugin/pathvirt/README
```diff
@@ -0,0 +1,46 @@
...
+  Any attempt to call a libc function on a previously registered path will
+  result in a new path being created with the old path prefix replaced by the
+  newly given one and the call forwarded on.
+
+  This flexibly allows path access to be redirected based on the conditions
+  the process is being restarted in.
+
+## constraints
+
+- due to a current limitation of dmtcp_get_restart_env(), the maximum length of
```
gc00: `due to` -> `Due to`

plugin/pathvirt/README
```diff
@@ -0,0 +1,46 @@
...
+  the colon lists that can be provided via the DMTCP_PATH_PREFIX variable
+  is 2982 bytes. This is because the maximum size of the name=value
+  string is 3000 bytes and "DMTCP_PATH_PREFIX=" occupies 18 of those. If
+  a environment variable that exceeds these limits is provided, it will be
```
gc00: `a environment` -> `an environment`

plugin/pathvirt/pathvirt.cpp
```diff
@@ -0,0 +1,250 @@
+/* NOTE:  if you just want to insert your own code at the time of checkpoint
+ *  and restart, there are two simpler additional mechanisms:
+ *  dmtcpaware, and the MTCP special hook functions:
+ *    mtcpHookPreCheckpoint, mtcpHookPostCheckpoint, mtcpHookRestart
+ */
```
gc00: @mossberg, dmtcpaware and mtcpHookPreCheckpoint are obsolete.  Here is some better wording for `*  and restart,  ...`:
```
 * and restart, the standard DMTCP plugins are available, using the events:
 *  DMTCP_EVENT_WRITE_CKPT and DMTCP_EVENT_RESUME and DMTCP_EVENT_RESTART.
 */

File: plugin/pathvirt/README [hunk 10]
```diff
@@ -0,0 +1,46 @@
...
+  truncated.
```
gc00: What is the reason for this 3000 byte limit on a `name=value` string?  Is it in a man page, or is it a limit that we set?  Usually, this happens due to the value of a configuration or environment variable.  What is the name of the ocnfiguration or environment variable, so that the end user can change it?

If the issue is related to this PR#261, please note that PR#261 is raising a limit to 12 KB in src/dmtcp
If the user is facing a limit due to `src/dmtcpplugin.cpp:MAXSIZE`, then please change the current limit reference in your comment from `3000` to `12*1024`, and explain that the end user can change this limit by modifying `src/dmtcpplugin.cpp:MAXSIZE`.

gc00:
@mossberg, I've now added some in-line notes.  I think one of them answers your review request, and I've now added some minor in-line notes that I overlooked on an earlier reading.

plugin/pathvirt/pathvirt.cpp
mossberg: @gc00 : ah, that comment was left from some other file that I copied as a starting point. I'll just delete the comment since it's not very relevant.

mossberg:
@gc00 : currently working on polishing the README. I also realized that currently, the `pathvirt_get_physical_path` hook is only applied to open and fopen libc hooks. i'd assume that this should be applied to all of the relevant functions in filewrappers.cpp that takes a `const char *path` as a parameter?

rohgarg:
> I also realized that currently, the pathvirt_get_physical_path hook is only applied to open and fopen libc hooks. i'd assume that this should be applied to all of the relevant functions in filewrappers.cpp that takes a const char *path as a parameter?

@mossberg: Thanks for catching that! I just verified that there are 11 more wrappers that we need to cover.

plugin/pathvirt/README
mossberg: At the time of this comment `src/dmtcpplugin.cpp:MAXSIZE` was set to 3000. Will update to the recently merged change to 12*1024 size

gc00:
@mossberg, Yes, thanks for catching that.  I forgot about all the other file wrappers.  It looks like we're almost done!  All we need now is the other file wrappers and the fixes for the minor things described in the inline notes.

mossberg:
@gc00 @rohgarg : I would imagine that there are *lots* of libc functions that operate on paths. Doesn't only hooking 11 libc functions mean incomplete support for programs? For example, I don't see chmod().

mossberg:
Todo:

- [x] handle case when environment variable is too large
- [x] submit issue for pathvirt test to be moved (done: https://github.com/dmtcp/dmtcp/issues/267)
- [x] update and polish documentation
- [ ] implement hooks for 11 other filewrapper.cpp functions
- [ ] squash all commits down into one commit for merging

mossberg:
@rohgarg would you be able to take care of the filewrapper hooks since you're most familiar?

src/plugin/ipc/file/filewrappers.cpp
```diff
@@ -371,7 +371,15 @@ static int _open_open64_work(int(*fn) (const char *path, int flags, ...),
...
+  int fd = -1;
+
+  /* hook_path was NULL, not swapping */
+  if (!doSwap)
+    fd = (*fn)(newpath, flags, mode);
+  else
+    fd = (*fn)(hook_path.c_str(), flags, mode);
 
   if (fd >= 0 && dmtcp_is_running_state()) {
     FileConnList::instance().processFileConnection(fd, newpath, flags, mode);
```
mossberg: @gc00 @rohgarg is always using `newpath` here a bug? if we have swapped, should we be using `hook_path.c_str()` in this `processFileConnection` call instead?

gc00: @rohgarg, feel free to comment further.

@mossberg, I believe the use of `newpath` is correct here.  The model is that `newpath` is the path seen prior to checkpoint.  This is considered the "virtual path", since it may be changed to point to a different physical path on restart.

Since the physical path can change many times (once for each restart), the virtual path becomes the canonical name of a file.  Since newpath is a virtual path, this is the one to use for the internal state of DMTCP.

(Stated differently, the internal state of DMTCP should always refer to the virtual path (the path prior to checkpoint).  Any system calls should convert between the physical path known to the filesystem and the virtual path used for the internal state of DMTCP.)

Also, please seem my next inline note, where I have a related suggestion to simplify and improve the readability of the code.

gc00: @mossberg,
    Wouldn't this code be conceptually simpler if we re-named `pathvirt_get_physical_path(newpath, dmtcp::string hook_path)` to be:
```

char * virtual_to_physical_path(char *newpath, dmtcp::string phys_path);
```

where the usage would be simplified to:
```
dmtcp::string phys_path_string = "");
char *phys_path = virtual_to_physical_path(newpath, phys_path_string);
...
fd = (*fn)(phys_path, flags, mode);
```
In this case, we would define `virtual_to_physical_path(...)` to return either `newpath` (if no swap) or `phys_path_string.c_string()` (if there is a swap).

-----------------------------

If you agree with the above comment, could you re-define/re-name the functions and variables as indicated?  (Sorry that I didn't catch this earlier.)

mossberg: done!

plugin/pathvirt/pathvirt.cpp
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/*
+ * virtual_to_physical_path - translate virtual to physical path
+ *
+ * Returns a const char* for the corresponding physical path to the given
+ * virtual path. If no path translation occurred, the given virtual path
+ * will simply be returned. If path translation *did* occur, the translated
+ * physical path will also be available as a dmtcp::string from the second
+ * parameter.
```
gc00: Just a short suggested additional comment:
```
 *     Conceptually, an original path prior to the first checkpoint is considered a "virtual path".
 * After a restart, it will be substituted using the latest list of registered paths.  Hence,
 * a newly registered path to be substituted is a "physical path".  Internally, DMTCP works
 * with the original "virtual path" as the canonical name.  But in any system calls, it must
 * translate the virtual path to the latest "physical path", which will correspond to the
 * current post-restart filesystem.
 */
```

gc00:
@mossberg, Nice work.  I have a suggested addition to the comment in the line note above.  Feel free to modify whatever is not clear.  Two pairs of eyes are always better.

With the added explanation of virtual versus physical, do you feel confident to write a first draft of the remaining wrapper functions?  (@rohgarg is still travelling, but he should be available later.)

I see that we're now getting a failure in the tests.  Since it happens prior to checkpoint, it may be easy for you to debug.  Maybe try:  `gdb --args dmtcp_launch test/python` as an example.  Add:  `break _exit`, `break exit`, and hope that GDB stops before it exits.  Then you can look at the stack.  Alternatively, set the ulimit to allow core dumps, and try:  `gdb test/python core`.

If you still have trouble finding the bug, tell me, and I'll try diagnosing it.

mossberg:
i'll be using this comment as a work log for myself. i've found it's usually helpful for me to do this so i have context when i resume work, and helpful for others as a little bit of transparency into what i'm doing :)

work log:

1/5 7:54pm

- i have two things to work on: failing tests, and implementing file wrappers. the failing tests make me uncomfortable, so i think i'll try to tackle those first, since i'm also curious at how the automated testing framework works
- failing tests are: syscall-tester, file1, stat, poll, frisbee, shared-memory1

```
== Tests ==
dmtcp1         ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
dmtcp2         ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
dmtcp3         ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
dmtcp4         ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
alarm          ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
sched_test     ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
dmtcp5         ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
syscall-tester ckpt:FAILED
               root-pids: [331] msg: checkpoint error, 1 expected, 0 found, running=0
file1          ckpt:FAILED
               root-pids: [340] msg: checkpoint error, 1 expected, 0 found, running=0
stat           ckpt:FAILED
               root-pids: [349] msg: checkpoint error, 1 expected, 0 found, running=0
plugin-sleep2  ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
plugin-example-dbckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
gettimeofday   ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
sigchild       ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
shared-fd1     ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
shared-fd2     ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
stale-fd       ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
poll           ckpt:FAILED
               root-pids: [558] msg: checkpoint error, 1 expected, 0 found, running=0
forkexec       ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
pthread_atfork1ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
pthread_atfork2ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
waitpid        ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
client-server  ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
frisbee        ckpt:FAILED
               root-pids: [710, 711, 712] msg: checkpoint error, 3 expected, 0 found, running=0
shared-memory1 ckpt:FAILED
               root-pids: [730] msg: checkpoint error, 2 expected, 0 found, running=0
shared-memory2 ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
sysv-shm1      ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
sysv-shm2      ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
sysv-sem       ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
sysv-msg       ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED
```

- looking at tests/autotest.py to study how the tests work
- looking at testCheckpoint(). looks like CHECK() contains the passing conditions. in test checkpoint, we check 1 thing pretty much.

rohgarg:
@mossberg, @gc00: I have now rebased on top of dmtcp/master and pushed in a fix for the failing tests.
Next, I'll add the path virtualization for the rest of the wrappers in `filewrappers.cpp`.

gc00:
@mossberg, @rohgarg: Thanks for fixing the failing tests, @rohgarg.  So, our remaining checklist is:

 1. add path virtualization to remaining wrappers (@rohgarg will do it)
 2. Merge the additional comment into the code (or some variation of the comment): see comment in my line note above
 3. Squash the commits into a single commit

mossberg:
@gc00 I believe the comment should be merged in (the inline not did not disappear because the line it's on was unchanged)

rohgarg:
@gc00, @mossberg: I have now added path virtualization to the rest of the wrappers in `filewrappers.cpp`. I haven't yet squashed everything down to one commit; I wanted to keep the history until the final ship-it.

gc00:
@rohgarg, Have you tried a few sanity tests of this sort?
```
export TMPDIR=/tmp/path1
rm -rf $TMPDIR
mkdir -f $TMPDIR
export DMTCP_PATH_PREFIX=$TMPDIR
make tidy
test/dmtcp_launch -i5 file1
# Kill after ckpt
export TMPDIR=/tmp/path2
rm -rf $TMPDIR
mv /tmp/path1 /tmp/path2
export DMTCP_PATH_PREFIX=$TMPDIR
test/dmtcp_restart ckpt_*.dmtcp
```

This example used the test `file1`.  But the tests `file2`, `inotify`, and `shared-mmeory2` also seem to use `TMPDIR`.

mossberg:
For the record, my manual tests only test `open` so it would probably be best to add some kind of tests for the other libc hooks.

mossberg:
@gc00 also, a lingering question of mine is about libc coverage. for example, we don't implement `chmod()` hooking. it seems like a lot of work, but in order for this to be a totally usable feature, wouldn't hooks need to exist for all (or nearly all) libc functions that take paths?

gc00:
@mossberg, your comments are all true.  Unfortunately, with our limited resources, it's not practical for us to design a full set of tests for path virtualization.  Your `chmod` case is a good example of something that hasn't been tested.  (Luckily, it's rare that an application calls `chmod` after restart.)

We've been forced to use a strategy of supporting the most common cases first.  There are users who typically test the `master` branch before we release.  This code will go into the `master` branch, and be planned for the 2.5.0 release.  Hence, we will have to rely on our implicit beta testers for more of the sanity tests.

In the meantime, if @rohgarg is occupied right now, would you have time to do some small sanity tests described above, for:  `file1`, `file2`, `inotify`, and `shared-memory2`?  This is not essential.  If none of us have time, we should push this in, and ask some of our users for testing of this feature on the `master` branch.

mossberg:
@gc00: Yes, I'll have time this wednesday, evening-ish. I'll comment what I find, and if it's good, perhaps we can plan to merge to master on thursday.

mossberg:
@gc00 : It seems like sanity testing those programs does not work because of errors like

```
vagrant research test (pathvirt-v3) ❯ ./dmtcp_restart_script.sh
[40000] ERROR at fileconnection.cpp:859 in openFile; REASON='JASSERT(jalib::Filesystem::FileExists(_path)) failed'
     _path = /tmp/path1/dmtcp_dir1_bQF0Vq/dmtcp_dir2_n9Zp9n/dmtcp_file_0gLSml
Message: File not present
file2 (40000): Terminating...
```

I think that error is because pathvirt is not active during the Fileconnection refill stage which reopens all the files. It seems like perhaps Fileconnection::openFile should be changed to check if pathvirt is active and call `virtual_to_physical_path` on `_path`?

mossberg:
@gc00 : please take a look at my latest commit, e7f63d2, it inserts pathvirt into the FileConnection refill process and allows the file1 test to work.

gc00:
@mossberg:  Thanks for doing the sanity test, and catching an area where we needed to extend the functionality!

@rohgarg, @mossberg:  Interesting.  Now the `script` and `java1` tests are failing during ordinary `make check`.  I'm seeing this in the travis tester of github.  In this case, there is no environment variable, DMTCP_PATH_PREFIX.  So, in principle, the changes of this pull request should not have prevented `make check` from working.

I suspect that this is an easy bug to fix.  But I won't have a chance to look at it seriously until late tomorrow (late Thursday).  Do one of you have time to see why `make check` is failing?

mossberg:
I think it's my most recent commit (e7f63d2), but i'm not sure how :/. `make check` passes if I checkout to one commit previous.

mossberg:
@gc00 @rohgarg : I think I know the issue, e7f63d2 is incomplete and did not change other relevant parts of the refill function, for example where stat() is called later on. I'll take another look at this today.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -856,13 +860,17 @@ bool FileConnection::checkDup(int fd)
 
 int FileConnection::openFile()
 {
-  JASSERT(jalib::Filesystem::FileExists(_path)) (_path)
+  dmtcp::string phys_path_string = "";
+  const char *phys_path =  virtual_to_physical_path ?
+                           virtual_to_physical_path(_path.c_str(), phys_path_string):
+                           _path.c_str();
+  JASSERT(jalib::Filesystem::FileExists(phys_path_string)) (_path)
```
mossberg: @gc00 @rohgarg : The bug is here I think. It's because the pathvirt plugin isn't loaded, so `phys_path_string` never gets reassigned from `""` and on line 685, it then asserts that the `""` exists as a file.

mossberg:
@gc00 @rohgarg: I'm starting to not really like the `virtual_to_physical_path` interface. I think it's a little messy to have both a return value and an output parameter that are essentially the same thing.

I'd like to propose a new interface:

```
dmtcp::string virtual_to_physical_path(const char *virt_path);
```

Also, instead of using a ternary operator all the time when using it, I'd like to also propose a macro:

```
#define VIRT_TO_PHYS_PATH(virt) (virtual_to_physical_path ? virtual_to_physical_path(virt) : dmtcp::string(virt))
```

with the following semantics:

- a dmtcp::string is always returned. if the dmtcp::string is not needed (like for filewrappers), then `.c_str()` can just be appended to get similar behavior/convenience for what exists now
- if pathvirt is not loaded, the macro user gets back a dmtcp::string of the path they gave in
  - previously, there was inconsistency between the return value and the output parameter. having one return value eliminates that
- the only possible return values are a dmtcp::string of the passed in string, or a dmtcp::string of the translated physical path. I feel like this is a lot cleaner for the receiving end, rather than having two output parameters

src/plugin/ipc/file/fileconnection.cpp
gc00: @mossberg:  Good catch!  I assume that the right fix is to change `phys_path_string` to `phys_path` in the `JASSERT` on line 857.  If you agree, could you change it?

And I'll comment on alternative styles for `virtual_to_physical_path()` in the general comment below.

gc00:
Concerning

```
dmtcp::string virtual_to_physical_path(const char *virt_path);
```
the issue is a possible memory leak.  This is one of the weaknesses of C++.  It's too easy to create a new object and forget to garbage collect it later.  To clarify, note that the new interface should include a new name:

Change the name from `virtual_to_physical_path()` to `virtual_to_physical_path_string()`.  (The new name tells the reader that we are creating a new string, and returning it from the function.  Hence, it is persistent and not local to a function.)

After this, it becomes necessary to call a destructor on the string object when we are done using it.  So, the choice in styles is to either add a destructor later in this code (and hope that the reader realizes that this code and the non-contiguous destructor must always be maintained as a single unit), or else go with the old style in which the string is a local object that will automatically be freed at the end of the current function.

Concerning
```
#define VIRT_TO_PHYS_PATH(virt)  (virtual_to_physical_path ? \
                                  virtual_to_physical_path(virt) : \
                                  dmtcp::string(virt))
```
I assume that C++ will require us to change `dmtcp::string(virt)` to `new dmtcp::string(virt)`.  and that we will want to change the macro name to `VIRT_TO_PHYS_PATH_STRING`.  As before, we then have to worry about memory leaks.  So, we need separate code to destroy the string at the end of this function -- which can be error-prone if a future developer forgets to maintain this constructor and the destructor at the end of the function calling this as a single logical unit of code.

src/plugin/ipc/file/filewrappers.cpp
```diff
@@ -371,7 +371,13 @@ static int _open_open64_work(int(*fn) (const char *path, int flags, ...),
     newpath = currPtsDevName;
   }
 
-  int fd = (*fn) (newpath, flags, mode);
+  dmtcp::string phys_path_string = "";
+  const char *phys_path =  virtual_to_physical_path ?
+                           virtual_to_physical_path(newpath, phys_path_string):
+                           newpath;
+
```
gc00: @mossberg: Concerning your interface proposal in the "conversation" tab, I do like your idea about using a macro.  These three lines get repeated often, and should be replaced by a macro.  How about the following?
```
// Note that virtual_to_physical_path is a weak symbol. 
// (It is a NULL ptr if not linked to an object defining this.)
#define VIRTUAL_TO_PHYSICAL_PATH(virt_path, phys_path_string) \
                  (virtual_to_physical_path ? \
                   virtual_to_physical_path(newpath, phys_path_string) : \
                   newpath)
```
Then, the line above becomes:
```
const char *phys_path = VIRTUAL_TO_PHYSICAL_PATH(newpath,
                                                 phys_path_string)
```
and we can invoke the macro in a similar way on the lines below.

(Note also that I try to keep the code within 80 columns.  And I use a long macro name for the macro (no abbreviations), since the macro name is not defined locally, and otherwise the user has to search for the exact name if one of several possible abbreviations is chosen.)

And note that I added to your line note where you diagnosed the bug concerning `JASSERT` and `phys_path_string`.  I'm proposing a bug fix there (and maybe it's the same one that you already had in mind).

plugin/pathvirt/pathvirt.cpp
gc00: @mossberg:  I lost track, but I think I still owe you a code review here.  My main comment is that we should be using the macro names here instead of the constants, -1, -2, -3.  The file `dmtcpplugin.cpp` already has macros for them:  `SUCCESS`, `NOTFOUND`, `TOOLONG`, etc.  Suppose we move those `#define`'s into `include/dmtcp.h` and give them a longer name to implicitly encode a kind of namespace:
```
// Sub-optimal.  See comment below for a macro name without DMTCP_
#define DMTCP_RESTART_ENV_SUCCESS 0
#define DMTCP_RESTART_ENV_NOTFOUND -1
#define DMTCP_RESTART_ENV_TOOLONG -2
#define DMTCP_RESTART_ENV_BUF_TOO_SMALL -3
#define DMTCP_RESTART_ENV_INTERNAL_ERROR -4
#define DMTCP_RESTART_ENV_NULL_PTR -5
#define DMTCP_RESTART_ENV_MAXSIZE 12288
```
If we put these macros into `include/dmtcp.h`, then we are forced to begin the macro names with `DMTCP_`.  This is because `include/dmtcp.h` can be included in the code of the end user's application.  So, we don't want to accidentally redefine a macro name that is already being used by the end user.

So, better still would be to find an include file shared between `dmtcpplugin.cpp` and this `pathvirt.cpp`.  If we do that then the "namespace" of `RESTART_ENV_` will suffice.  In this case, a logical choice would be to create a file, `dmtcpplugin.h`.  If you choose that mechanism (which is better), then you would add to `src/Makefile.am`.  Search on `syscallwrappers.h` and add this `.h` file in the section.  (We can get away here without doing a new `automake` to create a new `src/Makefile.in`.  That will happen before a new  a release anyway.)

src/plugin/ipc/file/fileconnection.cpp
mossberg: @gc00 : (do you mean on line 867?) One problem is that `FileExists` takes a cpp string, not a `char *`. I think a solution could be to initialize `phys_path_string` to `_path` instead of `""`. do you agree?

mossberg:
@gc00 : I have a test implementation of the interface I thought of available at https://github.com/mossberg/dmtcp/commit/37f752b0701748930281d4e95349f3d38a527253, it seems the macro without the `new` in `new dmtcp::string(virt)` seems to build, at least.

Maybe I misunderstand, but the `dmtcp::string` would be returned by value from `virtual_to_physical_path`, which performs no dynamic allocation (see my test impl above). Is the destruction still an issue?

mossberg:
@gc00 As I suspected, this nonnull parameter thing is happening all over the place.. open, open64, creat, access, opendir, stat, readlink

mossberg:
Another potential complication: the stat wrapper does something interesting, it checks for a bad address by making an initial test syscall. This should probably be done with all other calls who are defined to EFAULT on a bad address because currently the only bad address that would be checked is NULL, but yes technically we should handle any bad address.

gc00:
Does it work to put an 'is_null(address)' into your
VIRTUAL_TO_PHYS_XXX() macro?  If so, add 'is_null(address)' to util.cpp
src/util_misc.cpp.  (Include a comment explaining why we need it:
 some functions, such as access(), declare their arguments to be
  non_null, and still want to return an error if the argument is NULL.
  The compiler optimizes out any reference to NULL, and so we call
  this separate function.

If we later move this to a standalone plugin, you can copy the
'is_null(address)' function to the standalone plugin.

Best,
- Gene

mossberg:
current todo
- broken tests: shared-memory1, pthread1, pthread5 on my machine
- challenges with filewrappers
  - current inconsistencies. for example, readlink(). the pathvirt hook is currently immediately before the `_real_readlink` however, this produces an inconsistency if the user is trying to virtualize `/proc`. if they are trying to divert /proc elsewhere and post restart, the application opens `/proc/self/exe` for example, we still still access `/proc/self/exe` instead of where they tried to divert to because that logic check comes first. i suspect this is elsewhere too. to be correct the pathvirt hook really need to be the first thing that happens. it seems an elegant approach is to have a set of separate filewrappers which are called before the ipc/file plugin filewrappers. 
  - currently the filewrappers don't do a good job of checking for bad addresses. the only one i've converted to handle this properly is xstat. the only way to know if an address is bad is to do a test syscall with the original argument, check the and errno. if pathvirt is in the "front line" for all of the filewrappers, it will have to do this. it might be helpful to abstract this away with *another* set of filewrappers that go first, just dedicated to this purpose

gc00:
@mossberg:  Let's assume that the user is not trying to virtualize `/proc`.  Ultimately, DMTCP works by providing coverage for the sweet spot (the common cases), and then it exte4nds support to less common cases over time.  It's good to include a FIXME comment in the code about not currently supporting `/proc`.  If we thought many users would run into this, then we would also include a `JWARNING`, but I doubt that many users will require this.

Similarly, bad addresses will eventually need to be covered, but that can come in a later PR.

You wrote about inserting pathvirt into the FileConnection code to virtualize lingering file descriptors that are being reopened.  What is the scenario here?  Is it that we closed an fd prior to checkpoint, and will re-open it upon restart?  In any case, after discussions with @rohgarg today, we decided to insert the pathvirt plugin in a layer that is even below that of the ipc or file plugin.  In this way, we're hoping that many path virtualization issues will be handled directly in the path virt plugin, after the ipc/file plugin and ust before making the actual call to `libc`.

As for the failing tests on your machine:  shared-memory1, pthread1, pthread5
we're not seeing that here.  Could you tell us which distro you're using and the version number?
`lsb_release -a` will usually give you the full information.

It seems like this work is almost ready to go.  If we can diagnose the failing tests that you're seeing, that will help a lot.

mossberg:
@gc00 Ok, that all makes sense. I guess the priority really should be to get this out ASAP so we can get user feedback.

The scenario with lingering file descriptors is just if the process has fds that are open when the checkpoint occurs. In the early stages of restarting the process, those fds are "refilled" by reopening them. I had previously integrated that refill stage with DMTCP, but I reverted because it didn't seem right. I think that adding a pathvirt layer immediately before libc is an *excellent* idea.

My machine:

```
vagrant research ~ ❯ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.3 LTS
Release:	14.04
Codename:	trusty
```

If you're not seeing them, I guess it's possible that I have something configured strangely. Or possibly

rohgarg:
Merged via PR #318.

PR #268, opened 2015-12-21
Support for InfiniBand UD connection

jiajuncao:


contrib/infiniband/ibvidentifier.h
```diff
@@ -1,26 +1,22 @@
...
-#include <stdlib.h>
+#include <stdint.h>
 
-struct ibv_qp_id {
+typedef struct ibv_qp_id {
   uint32_t qpn;
   uint16_t lid;
   uint32_t psn;
-};
+} ibv_qp_id;
```
gc00: Let's change the new `typedef` type from `ibv_qp_id` to `ibv_qp_id_t` everywhere.  This is closer to standard conventions.  (`_t` for "type")

contrib/infiniband/ibvidentifier.h
```diff
@@ -1,26 +1,22 @@
...
 
-struct ibv_qp_pd_id {
+typedef struct {
   uint32_t qpn;
   uint32_t lid;
-};
+} ibv_qp_pd_id, ibv_ud_qp_id;
```
gc00: Similarly, let's change new typedef types from `ibv_qp_pd_id` and `ibv_ud_qp_id` to `ibv_qp_pd_id_t`and `ibv_ud_qp_id _t`everywhere.

contrib/infiniband/keys.ic
```diff
@@ -58,20 +62,59 @@ uint32_t update_rkey_restart(uint32_t rkey, int pd_id)
...
+  orig_id.lid = internal_ah->attr.dlid;
+
+  dmtcp_send_query_to_coordinator("qp_info",
+                                  &orig_id, sizeof(orig_id),
+                                  &curr_id, &size);
+
+  if (size == sizeof(ibv_ud_qp_id)) {
+    // The qp is created before checkpoint, qp_num changes
+    wr->wr.ud.remote_qpn = curr_id.qpn;
+  }
```
gc00: The "if" statement tests `if (size == sizeof(ibv_ud_qp_id))`.  This code eems unusual and difficult to maintain.  It seems to be saying that the `curr_id` of size `size` that was returned by a query could be of more than one type.  So, you're testing the _size_ of the data structure that was returned, to decide if it is the right type.

Normally, I would expect the first field of a struct to be a constant of a certain type.  So, I would do something like `if (curr_id.type == UD_QP_ID) { ... }`.  Then, anyone maintaining this code can debug it more easily by looking at the field `curr_id.type`.

It's also possible that I didn't understand this code correctly.  But in that case, the right thing to do would be to add an extra comment before the "if" statement, to help people like me to understand this code.

gc00:
@jiajuncao, I have some inline notes, but nothing major.  If the inline notes are clear, then it's fine to go ahead and push this in.  Otherwise, write your own comment, and we can continue to discuss this.

contrib/infiniband/keys.ic
jiajuncao: The "if" statement checks whether there is a corresponding key-value pair in the coordinator database. `curr_id` and `size` are the output of `dmtcp_send_query_to_coordinator()`, and the if statement is true when there is a value corresponding to `orig_id` in the coordinator database.

gc00: How about adding the following comment before the "if" statement:
```
// If the key in the query does not exist, size == 0.
```
This helps readers who don't know all the details of `dmtcp_send_query_to_coordinator()`.

PR #272, opened 2015-12-27
Fix race while reading /proc/self/maps for ckpt

gc00:
If the JAlloc allocator expands its memory, it will call mmap().  If this happens while reading `/proc/self/maps`, it will create a bug on restart.  The allocator will have saved a pointer to the freshly mmap'ed memory, but since it was mmap'ed after reading `/proc/self/maps` and before we copied memory into the checkpoint image, we will have a dangling pointer into memory that has not been saved.

This bug appears when doing either of:
```
make -j AUTOTEST=--stress check-pthread5
make -j AUTOTEST=--stress check-shared-memory1
```

### DIFF
--- a/include/procselfmaps.h
+++ b/include/procselfmaps.h
@@ -32,6 +32,7 @@ class ProcSelfMaps
     size_t numAreas;
     size_t numBytes;
     int fd;
+    int numAllocExpands;
 };
 
 }
--- a/jalib/jalloc.cpp
+++ b/jalib/jalloc.cpp
@@ -27,6 +27,10 @@
 #include "jalib.h"
 #include "jalloc.h"
 
+// Make highest chunk size large; avoid a raw_alloc calling mmap()
+//   during /proc/self/maps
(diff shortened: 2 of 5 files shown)


gc00:
@karya0, @rohgarg:  Would one of you have time to review this pull request?  I'd like to include it in the upcoming 2.4.4 release.

It should be easy to review.  The key to reviewing it is to first read the changes to `jalloc.cpp`.  That code defines two new jalloc functions:  `preExpand()` and `numExpands()`.  Then I use `preExpand()` in `src/procselfmaps.cpp` to pre-allocate in the jalloc memory arena (calling mmap() in advance).  This is done in the constructor `ProcSelfMaps::ProcSelfMaps()`, so that any calls to `readmapsline()` will not be disturbed by the jalloc allocator calling `mmap()` in the middle of it.

The rest is just comments to polish and document the logic.

rohgarg:
LGTM.

PR #274, opened 2016-01-03
Fixed multi-architecture support.

karya0:
The steps listed in doc/multi-arch.txt should work now.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -69,6 +69,10 @@ endif
 	$(MKDIR_P) $(targetdir)/lib/dmtcp
 
 build: mkdirs dmtcp plugin contrib
+ifeq ($(M32),1)
+	cd lib/dmtcp && $(LN_S) 32/lib/dmtcp/* ./
+	mkdir -p bin && cd bin && $(LN_S) ../lib/dmtcp/32/bin/* ./
+endif
 
 all: default
 
@@ -109,10 +113,6 @@ display-config:
 dmtcp:
 	rm -f src/config.h
 	cd src && $(MAKE)
-ifeq ($(M32),1)
-	cd lib/dmtcp && ln -s 32/lib/dmtcp/* ./
-	mkdir -p bin && cd bin && ln -s ../lib/dmtcp/32/bin/* ./
-endif
 
 plugin: dmtcp
 	cd plugin && $(MAKE)
--- a/Makefile.in
+++ b/Makefile.in
@@ -59,12 +59,7 @@ MANPAGES=dmtcp.1.gz dmtcp_coordinator.1.gz dmtcp_command.1.gz dmtcp_launch.1.gz 
 
 default: display-build-env build
 
-# If M32, 'rm -rf lib bin'.  Can't do a 32-bit build on top of a 64-bit build,
-#   neither './configure --enable-m32' nor 'make clean' will remove these dirs.
 mkdirs:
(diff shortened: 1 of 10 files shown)


gc00:
@karya0, I've now spotted why this pull request doesn't work correctly for the dlopen1 test.

First, I'd like to keep this as a conservative pull request to go into DMTCP 2.4.4 (bug fixes only).  The more aggressive commits can go into a separate pull request, for DMTCP 2.5.0.  So, here are the changes that I'm proposing:

1. It seems that commit 735077a0 changes test/Makefile.in.  This causes the dlopen1 test not to be built correctly.  The commit 735077a0 is not needed for the bug fix.  Can we move that to a separate PR (and later fix the logic for the dlopen1 test)?

2. Separately, could we remove the commit abf356b8d6bc0e6.  We had an earlier discussion about the `clean-64` Makefile target.  It allows the knowledgeable user to build a multi-arch within their local directory instead of a separate build directory.  This was why clean-64 was also included in the instructions in the doc subdirectory.  We had decided that it does no harm if a user wants to use the clean-64 target to build multi-arch within their source directory (and not need a separate build directory).  In any case, removing the clean-64 target (whether a feature or misfeature) should not be done as part of a bug fix release.

3. Similarly, commit ae5473fcb1c9138b7 tries to delete the instructions in doc/multi-arch.txt about how to use clean-64 to build multi-arch within the source directory.  Could we please restore the instructions for multi-arch/clean-64 in doc/multi-arch.txt?  Any changes in the doc directory can also be deferred to a 2.5.0 release.

4. After this, if you can do a `git pull --rebase origin master`, you'll pick up a new bug fix that I pushed in recently as part of PR #275.  I was able to discover and fix my own earlier bug due to the work that you did on this PR, and I thank you for it.

(My earlier bug, by itself, was already going to break multi-arch for dmtcp-2.4.0.  So, I broke multi-arch even before our automake code broke it.  :-) )

Thanks again for your careful fixes.

karya0:
@gc00: I've now updated the PR.

(1): I found the bug, but you are correct about delaying it to 2.5.0, so I've created #276 now. 

(2): I have re-added 'clean-64' by making it call 'clean'. I tested locally to build both 32- and 64-bit builds in the root directory (w/o any build directory) and it works as expected. Apparently, the clean-64 was needed because of an earlier bug/feature that removed the bin/lib directories from the build tree. The commit handles both.

(3): I think it's okay to update the build instructions specifically if we are making it simpler. I don't have a strong opinion on this one so I can revert it if you want.

(4): Done.

gc00:
@karya0 wrote:
> Apparently, the clean-64 was needed because of an earlier bug/feature that removed the bin/lib directories from the build tree.
> ...
> I think it's okay to update the build instructions specifically if we are making it simpler.

I've also checked now, and I agree.  In the current version of DMTCP (both master and version 2.4.3), clean-64 is no longer needed.

1. Let's do as you suggest:  update the build instructions to use `make clean` instead of `make clean-64` in `doc/multi-arch.txt`.
2. As for the `clean-64` target of `Makefile.in`, let's leave that as is for this PR (bug fixes only for this PR, since there's a tiny chance that some user's code somewhere refers to `clean-64`).  It's fine to delete the clean-64 target in a 2.5.0 PR (e.g., in PR #276, if you like).

After that, it looks good to me.  Nice work!

PR #276, opened 2016-01-04
Minor updates to test/Makefile.in

karya0:


gc00:
@karya0,
    Of the two commits, I'll propose to _not_ do "Compacted patterns for test/Makefile rules."

1. The commit (" Fixed 'make tests' for independent build root.") is a good one and it can be pushed in.  It uses standard idioms that we use throughout and is less error-prone when we modify Makefile.
2.  The commit ("Compacted patterns for test/Makefile rules.") uses a more recent extension to Makefile's.  The pros are that it's a nice feature, and once one thinks carefully about the syntax that you use, it becomes obvious what it is trying to do.  The cons are that students often want to contribute to DMTCP.  Also, we want to maintain good portability -- potentially even with other Makefile systems.  So, I'd prefer not to standardize DMTCP on a new feature -- for the sake of saving a few lines of code in the Makefile.

PR #277, opened 2016-01-07
Add test/file2.c: the test opens file and dirs and unlinks

gc00:
This should be trivial to review.  The test for `autotest.py` was written a while ago, but we had not been allowing enough time for the test to complete in the past.


rohgarg:
LGTM.

PR #278, opened 2016-01-09
configure: --enable-static-libstdcxx

gc00:
This configure option causes $CXX to use the g++ flag `-static-libstdc++`
  When a newer and older distro share a filesystem, and when
  g++ builds on the newer distro, then the executable fails on
  the older distro.  This should fix it.

An explanation of the problem occurs here:
  https://hosunghwang.wordpress.com/2015/07/15/libstdc-so-6-library-mismatch-problem-and-solution/
(except that the author meant to refer to `-static-libstdc++` at the end of the article).

I haven't yet found a testing machine with the right packages to test this.  So, it's not yet ready to be pushed in.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -289,6 +289,48 @@ else
   AC_SUBST([M32], [0])
 fi
 
+AC_ARG_ENABLE([static_libstdcxx],
+            [AS_HELP_STRING([--enable-static-libstdcxx],
+                            [Compile $CXX with -static-libstdc++.
+                             This allows DMTCP executables compiled on later
+                             distro to run correctly on earlier distros.)])],
+            [use_static_libstdcxx=$enableval],
+            [use_static_libstdcxx=no])
+
+if test "$use_static_libstdcxx" = "yes"; then
+  CXXFLAGS="$CXXFLAGS -static-libstdc++"
+
+  AC_MSG_CHECKING([whether $CXX compiles using -static-libstdc++])
+  AC_LANG_PUSH([C++])
+  AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+  #include <stdio.h>
+  #include <features.h>  /* This often causes a problem. */
+  ]], [[
+  printf("Hello, world!\n");
+  ]])], [cplusplus_static_libstdcxx='yes'], [cplusplus_static_libstdcxx='no'])
+  AC_LANG_POP([C++])
+  AC_MSG_RESULT([$cplusplus_static_libstdcxx])
+  if test "$cplusplus_static_libstdcxx" = "no"; then
+    AC_MSG_FAILURE([[Failed to build $CXX program with -static-libstdc++
+    Typically, this means that you are missing a static package.
+    Consider installing libstdc++-static (Red Hat/Fedora/CentOS)
+     or the equivalent packages for your Linux distro.]])
+  fi
+
+  # We do this hack for cross-distro compatibility only if we already
+  #   requested the other hack, --enable-static-libstdcxx
+  OLDPATH="$PATH"
+  if test "$MKDIR_P" = "/usr/bin/mkdir -p" && \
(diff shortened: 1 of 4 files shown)


gc00:
@karya0, I've now tested this.  If I configure and make on Fedora 22, I can now use DMTCP on the older Ubuntu 10.04.  So, this is ready to re-re-review.

(Minor caveat:  Fedora 22 configures to use /usr/bin/mkdir and Ubuntu 10.04 only supports /bin/mkdir.  Also, Ubuntu wants to re-make some files that are already made (due to clock skew??).  But these are minor things that can trivially be fixed later.  Let's push this in, and worry about the minor stuff at a later time.)

gc00:
And the fix for`mkdir` so that Fedora 22 uses `/bin/mkdir` is to change `configure.ac` to:
```
OLDPATH="$PATH"
PATH="/bin:$PATH"
MKDIR_P
PATH="$OLDPATH"
```
I can create a separate commit and pull request for this later.

src/Makefile.am
```diff
@@ -115,12 +113,21 @@ __d_libdir__libdmtcp_so_LDADD = libdmtcpinternal.a libjalib.a \
...
 			  libnohijack.a -lpthread -lrt
 __d_bindir__dmtcp_launch_LDADD  = libdmtcpinternal.a libjalib.a \
-			  libnohijack.a -ldl -lpthread -lrt
+			  libnohijack.a -lpthread -lrt -ldl
 __d_bindir__dmtcp_restart_LDADD     = libdmtcpinternal.a libjalib.a \
 			  libnohijack.a -lpthread -lrt -ldl
 __d_bindir__dmtcp_command_LDADD     = libdmtcpinternal.a libjalib.a \
 			  libnohijack.a -lpthread -lrt -ldl
 
+# On a shared filesystem, a newer version of a distro will use a newer
```
karya0: Shouldn't we add this with an AM_CONDITIONAL check?

gc00:
@karya0: I added the login for `/bin/mkdir` to the current commit.  The commit message now says it all:
```
    configure: --enable-static-libstdcxx, use /bin/mkdir
    
    * This causes $CXX to use flag -static-libstdc++
      When a newer and older distro share a filesystem, and when
      g++ builds on the newer distro, then the executable fails on
      the older distro.  This should fix it.
    * If /bin/mkdir present, but not in user path, /usr/bin/mkdir
      is used instead.  If another distro (e.g., Ubuntu doesn't provide
      /usr/bin/mkdir, this creates a problem.).  /bin/mkdir is now preferred.
```

gc00:
@karya0:  Concerning your inline comment, "Shouldn't we add this with an AM_CONDITIONAL check?", I'll rely on you for best coding practices.  I tend to just write code for automake following the template, since we're planning to move to Cmake anyway.

gc00:
@karya0, I believe this now has the AM_CONDITIONAL check.  Is it ready to go?

configure.ac
```diff
@@ -289,6 +289,53 @@ else
...
+            [AS_HELP_STRING([--enable-static-libstdcxx],
+                            [Compile $CXX with -static-libstdc++.
+                             This allows DMTCP executables compiled on later
+                             distro to run correctly on earlier distros.)])],
+            [use_static_libstdcxx=$enableval],
+            [use_static_libstdcxx=no])
+
+AM_CONDITIONAL(USE_STATIC_LIBSTDCXX, [test "$use_static_libstdcxx" = "yes"])
+if test "$use_static_libstdcxx" = "yes"; then
+  AC_SUBST([STATIC_LIBSTDCXX], [1])
```
karya0: We don't seem to be using this variable, so why not delete this here and in the else part.

karya0:
I have a few high-level questions. Why are we not compiling libdmtcp.so with static libstdc++? Wouldn't this cause failures too? However, if we compile it with static libstdc++, wouldn't that potentially conflict with the application binaries that are compiled with the local version of libstdc++?

PR #282, opened 2016-01-15
Wrappers for pthread_mutex lock/unlock/trylock.

jiajuncao:
Mutexes have a field recording the real tid of the owner. If a
checkpoint/restart happens between pthread_mutex_lock() and
pthread_mutex_unlock(), then the real tid will change.

Support for rwlock/condition variables are still needed.

Duplicate of #173.


jiajuncao:
```
pthread_mutex_lock(mutex):
  rc = _real_pthread_mutex_lock(mutex)
  mapMutexVirtOwner[mutex] = dmtcp_gettid() // virtual tid
  return rc

pthread_mutex_unlock(mutex):
  rc =  _real_pthread_mutex_unlock(mutex)
  if error(rc) == EPERM:
      if mapMutexVirtOwner[mutex] == dmtcp_gettid(): // same virtual tid as prior to ckpt
        // then it's a DMTCP issue; patch the owner
        patch_mutex_owner(mutex, kernel_gettid()) // set owner to kernel tid (same as what libc saved)
        rc =  _real_pthread_mutex_unlock(mutex)  // This second time, there will be no EPERM
        mapMutexVirtOwner[mutex] = 0 // not strictly necessary, but helpful for debugging
  return rc
```

gc00:
Here are three other cases for which we'll need to check if we need this type of solution.
  pthead_mutex_timedlock
  pthread_cond_signal/wait
  pthread_rwlock_*
Those cases can be the subject of a later PR (and github issue) as needed.
(Thanks @jiajuncao and @ankitcse07 for suggesting these additional cases.)

gc00:
I tested this PR on a simple test program (that I'll place later in a separate PR).  This commit worked.  The test program uses  `PTHREAD_MUTEX_RECURSIVE` with a recursive use of  `pthread_mutex_timedlock()`.  For some reason, the same test failed when I tried it using `pthread_mutex_trylock()`.  I'll investigate further.

Anyway, normal mutexes work fine even without this Pr.  Mutexes with `PTHREAD_MUTEX_RECURSIVE` require this code to work correctly on restart, if the restart does a recursive pthread_mutex_lock() (or timedlock).  In my experiments, this code slows down pthread_mutex_lock/unlock by a factor of about ten.  Perhaps it is the extra call, `rc = __real_pthread_mutex_lock(mutex);`, and we should just accept this slowdown.

At least, let's add a comment at the wrappers for pthread_mutex_lock/unlock/trylock in which we say that this wrapper slows down code calling mutexes in a tight loop by a factor of about ten.

gc00:
@jiajuncao, here are some changes that I'd like to propose:

1.  Let's change the filename for `pthread_mutex_wrappers.cpp` to `mutexwrappers.cpp`  (no underscore to be consistent with other filenames, and keep the name short but understandable).
This is analogous to the existing `threadwrappers.cpp`.

2.  `pthread_mutex_trylock` can fail with `EBUSY` if this is a recursvie lock and the original cal to trylock() was by the same thread, but happened before a ckpt/restart.  To fix this bugy, we need to add to the trylock wrapper code that is similar to the code for `pthreadm_mutex_unlock`, but it should be for `EBUSY` instead of for `EPERM`.

3.  Let's also add `pthread_mutex_timedlock`.  It's a related syscall.  For `phread_mutex_timedlock()`, it can fail with `ETIMEDOUT` (and possibly `EDEADLOCK` if a previous tid happens to coincide with a current tid and make it look like deadlock).  So, if we own the mutex, we should set the owner back to us.

4.  DMTCP currently has wrappers for `pthread_mutex_lock` and `pthread_mutex_unlock` in src/miscwrappers.cpp.  I assume we don't need this.  They were added in commit `commit e0f8e8` by @tdenniston.  Let's remove them in a separate commit by itself (but as part of this same pull request), so that it will be easy to debug if that causes a problem.

5.  The reason that the wrapper for mutex lock/unlock/etc is slow is because the mapMutexVirtTid[] thing from STL/C++ is quite slow.  (Luckily, we don't invoke these maps in `unlock` unless we see an `EPERM`.)  Let's open a separate issue about this.  C++ appears to support hashing (http://en.cppreference.com/w/cpp/utility/hash).  Here's another commentary on C++ and hashing: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3333.html

jiajuncao:
@gc00,
I'd propose to change the filename from `pthread_mutex_wrappers.cpp` to `pid_mutexwrappers.cpp`, since it's in the pid plugin.
Also, can you try the STL version of hash map, `unordered_map`, and see if the runtime overhead is lower?

gc00:
@jiajuncao:  I like the name `pid_mutexwrappers.cpp` better.  Good catch!  These wrappers are changing the pid of the mutexes, and so the prefix `pid_` is more accurate.  (It's also why it belongs in the pid plugin.)

I also like your idea of using `unordered_map` for this.  Almost surely, that's the right choice here.  If I understand the documentation, STL will still force us to write our own hash function.  But since a tid is just an integer, it should be easy fo find a good hash function on the web.

Apparently, the DMTCP utils don't support `unordered_map` at this time.  So, let's finish this PR using the plain `map` instead of `unordered_map`.  But let's write a `FIXME:` where we define this `map`, saying that it should be an `unordered_map`.  I'll open the github issue referred to [in this earlier comment (as item 5)](https://github.com/dmtcp/dmtcp/pull/282#issuecomment-172246960).  We can switch from `map` to `unordered_map` in a later PR.

Finally, I modified my PR #283 to include tests both for recursive mutexes and for error-checking mutexes.  When you have the next version of this PR ready, could you test it against all three mutex tests to be found in PR #283?

src/plugin/pid/pthread_mutex_wrappers.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include "jassert.h"
+#include "pidwrappers.h"
+
+#define __real_pthread_mutex_lock    NEXT_FNC(pthread_mutex_lock)
+#define __real_pthread_mutex_trylock  NEXT_FNC(pthread_mutex_trylock)
+#define __real_pthread_mutex_unlock  NEXT_FNC(pthread_mutex_unlock)
+
+/* Globals structs */
+/* This maps mutex addresses to the virtual tid of the owner thread. */
```
gc00: @jiajuncao, please add a comment -- something like:
```
// FIXME:  This STL `map` should be an `unordered_map` (hash array) for better efficiency.
```

jiajuncao:
@gc00,
`unordered_map` does provide a default hash function, see: http://www.cplusplus.com/reference/unordered_map/unordered_map/

>Hash
A unary function object type that takes an object of type key type as argument and returns a unique value of type size_t based on it. This can either be a class implementing a function call operator or a pointer to a function (see constructor for an example). This defaults to hash\<Key\>, which returns a hash value with a probability of collision approaching 1.0/std::numeric_limits\<size_t\>::max().
The unordered_map object uses the hash values returned by this function to organize its elements internally, speeding up the process of locating individual elements.
Aliased as member type unordered_map::hasher.

jiajuncao:
@gc00,
I've been rethinking this PR when implementing the remaining functionalities required by `pthread_mutex_trylock()`, `pthread_mutex_timedlock()`, and different mutex types, such as `PTHREAD_MUTEX_RECURSIVE` and `PTHREAD_MUTEX_ERRORCHECK`. The logic can be quite complex if all these have to be done, because these functions may have different behavior when dealing with different mutex types. Also, if we look at the issue out of the box, all the work in the PR is the result of the changing of the kernel tid of the mutex, due to a checkpoint/restart. Can we solve it by simply fixing the owner of the mutex on restart? Since we have the `dmtcp_event_hook()` functionality, it's easy to do so. Here is the pseudocode:

```

list mutexList; // Global variable to record the list of mutexes

// In the following wrappers, add the mutex to mutexList
pthread_mutex_lock(pthread_mutex_t *mutex)
pthread_mutex_trylock(pthread_mutex_t *mutex)
pthread_mutex_timedlock(pthread_mutex_t *mutex)

dmtcp_event_hook(DmtcpEvent_t event, DmtcpEventData_t* data) {
  switch(event)  {
    case DMTCP_EVENT_THREADS_RESUME:
      if (data->resumeInfo.isRestart) {
        for mutex in mutexList {
          // The mutex is currently held by certain thread
          if (mutex->__data.__owner != 0) {
            mutex->__data.__owner = dmtcp_virtual_to_real_pid(dmtcp_gettid());
          }
        }
      }
      break;
  }
}

```

gc00:
@jiajuncao, I've now added a mutex4 test to PR #283.  (I modified the existing commit, instead of creating a second one.)  When your revision of this PR is ready, could you test it against the four mutex tests in PR #283?

@ankitcse07, the next revision of this PR by @jiajuncao should be ready for testing.  We believe that it will fix the bug that you were seeing.  Any feedback is welcome.

gc00:
@jiajuncao:  If this version with the fix for `theCkptCanStart` is passing your tests, let's commit it.  I agree that it would be nice to know why the earlier version had a bad interaction with libpthread.so, but we can look at that some other time.

PR #284, opened 2016-01-17
dmtcp_restart: obey DMTCP_CHECKPOINT_DIR env var

gc00:
`dmtcp_restart --help` says that `--ckptdir` and the environment variable `DMTCP_CHECKPOINT_DIR` can both be used to set the checkpoint directory.  But only `--ckptdir` was implemented.  This now also implements the environment variable `DMTCP_CHECKPOINT_DIR`.


karya0:
LGTM

PR #286, opened 2016-01-17
modify-env plugin now warns if no "dmtcp_env.txt"

gc00:
The title shows what this PR does.

Ideally, the code would also check if the `"DMTCP_QUIET"` environment variable was set prior to checkpoint.  But when the code calls `getenv("DMTCP_QUIET")` at restart time, the result always points to an unmapped address.  (Note that prior to checkpoint, DMTCP sets this environment variable to the string "0" if the environment variable had not been set.  But the pointer to the string "0" is being corrupted when the plugin tries to test for it during restart.)

Therefore, the current code has commented out the test for `"DMTCP_QUIET"`.

### DIFF
--- a/plugin/modify-env/modify-env.c
+++ b/plugin/modify-env/modify-env.c
@@ -30,6 +30,7 @@
 char * read_dmtcp_env_file(char *file, int size);
 int readAndSetEnv(char *buf, int size);
 int readall(int fd, char *buf, int maxCount);
+extern void warning(const char *warning_part1, const char *warning_part2);
 
 EXTERNC int dmtcp_modify_env_enabled() { return 1; }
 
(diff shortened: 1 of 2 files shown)


plugin/modify-env/modify-env.c
```diff
@@ -49,7 +49,22 @@ void dmtcp_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
...
+    if (buf != NULL) { // If env_file exists
+      readAndSetEnv(buf, size);
+    } else { // else env_file doesn't exist (buf == NULL)
+#if 0
+      // FIXME:  This "if" condition to check environ var. always triggers,
+      //   even if environ var. was never present.  Uncomment this when fixed.
+      if (!getenv("DMTCP_QUIET") &&
+          strcmp(getenv("DMTCP_QUIET")[0], "0") != 0) {
+#endif
+        fprintf(stderr,
```
karya0: Let's use JWARNING instead.

karya0:
LGTM modulo the comment about JWARNING.

gc00:
@karya0, I've added a new file, warning.cpp, which implements JWARNING.  I checked it, and it correctly warns if "dmtcp_env.txt".  Shall I push this in?

karya0:
This is alright, but I would have liked it more if we had converted this file into a cpp file and used JWARNING directly inside it rather then jumping through the hoop and using a separate cpp file just for warning. Having said that, it's inside a plugin and so I can live with that :-).

gc00:
Yep.  I still prefer C for lower-level stuff.  This way, I can copy my cpp file to other C code, just to be ornery.  :-)

PR #288, opened 2016-01-20
dmtcp_coordinator: fix --exit-after-ckpt

gc00:
Currently, `dmtcp_coordinator --exit-after-ckpt` causes the processes of the current computation to all be killed immediately after the first checkpoint.  It is similar to  `dmtcp_coordinator --exit-on-last`, except that `--exit-on-last` causes the coordinator to exit when the last peer process exits.

The bug is that after a checkpoint, the coordinator correctly kills all peer processes, and then the coordinator turns off the `--exit-after-ckpt` feature.  The help message implies that the `--exit-after-ckpt` feature is persistent, and if a new computation starts on this coordinator, the new computation should also be killed after the checkpoint.  But instead, the feature is turned off after the first checkpoint.

The original code is by @karya0, and dates from May, 2013.  Maybe this is old code that always had this bug.  Comments?

This PR makes the  `--exit-after-ckpt` feature persistent, and also provides an improved error message.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -116,7 +116,7 @@ static const char* theUsage =
   "  --exit-on-last\n"
   "      Exit automatically when last client disconnects\n"
   "  --exit-after-ckpt\n"
-  "      Exit automatically after checkpoint is created\n"
+  "      Kill peer processes of computation after first checkpoint is created\n"
   "  --daemon\n"
   "      Run silently in the background after detaching from the parent process.\n"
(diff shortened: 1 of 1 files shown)


karya0:
We need to support "xc" command which kills the computation after checkpoint even if the "--exit-after-checkpoint" flag was not specified when launching the coordinator. We can do this by creating an additional variable to handle the "xc" command.

gc00:
@karya0: I've now changed the commit to also support the "xc" command typed directly to the coordinator.  The code now uses two variables in dmtcp_coordinator.cpp:  `exitAfterCkpt` and `exitAfterCkptOnce`.

karya0:
Let's merge it.

PR #289, opened 2016-01-24
Added support for non-blocking connect.

karya0:
Added a new TCP type, TCP_CONNECT_IN_PROGRESS, that indicates a
non-blocking connect. During drain, we call select() to check whether
the socket is writable or not. If it's still not writable after 60
seconds, we mark the socket as external and continue.

### DIFF
--- a/src/plugin/ipc/socket/socketconnection.cpp
+++ b/src/plugin/ipc/socket/socketconnection.cpp
@@ -319,7 +319,9 @@ void TcpConnection::onListen(int backlog)
   _listenBacklog = backlog;
 }
 
-void TcpConnection::onConnect(const struct sockaddr *addr, socklen_t len)
+void TcpConnection::onConnect(const struct sockaddr *addr,
+                              socklen_t len,
+                              bool connectInProgress)
 {
   if (really_verbose) {
     JTRACE("Connecting.") (id());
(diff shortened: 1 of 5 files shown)


src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -381,6 +386,38 @@ void TcpConnection::drain()
...
+    } else if (FD_ISSET(_fds[0], &wfds)) {
+      int val = -1;
+      socklen_t sz = sizeof(val);
+      getsockopt(_fds[0], SOL_SOCKET, SO_ERROR, &val, &sz);
+      JTRACE("Connect-in-progress socket is now writable.") (_fds[0]);
+      _type = TCP_CONNECT;
+    } else {
+      JWARNING(false) (_fds[0])
+        .Text("Failed to wait for EINPROGRESS connect to return.\n"
+              "Marking the socket as external and continuing.");
```
gc00: @karya0:  I propose this more verbose explanation for the `JWARNING`.  It will help users unfamiliar with the internals of DMTCP.
```
       .Text("connect() returned EINPROGRESS.  Socket still not writable\n"
             "after 60 seconds.   The socket is probably connected to an\n"
             "external process not under DMTCP control.\n"
             "Marking this socket as external and continuing to checkpoint.");
```

PR #296, opened 2016-01-28
Support for rsh command in dmtcp package. 

ankitcse07:
This enhancement also take cares if both rsh/ssh are simultaneously used for launching processes attached to same coordinator. It maintains information of remote shell command used for each process. This information is stored in checkpoint database as well as restart script maintains list of processes lauched via particular remote shell command.

Architecture is exactly same, its just maintaining bookeeping information for each process shell type.

This was long pending support required for Veloce Integration

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -665,18 +665,18 @@ void CoordinatorAPI::sendCkptFilename()
     msg.type = DMT_CKPT_FILENAME;
   }
   // Tell coordinator type of remote shell command used ssh/rsh
-  string shellType("");
+  const char* shellType = "";
   const char *remoteShellType = getenv(ENV_VAR_REMOTE_SHELL_CMD);
   if(remoteShellType != NULL) {
     shellType = remoteShellType;
   }
 
   JTRACE("recording filenames") (ckptFilename) (hostname) (shellType);
-  msg.extraBytes = ckptFilename.length() + 1 + hostname.length() + 1+ shellType.length() + 1;
+  msg.extraBytes = ckptFilename.length() + 1 + hostname.length() + 1+ strlen(shellType) + 1;
  
   _coordinatorSocket << msg;
   _coordinatorSocket.writeAll(ckptFilename.c_str(), ckptFilename.length() + 1);
-  _coordinatorSocket.writeAll(shellType.c_str(), shellType.length() + 1);
+  _coordinatorSocket.writeAll(shellType, strlen(shellType) + 1);
   _coordinatorSocket.writeAll(hostname.c_str(), hostname.length() + 1);
 }
 
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -658,16 +658,25 @@ void CoordinatorAPI::sendCkptFilename()
   // Tell coordinator to record our filename in the restart script
   string ckptFilename = ProcessInfo::instance().getCkptFilename();
   string hostname = jalib::Filesystem::GetCurrentHostname();
-  JTRACE("recording filenames") (ckptFilename) (hostname);
   DmtcpMessage msg;
   if (dmtcp_unique_ckpt_enabled && dmtcp_unique_ckpt_enabled()) {
     msg.type = DMT_UNIQUE_CKPT_FILENAME;
   } else {
     msg.type = DMT_CKPT_FILENAME;
   }
-  msg.extraBytes = ckptFilename.length() + 1 + hostname.length() + 1;
+  // Tell coordinator type of remote shell command used ssh/rsh
+  string shellType("");
+  const char *remoteShellType = getenv(ENV_VAR_REMOTE_SHELL_CMD);
+  if(remoteShellType != NULL) {
+    shellType = remoteShellType;
+  }
+
+  JTRACE("recording filenames") (ckptFilename) (hostname) (shellType);
+  msg.extraBytes = ckptFilename.length() + 1 + hostname.length() + 1+ shellType.length() + 1;
+ 
   _coordinatorSocket << msg;
   _coordinatorSocket.writeAll(ckptFilename.c_str(), ckptFilename.length() + 1);
+  _coordinatorSocket.writeAll(shellType.c_str(), shellType.length() + 1);
   _coordinatorSocket.writeAll(hostname.c_str(), hostname.length() + 1);
 }
 
--- a/src/plugin/ipc/ssh/dmtcp_ssh.cpp
+++ b/src/plugin/ipc/ssh/dmtcp_ssh.cpp
@@ -146,16 +146,14 @@ int main(int argc, char *argv[], char *envp[])
    */
 
   shift;
-  while(true) {
+  while (true) {
     if ( strcmp(argv[0], "--noStrictHostKeyChecking") == 0 ) {
       noStrictHostKeyChecking = 1;
       shift;
-    }
-    else if ( strcmp(argv[0], "--rsh-slave") == 0 ) {
+    } else if ( strcmp(argv[0], "--rsh-slave") == 0 ) {
       isRshProcess = 1;
       shift;
-    }
-    else if ( strcmp(argv[0], "--ssh-slave")== 0 ) {
+    } else if ( strcmp(argv[0], "--ssh-slave")== 0 ) {
       isRshProcess = 0;
       shift;
     } else {
--- a/src/plugin/ipc/ssh/dmtcp_ssh.cpp
+++ b/src/plugin/ipc/ssh/dmtcp_ssh.cpp
@@ -16,6 +16,7 @@
 
 static int listenSock = -1;
 static int noStrictHostKeyChecking = 0;
+static int isRshProcess = 0; 
 
 extern "C" void dmtcp_get_local_ip_addr(struct in_addr *addr) __attribute((weak));
 
@@ -125,6 +126,9 @@ static int waitForConnection(int listenSock)
   return fd;
 }
 
+//shift args
+#define shift argc--,argv++
+
 int main(int argc, char *argv[], char *envp[])
 {
   int in[2], out[2], err[2];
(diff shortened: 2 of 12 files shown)


ankitcse07:
Hi All 
      Please review this change, currently this support is in my local repository which I would really like to part of public repository.

src/coordinatorapi.cpp
```diff
@@ -658,16 +658,25 @@ void CoordinatorAPI::sendCkptFilename()
...
     msg.type = DMT_UNIQUE_CKPT_FILENAME;
   } else {
     msg.type = DMT_CKPT_FILENAME;
   }
-  msg.extraBytes = ckptFilename.length() + 1 + hostname.length() + 1;
+  // Tell coordinator type of remote shell command used ssh/rsh
+  string shellType("");
+  const char *remoteShellType = getenv(ENV_VAR_REMOTE_SHELL_CMD);
+  if(remoteShellType != NULL) {
+    shellType = remoteShellType;
```
gc00: Will this create a memory leak?  We're converting a C string to a new string object.  shellType used to point to a string object, "", and now points to a different string object.  How do we know that these two string objects are both freed?

I'd have to check up on the rules of C++ to be sure what happens here.  But if it's possibly to make shellType a `char *` instead of a `string`, then anybody reading this code will immediately know that it doesn't leak memory.

src/restartscript.cpp
```diff
@@ -409,17 +429,49 @@ void writeScript(const string& ckptDir,
...
+    }
+    string defaultShellType;
     for ( host=restartFilenames.begin(); host!=restartFilenames.end(); ++host ) {
       fprintf ( fp, "\n :: %s :bg:", host->first.c_str() );
       for ( file=host->second.begin(); file!=host->second.end(); ++file ) {
         fprintf ( fp," %s", file->c_str() );
       }
+
+      /* 
+       * Process which are launched on local machine without using rsh/ssh
```
gc00: `using` -> `using an`

src/restartscript.cpp
```diff
@@ -409,17 +429,49 @@ void writeScript(const string& ckptDir,
...
+       * command are part of restartFilenames. So we have to choose default
```
gc00: `default` -> `the default`

src/restartscript.cpp
```diff
@@ -409,17 +429,49 @@ void writeScript(const string& ckptDir,
...
+       * shell command for them.  Search if some process is launched on local
```
gc00: `local` -> `the local`

src/restartscript.cpp
```diff
@@ -409,17 +429,49 @@ void writeScript(const string& ckptDir,
...
+       * machine via rsh/ssh command and use that, in case its not found
```
gc00: `its` -> `it's`

src/restartscript.cpp
```diff
@@ -409,17 +429,49 @@ void writeScript(const string& ckptDir,
...
+       * give preference to ssh if no rsh command is absent.
```
gc00: ` no rsh command is absent.` -> ` when no rsh command is found.`

gc00:
@ankitcse07, I still have to do some more reviewing.  My remaining worry is whether the maps for rshFilenames, etc., can create problems.  DMTCP uses its own small memory allocator for DMTCP code.  This is so that it doesn't interfere with the user's allocator.  If we use too much memory, then bad things can happen.

We also have to be careful about memory allocation at checkpoint time, since a checkpoint request can happen at any time.  But after I inspect the coede, I think that this won't be a problem.

src/plugin/ipc/ssh/ssh.cpp
```diff
@@ -338,17 +362,22 @@ static void prepareForExec(char *const argv[], char ***newArgv)
...
-  memset(new_argv, 0, sizeof(char*) * (nargs + 10));
+  char** new_argv = (char**) JALLOC_HELPER_MALLOC(sizeof(char*) * (nargs + 11));
+  memset(new_argv, 0, sizeof(char*) * (nargs + 11));
 
   size_t idx = 0;
   new_argv[idx++] = (char*) dmtcp_ssh_path.c_str();
   if (noStrictChecking) {
     new_argv[idx++] = const_cast<char*>("--noStrictHostKeyChecking");
   }
+  if(isRshProcess)
```
rohgarg: nit: Please change:
```C
if(..)
  statement;
else
  statement;
```
to:
```C
if (...) {
  statement;
} else {
  statement;
}
```
everywhere.

src/plugin/ipc/ssh/ssh.cpp
```diff
@@ -465,10 +505,14 @@ extern "C" int execvp (const char *filename, char *const argv[])
 extern "C" int execvpe (const char *filename, char *const argv[],
                          char *const envp[])
 {
-  if (jalib::Filesystem::BaseName(filename) != "ssh") {
+  if ((jalib::Filesystem::BaseName(filename) != "ssh") && 
+      (jalib::Filesystem::BaseName(filename) != "rsh")) {
     return _real_execvpe(filename, argv, envp);
   }
 
+ if(jalib::Filesystem::BaseName(filename) == "rsh")
```
rohgarg: nit:
```C
if(...)
  statement;
```
-->
```C
if (...) {
  statement;
}
```

src/plugin/ipc/ssh/dmtcp_ssh.cpp
```diff
@@ -136,9 +140,27 @@ int main(int argc, char *argv[], char *envp[])
...
-    noStrictHostKeyChecking = 1;
-    argv++;
+  /* command line parsing was assuming the location of arguments
+   * so moving to more robust shift based mechanism used at other
+   * places too
+   */
+
+  shift;
+  while(true) {
+    if ( strcmp(argv[0], "--noStrictHostKeyChecking") == 0 ) {
```
rohgarg: nit:
```C
if ( ...) {
  ...
}
else if ( ...) {
  ...
} else {
  ...
}
```
-->
```C
if (...) {
  ...
} else if (...) {
  ...
} else {
 ...
}
```

src/plugin/ipc/ssh/dmtcp_ssh.cpp
```diff
@@ -136,9 +140,27 @@ int main(int argc, char *argv[], char *envp[])
...
-  if (strcmp(argv[1], "--noStrictHostKeyChecking") == 0) {
-    noStrictHostKeyChecking = 1;
-    argv++;
+  /* command line parsing was assuming the location of arguments
+   * so moving to more robust shift based mechanism used at other
+   * places too
+   */
+
+  shift;
+  while(true) {
```
rohgarg: nit: `while(true)` --> `while (true)`

rohgarg:
@ankitcse07: Thanks for the patch. I just have some minor inline comments about the coding style. Other than that, it looks good to me.

gc00:
@ankitcse07, I'm done reviewing this.  I have some minor inline comments.  Aside from that, I have one larger suggestion (also stated inline, above):

* **UPDATE:** I checked the C++ rules. I think that C++ will not leak memory with shellType. But it would still be better to instead use: char *shellType=NULL;
This way, you're not stopping to create a complex string object, only to later use .c_str() to extract the actual string. Note also that strcat(), strcpy(), and strcmp() exist to directly manipulate `char *'. In this way, you avoid the overhead of thinking about the hidden C++ code in its constructors and destructors. What you see is what you get. And it's rougly the same number of lines of code as with C++.

@karya0, you know C++ better than me.  Could you also do a quick review, to make sure that the C++ style is good?  Thanks,

ankitcse07:
Thanks @gc00  @rohgarg for the detailed review. I had incorporated all changes suggested by you.

ankitcse07:
I am creating testcases, what is procedure these to the repository ? Should I create another pull request for the same

gc00:
@ankitcse07, You can examine the examples in the subdirectory, "test".  The basic idea is to write a C program, and to specify how many processes should be running at the time of the checkpoint or restart.  The system will check if the correct number of processes are running.  If they aren't, the system will indicate "FAIL".

After creating a new test (for example, rsh.c), you would add that test to autotest.py.  It should be easy to add a test to autotest.py by following the other examples there.  But I'm happy to do that for you if you like.  If rsh.c is normally compiled with no special flags, then the Makefile will automatically detect rsh.c in the test directory, and compile it.

ankitcse07:
Are there any milestones for this

ankitcse07:
Should I rebase this from 2.5 branch , there seems to be conflicts with master branch.

gc00:
@ankitcse07, Sorry that no one answered your question above from March 22.  I'll do that now.

1.  Thank you for pushing this into the 2.5 branch as commit #363.

2.  Could you also resolve the conflicts and push this into the branch master?  If that would be a burden for you, please tell me, and I'm happy to do that myself.  The issue is that version 2.5 will be the last version in the 2.x series.  After that (maybe at the beginning of 2017), we will switch to a slightly incompatible version 3.0.  So, we'll want your enhancement there also.  Naturally, we will add documentation on 3.0, and also work with the users to help them fix any incompatibilities between 2.5 and 3.0.  Thanks for your understanding.

ankitcse07:
@gc00 Thanks for responding on this.  I will push this into master branch but there were some hang issues with my merge due to architectural changes related to barriers.  I would point the test hanging without my changes and then would push this into master branch.  

One unrelated question, you said beginning 2017, you will switch to 3.0, does this means 2.5 would not have any future development check-ins and would only be bug fixes. If this is the case I might need to switch to 3.0.0 sometimes later in this year

gc00:
@ankitcse07 wrote:
> does this means 2.5 would not have any future development check-ins and would only be bug fixes.

That's exactly right.  In case there are any issues in porting to 3.0, we were planning to work with any contributors to DMTCP in the Fall.

Also, let us know if you have any problems with the test that hangs under 3.0 for this commit.

ankitcse07:
@gc00 Please find the testcase resulting in hang while invoking checkpoint from coordinator.  Just change the machines in rshell_test.cxx, invoke a.out under dmtcp_launch, further take checkpoint from dmctcp_coordinator window.  This will hang with processes stuck at SUSPENDED state. 

Let me know if you are not able to reproduce the issue with master branch. 

[test_ssh_hang_during_ckpt (1).tar.gz](https://github.com/dmtcp/dmtcp/files/312714/test_ssh_hang_during_ckpt.1.tar.gz)

ankitcse07:
@gc00 I had merged this change in master branch but getting above issue. So once this issue is resolved, I will commit my change

gc00:
@ankitcse07, I've now created a branch:  https://github.com/gc00/dmtcp/tree/ankit_rsh_support

I tested with:  `./configure && make -j clean && make -j ankit_rsh_support`

It works for `ssh`.  It wouldn't be easy for me to test for `rsh`.  Could you also test for `rsh` in your environment?  If it works for you also, you are welcome to either directly push it into the master branch, or else let me know, and I'll push it in.  The commits are still owned by you, and so you will still get the credit.  :-)

ankitcse07:
@gc00 I am not finding free cycles for this,  I will get back to you soon once I am done with my releases

gc00:
@ankitcse07, just pinging you on this.  Would you mind testing this PR again for "rsh".  If it works for you, let me know, and I'll be happy to resolve the conflicts and push it in.  Thanks.

rohgarg:
It seems like the commits have already been pushed to the 2.5 branch. Gene or I will port this 3.0.

gc00:
This commit has been pushed into 2.5.  It still has to be pushed into 3.0/master.  One of us will add a new PR against 3/0/master, to replace this PR, and then close this PR at that time.

gc00:
I've now created PR #478, in order to push a port of this code into DMTCP master (3.0).  So, it is now safe to close the current PR, and I am doing that here.

PR #297, opened 2016-01-28
IB: added detection of nested verbs call.

jiajuncao:
If one verb function is called by another internally, the real resources should be used, instead of the shadowed structures.

### DIFF
--- a/contrib/infiniband/ibv_internal.h
+++ b/contrib/infiniband/ibv_internal.h
@@ -1,9 +1,33 @@
 /*! \file ibv_internal.h */
 #include <infiniband/verbs.h>
 #include <stdbool.h>
+#include <inttypes.h>
 #include "ibvidentifier.h"
 #include "lib/list.h"
 #include "debug.h"
+
+
+/* Two 64-bit fixed, random numbers to verify whether a struct is shadowed
+ * (struct internal_ibv_XXX) or real (struct ibv_XXX).  This signature
+ * allows us to detect recursive cases:  A target application may call our
+ * wrapper using the shadowed struct.  We then pass on the real struct to the
+ * real function in the IB library.  But if the real function calls a second
+ * real IB function, then it will pass the real struct to our wrapper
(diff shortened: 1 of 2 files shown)


contrib/infiniband/ibv_internal.h
```diff
@@ -1,10 +1,26 @@
 /*! \file ibv_internal.h */
 #include <infiniband/verbs.h>
 #include <stdbool.h>
+#include <inttypes.h>
 #include "ibvidentifier.h"
 #include "lib/list.h"
 #include "debug.h"
 
+/* Two 64-bit random numbers to verify whether
+ a struct is shadowed or real. */
```
gc00: @jiajuncao, Here is a more verbose comment to replace the above lines.  I'd like to propose this to help other readers of the code, who don't know the background.
```
/* Two 64-bit fixed, random numbers to verify whether a struct is shadowed 
 * (struct internal_ibv_dev) or real (struct ibv_dev).  This signature
 * allows us to detect recursive cases:  A target application may call our
 * wrapper using the shadowed struct.  We then pass on the real struct to the
 * real function in the IB library.  But if the real function calls a second
 * real IB function, then it will pass the real struct to our wrapper 
 * for the second IB function. 
 */
```

gc00:
This PR will help a lot.  I have three comments:

1.  Expanded comment (see my inline note above)
2.  The field names `sig1` and `sig2` are not meaningful to me.  Suppose we called them `magic1` and `magic2`?  The term "magic number" is well-known in Linux, where it's used at the beginning of a file.
3.  Also, the names `STRUCT_SIG_INIT` and `STRUCT_SIG_VERIFY` are not meaningful to me.  What about the names `INIT_INTERNAL_IBV_TYPE` and `IS_INTERNAL_IBV_STRUCT`?  (The magic numbers are part of the internal implementation, but these names might be more meaningful.)

Please do improve on my suggestions, and between the two of us, we'll arrive at meaningful names for others.

PR #299, opened 2016-01-29
pathvirt-plugin: Initial commit

rohgarg:
This supersedes PR #265.

### DIFF
--- a/plugin/pathvirt/README
+++ b/plugin/pathvirt/README
@@ -1,0 +1,110 @@
+# pathvirt (Path Virtualization DMTCP Plugin)
+
+pathvirt is a DMTCP plugin that virtualizes access to file system paths for
+programs run under DMTCP. This allows path access to be transparently
+redirected elsewhere on the file system on DMTCP restart.
+
+## Usage
+
+pathvirt operates on the basis of path prefixes and is interfaced with using
+the DMTCP_ORIGINAL_PATH_PREFIX/DMTCP_NEW_PATH_PREFIX environment variable.
+
+On DMTCP init, set DMTCP_ORIGINAL_PATH_PREFIX to a colon delimited list of path
+prefixes to register them with pathvirt. This is required, but does not
+actually change anything about the execution of the program until the first
+restart. For example:
+
+    $ DMTCP_ORIGINAL_PATH_PREFIX=/home/user1/bin1:/home/user1/lib1:/home/user1/doc1 \
+        dmtcp_launch -i4 --with-plugin $DMTCP_ROOT/lib/dmtcp/libdmtcp_pathvirt.so \
+        ./my-program
+
+Then, on any DMTCP restart, one can optionally set DMTCP_NEW_PATH_PREFIX to a
+second colon-delimited list of path prefixes. For example:
+
+    $ DMTCP_NEW_PATH_PREFIX=/home/user2/bin2:/home/user2/lib2:/home/user2/doc2 \
+        ./dmtcp_restart_script.sh
+
+If one *does not* provide DMTCP_NEW_PATH_PREFIX on restart, pathvirt does nothing,
+and the program is restarted from the last checkpoint as normal.
+
+If one *does* provide it, pathvirt becomes active and starts performing path
+translation.  At this point, prefixes in this second list will dynamically
+replace corresponding prefixes in the first list when a program uses a path
+that is prefixed with an entry in the first colon list in a call to libc. For
+example:
+
+    open("/home/user1/bin1/ls") => open("/home/user2/bin2/ls")
+    open("/home/user1/lib1/libc.so.6") => open("/home/user2/lib2/libc.so.6")
+    open("/home/user1/doc1/proj/README") => open("/home/user2/doc2/proj/README")
+    open("/not/registered/prefix.txt") => open("/not/registered/prefix.txt")
+
+### Usage Rules
+
+1.  Environment variables:
+     - DMTCP_ORIGINAL_PATH_PREFIX: recognized at time of launch or restart (to override)
+     - DMTCP_NEW_PATH_PREFIX: recognized at time of restart only
+
+2.  Programmatic API:
+     - dmtcp_set_original_path_prefix_list()
+     - dmtcp_get_original_path_prefix_list()
+     - dmtcp_get_new_path_prefix_list()
+    NOTE: There is no dmtcp_set_new_path_prefix()
+
+3.  dmtcp_set_original_path_prefix() can be called at any time
+    (at checkpoint time, within a user's plugin wrapper function, etc.).
+
+    However, this will have no effect until the next restart.
+    The user must then set the environment variable DMTCP_NEW_PATH_PREFIX
+    prior to restart.
+
+4.  On restart, if the original path and new path are of different lengths,
+    this is a user error, and DMTCP will report the current old and new path,
+    and exit.
+    (The DMTCP_NEW_PATH_PREFIX can be set prior to restart to guarantee that
+    the path lists are of the same length.)
+
+5.  If there is more than one match of a pathname with a path prefix
+    in DMTCP_ORIGINAL_PATH_PREFIX, then the first match is used.  This is
+    useful for creating exceptions to a path translation.
+
+    As an example, suppose user2 wants to restart by using a checkpoint
+    image created by user1.  User2 wants to start in the directory of
+    user2, since he or she has no permission to write in the directory
+    of user1.  But user2 should continue to use the original configuration
+    file, /home/user1/config.txt .  The following solution will work:
+    > DMTCP_ORIGINAL_PATH_PREFIX=/home/user1/config.txt:/home/user1
+    > DMTCP_NEW_PATH_PREFIX=/home/user1/config.txt:/home/user2
+
+## Constraints
+
+- The maximum size permitted for the DMTCP_*_PATH_PREFIX environment variables
+  is 10*1024 bytes at time of writing. This configuration can be increased by
+  editing `plugin/pathvirt/pathvirt.cpp:MAX_ENV_VAR_SIZE` and recompiling.
+- The DMTCP_*_PATH_PREFIX environment variables are also constrained by a DMTCP
+  environment variable limit of 12*1024 bytes at time of writing. This
+  configuration can be increased by editing `src/dmtcpplugin.cpp:MAXSIZE` and
+  recompiling.
+- Virtualization of certain path prefixes such as `/proc` or `/dev` is
+  experimental and unsupported.
+- Strict POSIX compliance in regards to certain error handling (in particular,
+  EFAULT on invalid address) is not fully supported.
+
+## Testing
+
+To run a basic test, cd into plugin/pathvirt/test/slot5.
+
+    make check
+    *** wait for a checkpoint, then ctrl-c ***
+    make restart
+    *** wait for a checkpoint, then ctrl-c ***
+    make test
+
+You should observe that on launch, the slot5/***5/pv-test.txt files were being
+written to, but after restart the ../misc/slot7/***7/pv-test.txt files were
+being written to.
+
+---
+
+Authors:
+Mark Mossberg <mark.mossberg@gmail.com>
+Rohan Garg <rohgarg@ccs.neu.edu>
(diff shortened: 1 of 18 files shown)


balajir1990:
HI Rohan
  May i know which file has the new API. I couldn't find the dmtcp.h in the list of files changed
-Balaji R

rohgarg:
@balajir1990: The four new APIs are in `plugin/pathvirt/pathvirt.cpp`. Since this doesn't affect DMTCP in general, the APIs are restricted to the pathvirt plugin.

(In my most recent commit, I have moved the declaration of the four APIs to a separate header file.)

balajir1990:
oh I see but if i have moved the checkpoint and open files to another directory and try restart from there, how can i call the plugin's set API since dmtcp_restart would throw error about file not found before even returnig from dmtcp_checkpoint API call

rohgarg:
@balajir1990: The API was intended to allow changes at runtime and not during checkpoint/restart. Anyway, I have now extended the interface with two new environment variables -- `DMTCP_OLD_PATH_PREFIX` and `DMTCP_NEW_PATH_PREFIX` -- that you can use to specify the path prefixes at restart time. Is this what you had in mind?

balajir1990:
Exactly, this will allow me to change the path anytime. And is removing the size limit of 10kb part of this pull request?

rohgarg:
@gc00: Could you please review this one? I have incorporated the changes (and updated the README) as we discussed.

@balajir1990, @ankitcse07: Could you please check if this works for you?

plugin/pathvirt/README
```diff
@@ -0,0 +1,95 @@
+# pathvirt (Path Virtualization DMTCP Plugin)
+
+pathvirt is a DMTCP plugin that virtualizes access to file system paths for
+programs run under DMTCP. This allows path access to be transparently
+redirected elsewhere on the file system on DMTCP restart.
+
+## Usage
+
+pathvirt operates on the basis of path prefixes and is interfaced with using
+the DMTCP_PATH_PREFIX environment variable.
```
gc00: Did we agree that we will only support `DMTCP_ORIGINAL_PATH_PREFIX` and `DMTCP_NEW_PATH_PREFIX`?  If so, let's modify the sentence above.

plugin/pathvirt/README
```diff
@@ -0,0 +1,95 @@
...
+On DMTCP init, set DMTCP_ORIGINAL_PATH_PREFIX to a colon delimited list of path
+prefixes to register them with pathvirt. This is required, but does not
+actually change anything about the execution of the program until the first
+restart. For example:
+
+    $ DMTCP_ORIGINAL_PATH_PREFIX=/home/user1/bin1:/home/user1/lib1:/home/user1/doc1 \
+        dmtcp_launch -i4 --with-plugin $DMTCP_ROOT/lib/dmtcp/libdmtcp_pathvirt.so \
+        ./my-program
+
+Then, on any DMTCP restart, you can optionally set DMTCP_NEW_PATH_PREFIX to a
```
gc00: `you can` -> `one can`

plugin/pathvirt/README
```diff
@@ -0,0 +1,95 @@
...
+second colon delimited list of path prefixes. For example:
```
gc00: `colon delimited` -> `colon-delimited`

plugin/pathvirt/README
```diff
@@ -0,0 +1,95 @@
...
+
+    $ DMTCP_PATH_PREFIX=/home/user2/bin2:/home/user2/lib2:/home/user2/doc2 \
+        ./dmtcp_restart_script.sh
+
+If you *do not* provide DMTCP_NEW_PATH_PREFIX on restart, pathvirt does nothing,
```
gc00: `one *does not*`, or `If DMTCP_NEW_PATH_PREFIX is not provided on restart,`

`pathvirt does nothing` -> `pathvirt has no effect`

plugin/pathvirt/README
```diff
@@ -0,0 +1,95 @@
...
+and the program is restarted from the last checkpoint as normal.
+
+If you *do* provide it, pathvirt becomes active and starts performing path
```
gc00: `you *do*` -> `one *does*`

plugin/pathvirt/README
```diff
@@ -0,0 +1,95 @@
...
+    However, this will have no effect until the next restart.
+    The user must then set the environment variable DMTCP_NEW_PATH_PREFIX
+    prior to restart.
+
+4.  On restart, if the original path and new path are of different lengths,
+    this is a user error, and DMTCP will report the current old and new path,
+    and exit.
+    (The DMTCP_NEW_PATH_PREFIX can be set prior to restart to guarantee that
+    the path lists are of the same length.)
+
```
gc00: I suggest adding this fifth rule:
```
5.  If there is more than one match of a pathname with a path prefix 
    in DMTCP_ORIGINAL_PATH_PREFIX, then the first match is used.  This is
    useful for creating exceptions to a path translation.

    As an example, suppose user2 wants to restart by using a checkpoint 
    image created by user1.  User2 wants to start in the directory of
    user2, since he or she has no permission to write in the directory
    of user1.  But user2 should continue to use the original configuration
    file, /home/user1/config.txt .  The following solution will work:
    > DMTCP_ORIGINAL_PATH_PREFIX=/home/user1/config.txt:/home/user1
    > DMTCP_NEW_PATH_PREFIX=/home/user1/config.txt:/home/user2
```

plugin/pathvirt/README
```diff
@@ -0,0 +1,95 @@
...
+  10*1024 bytes at time of writing. This configuration can be increased by
+  editing `plugin/pathvirt/pathvirt.cpp:MAX_ENV_VAR_SIZE` and recompiling.
+- DMTCP_PATH_PREFIX is also constrained by a DMTCP environment variable
+  limit of 12*1024 bytes at time of writing. This configuration can be
+  increased by editing `src/dmtcpplugin.cpp:MAXSIZE` and recompiling.
+
+## Development
+
+### Testing
+
```
gc00: Since "Development" has just one sub-topic, let's remove "Development" and use just one second-level heading:  `## Testing`

plugin/pathvirt/pathvirt.h
```diff
@@ -0,0 +1,27 @@
...
+#  define EXTERNC
+# endif
+#endif
+
+/*  dmtcp_set_original_path_prefix() can be called at any time
+ *  (at checkpoint time, within a user's plugin wrapper function, etc.).
+ *  However, this will have no effect until the next restart.
+ *  The user must then set the environment variable DMTCP_NEW_PATH_PREFIX
+ *  prior to restart.  (Note that dmtcp_set_original_path_prefix()
+ *  is not guaranteed thread-safe.)
```
gc00: I propose this modified comment:
```
*  is not guaranteed thread-safe.  Note also that it is not possible to
*  change DMTCP_NEW_PATH_PREFIX after dmtcp_restart is invoked.)
```

plugin/pathvirt/pathvirt.h
```diff
@@ -0,0 +1,27 @@
...
+ *  [ USE CASE:  Only after launch was a new directory assigned
```
gc00: I propose this more verbose version:
```
*  [ POSSIBLE USE CASE:  An application is launched, and then talks
*      to a server to discover what directory to use.  Hence,
*      DMTCP_ORIGINAL_PATH_PREFIX cannot be fully specified until after
*      launch the application. ]
```

plugin/pathvirt/pathvirt.h
```diff
@@ -0,0 +1,27 @@
...
+ *     by some external server. ]
```
gc00: I propose this more verbose version:
```
*  [ POSSIBLE USE CASE:  An application is launched, and then talks
*      to a server to discover what directory to use.  Hence,
*      DMTCP_ORIGINAL_PATH_PREFIX cannot be fully specified until after
*      launch the application. ]
```

gc00:
@rohgarg and @mossberg:  I've proposed some changes to the comments.  Aside from that, it looks good to me.

plugin/pathvirt/pathvirt.cpp
```diff
@@ -0,0 +1,586 @@
...
+        memset(tmpOldPathPrefixList, 0, sizeof(tmpOldPathPrefixList));
+    }
+
+    /* If the user specified the DMTCP_ORIGINAL_PATH_PREFIX on restart,
+     * this will overrided previous calls to set the original buffer.
+     */
+    ret = dmtcp_get_restart_env(ENV_ORIG_DPP, tmp, sizeof(tmp) - 1);
+    errCheckGetRestartEnv(ret);
+    if (ret == RESTART_ENV_SUCCESS) {
+        memset(oldPathPrefixList, 0, sizeof(oldPathPrefixList));
```
mossberg: @rohgarg I don't think this `memset` is necessary because `snprintf` NULL terminates the string for you

plugin/pathvirt/pathvirt.cpp
```diff
@@ -0,0 +1,592 @@
...
+        if (oldEnv) {
+            /* if so, save it to buffer */
+            snprintf(oldPathPrefixList, sizeof(oldPathPrefixList), "%s", oldEnv);
+        }
+        pthread_rwlock_init(&listRwLock, NULL);
+        break;
+    }
+    case DMTCP_EVENT_RESTART:
+    {
+        JTRACE("\n*** The plugin %s is being called after restart. ***");
```
mossberg: I suppose we can delete these now, they were just from a plugin I copied when initially writing this.

plugin/pathvirt/README
mossberg: addressed :+1:

plugin/pathvirt/pathvirt.cpp
rohgarg: @mossberg: I'd still keep it for safety and security. :-)

plugin/pathvirt/pathvirt.h
rohgarg: @gc00: I added RW locks for thread-safety. I'll fix this.

balajir1990:
The interface looks good to me apart from the size limit restrictions. May i know if there is a plan to address it?

mossberg:
@balajir1990 @rohgarg @gc00 At one point, I implemented pathvirt using dynamic allocation so there was no size limit, but we decided against this because I was using malloc functions inside the restart checkpoint. I'm not sure if things are different now, though?

mossberg:
That said, I think the size limits are generous enough for most use cases right now, and we should push it out the door as soon as possible.

karya0:
BTW, you will need to retarget this PR to the 2.5 branch as I have merged the new plugin API code to the master.

balajir1990:
@mossberg @rohgarg I understand that and I am really happy that this plugin just works out of the box .I just wanted to know when we can expect a fix for the size limitation. I dont quite understand what the issue is if we use malloc maybe we can discuss over email

mossberg:
@rohgarg what's the status with this?

kwharrigan:
Has this been tested with `--checkpoint-open-files`?  I'd like to see a test case where files are open during the checkpoint, and checkpoint-open-files saves off copies.  I'd then like to remove all original file locations, and force a rebuild and relocatin from the ckpt_<procstuff>_files copies.  In my testing I wasn't able to make this work.

rohgarg:
@kwharrigan: I do have small test that uses the `--checkpoint-open-files` flag. Please see the target `newtest` in `plugin/pathvirt/test/slot5/Makefile`.

The `newtest` target uses `file-test.c` as the test program. Basically, the test program opens a file for writing in `/tmp`, then at checkpoint time, the file is saved by DMTCP because of the `--checkpoint-open-files` flag, and then finally, we restore the file to a different location on restart because of the pathvirt plugin. Is this what you have in mind?

Note that this needs to be re-targeted for the DMTCP-2.5 branch. The plugin in its current form will not work on the master (or DMTCP-3.0).

kwharrigan:
I see it now.  Thanks.  Yes, that is what I had in mind.  I just wanted to make sure there are, in fact, no dependencies other than `ckpt*.dmtcp` and `ckpt*_files` when running in this prefix mode and when restoring open files.  In this case, I'm able to delete the original `/tmp/msg`, and on restart the old file is restored in the new location, so it appears to work as I had hoped.  More robust testing will commence soon enough. :-)

I was able to rebase of a copy mossberg/pathvirt-v4 onto 2.4.4 for my testing.  I haven't moved to 2.5 yet.

rohgarg:
I'm going to close this PR and open another one targeted for the 2.5 branch.

PR #312, opened 2016-02-22
Fix Makefile for prev commit:  3a53d4f676ed

gc00:
This fixes an earlier commit to provide warnings when using plugin/modify-env


gc00:
@rohgarg has reviewed this, and it's ready to go.

PR #315, opened 2016-02-23
Remove old "DEBUGGING" stmt

gc00:
Remove stray `echo "DEBUGGING"` in` test/plugin/applic-delayed-ckpt`


gc00:
I'm committing this trivial change.  I've labelled it 2.4.5 so that it gets picked up in 2.4.5 and in 2.5.0

Issue #314, opened 2016-02-23
pthread4 test sometimes segfaults under dmtcp_launch


rohgarg:
This affects both DMTCP-3.0 and DMTCP-2.5 on my local setup. Here's my configuration: "openSUSE Tumbleweed", Linux kernel 4.4.0-3, GNU libc 2.22, and GCC 5.3.1.

Here's the backtrace when it fails:
```
pthread4       *** Error in `./test/pthread4': free(): invalid pointer: 0x00007ffbe7b3f008 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x73755)[0x7ffbe63a9755]
/lib64/libc.so.6(+0x79096)[0x7ffbe63af096]
/lib64/libc.so.6(+0x7987e)[0x7ffbe63af87e]
/home/rohgarg/proj/dmtcp/lib/dmtcp/libdmtcp_alloc.so(free+0x66)[0x7ffbe773ae84]
/home/rohgarg/proj/dmtcp/lib/dmtcp/libdmtcp.so(__clone+0xd1)[0x7ffbe6b81975]
/lib64/libpthread.so.0(+0x621a)[0x7ffbe66e121a]
/lib64/libpthread.so.0(pthread_create+0x4ec)[0x7ffbe66e2c6c]
/home/rohgarg/proj/dmtcp/lib/dmtcp/libdmtcp.so(+0x9798c)[0x7ffbe6bbe98c]
/home/rohgarg/proj/dmtcp/lib/dmtcp/libdmtcp.so(pthread_create+0x69)[0x7ffbe6b81ce3]
./test/pthread4[0x40093a]
/home/rohgarg/proj/dmtcp/lib/dmtcp/libdmtcp.so(+0x5abfc)[0x7ffbe6b81bfc]
/lib64/libpthread.so.0(+0x74a4)[0x7ffbe66e24a4]
/home/rohgarg/proj/dmtcp/lib/dmtcp/libdmtcp.so(+0x5a897)[0x7ffbe6b81897]
*** Error in `./test/pthread4': free(): invalid pointer: 0x00007ffbe7b43008 ***
======= Memory map: ========
======= Backtrace: =========
/lib64/libc.so.6(+0x73755)[0x7ffbe63a9755]
00400000-00401000 r-xp 00000000 08:23 6043649                            /home/rohgarg/proj/dmtcp/test/pthread4
[...]

PR #316, opened 2016-02-23
Fix threadwrappers.cpp

rohgarg:
This fixes issue #314.

### DIFF
--- a/src/threadwrappers.cpp
+++ b/src/threadwrappers.cpp
@@ -111,7 +111,7 @@ extern "C" int __clone(int (*fn) (void *arg), void *child_stack, int flags,
   if (tid == -1) {
     JTRACE("Clone call failed")(JASSERT_ERRNO);
     ThreadSync::decrementUninitializedThreadCount();
-    delete thread;
+    JALLOC_HELPER_FREE(thread);
   }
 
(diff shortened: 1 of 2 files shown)


src/threadwrappers.cpp
```diff
@@ -111,7 +111,7 @@ extern "C" int __clone(int (*fn) (void *arg), void *child_stack, int flags,
   if (tid == -1) {
     JTRACE("Clone call failed")(JASSERT_ERRNO);
     ThreadSync::decrementUninitializedThreadCount();
-    delete thread;
+    JALLOC_HELPER_FREE(thread);
```
karya0: Why not change this to `ThreadList::threadIsDead(thread)`? That way we don't have to worry about ThreadList's internal memory management.

gc00: Good catch!  @rohgarg, the issue is that the function for this code had previously called `getNewThread()`, which got a thread from the `threads_freelist`.  We should give it back to the free list.  By calling `threadIsDead()`, we will put it back on the free list.

gc00:
This pull request is also being labelled 2.4.5.  When it's ready, please push it into both the branch 2.5, and the branch 3.0 (master).

Issue #33, opened 2015-01-30
Provide native java bindings to initiate checkpoint from within a Java app


karya0:
We could use Java Native Access (JNA) to quickly prototype it.

gc00:
Hi Kapil,
    I like this idea a lot.  It seems like it's too short for a student
project.  But I do have more than one student now asking about
an MS project or even MS thesis.

Best,
- Gene

Issue #32, opened 2015-01-29
simple fork test doesn't work without the pid plugin


rohgarg:
The following test program fails to restart from the second checkpoint (ckpt-rst-ckpt-rst) when the PID plugin is disabled:

```C
main() {
  pid_t pid = fork();
  if (pid > 0) {
    while (1) {
      printf("Parent\n"); sleep(1);
    }
  } else if (!pid) {
    while (1) {
      printf("Child\n"); sleep(1);
    }
  } else {
    return -1;
  }
  return 0;
}
```

rohgarg:
On the second restart, the process just hangs. The coordinator reports only one of the two processes, which is stuck in the "CHECKPOINTED" phase.

karya0:
Can you verify that the other process is indeed restarting?  Further, can you also verify the pid, ppid, session/group ids before and after checkpoint/restart?

PR #317, opened 2016-02-23
siginfo: Save/restore signal handlers using _real_syscall

rohgarg:
This fixes an issue with programs that use pthread_cancel() reported by
a DMTCP user.

At the time of checkpoint, DMTCP saves all signal handlers using libc's
sigaction() call. Then, DMTCP saves the per-thread signal mask using
libc's pthread_sigmask() call.

On restart, DMTCP first restores the signal handlers of the process
using libc's sigaction() function. And finally, it restores the signal
mask using libc's pthread_sigmask() call.

The NPTL library uses SIGCANCEL (signal #32) and SIGSETXID (signal #33)
internally. The corresponding signal handlers are installed when the
library is initialized by the Linux dynamic linker/loader when loading
a target application that's linked against libc.

The root cause behind the issue is that libc prevents any request that has
anything to do with the two signals from going through to the kernel. In
some cases, libc throws an error, and in some other cases, it simply
ignores the request without any errors. In particular, using sigaction()
to read or write signal handlers for the two signals results in a EINVAL
error, and the call to pthread_sigmask() to read or write the signal
mask for the two signals is simply ignored by libc without any errors.

So, the solution is to not go to libc to read/write the signal handlers,
and to directly make a syscall.

### DIFF
--- a/src/siginfo.cpp
+++ b/src/siginfo.cpp
@@ -109,7 +109,7 @@ void SigInfo::saveSigHandlers()
   /* Now save all the signal handlers */
   JTRACE("saving signal handlers");
   for (sig = SIGRTMAX; sig > 0; --sig) {
-    if (sigaction (sig, NULL, &sigactions[sig]) < 0) {
+    if (_real_syscall(SYS_rt_sigaction, sig, NULL, &sigactions[sig], _NSIG / 8) < 0) {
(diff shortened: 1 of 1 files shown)


gc00:
LGTM.  Please push this into both the 2.5 branch and the master branch (3.0).

karya0:
LGTM.

PR #325, opened 2016-02-25
Fix for #320. Robustify parsing of readelf call used to find the loader

### LINKED ISSUE #320: :q

kwharrigan:
Based on suggestion from @gc00, this appears to resolve #320 for my test case.


kwharrigan:
Sorry, didn't see @gc00  had already put in a pull request.  This would be a duplicate.

Issue #321, opened 2016-02-25
Commit ee391417643d2 (Feb 17, 2016) introduces problem w/ mutex4 test


gc00:
In commit ee391417643d2, one can do the following:
```
make -j6 AUTOTEST=--stress check-mutex4
```
and one will see:
```
mutex4         ckpt:PASSED rstr:PASSED; ckpt:PASSED rstr:PASSED; 
               ckpt:PASSED rstr:FAILED (first process exited: oldstatus 1) retry:PASSED; ckpt:PASSED rstr:PASSED;
```
where restarts will randomly fail.  If one is really unlikely, the second random restart will also fail, and the test will fail.

Prior to this commit, the stress test of mutex4 ran reliably without any restart failures.

@karya0, could you take a look at this?  It happens only on the 3.0 branch (not the 2.5 branch).

karya0:
@gc00: I tried reproducing the bug locally on my laptop as well as on CCIS servers and couldn't. The stress tests went on fine for over 100 lines. Where did you reproduce it? Can you look a little deeper into it and see what is the error message, if any?

gc00:
@karya0, as in our earlier conversation, I can still reproduce the bug when doing a stress test on my office machine.   The bug appears randomly, but almost always within the first five lines.  My office machine is an older (i.e., slower) computer.  And travis is slower.  Maybe the bug appears only on slower machines???

I can now see the bug in much earlier commits on my office machine.
**QUESTION:** If you do the stress test on your machine at commit  a3bdaa1, can you see the bug there?

What follows is a closer diagnosis of where the bug appears.

----

On my office machine, I try the stress test on the 2.5 branch, and the bug does not appear.  It appears only in the master branch.  Yet, I just noticed that the equivalent of commit ee39141 (same commit message) does appear in the 2.5 branch.

So, if this commit appears in master and in 2.5, and if we see the bug only in master, then perhaps something more subtle is going on within master.  Maybe some type of bug appears even earlier in the master branch.

Based on this, I just did some more careful tests.  On the aforementioned commit ee39141, I can see the bug within five lines.  But on earlier commits, I can see the bug less often, but within 20 lines.  Based on this, I just tested branch 2.5 again.  Branch 2.5 usually passes the stress test for 100 lines or more.  But on a rare occasion, I might have seen it fail a stress test.  I don't know how reproducible that is.

[...]

PR #327, opened 2016-02-25
Slurm FT plugin

rohgarg:


gc00:
Nice submission.  (As before, issue #321 is the reason for travis failing.)  I'd like to propose changing the plugin name to `slurm-ckpt-on-kill` or some or some other more meaningful name.  I'd like to get the right balance between short and meaningful.  Suggestions, anyone?

rohgarg:
@gc00: Sure, I don't have a strong opinion either way. I have now also tested the plugin with SLURM, and it seems to work okay. I'll add a sample batch script in the test subdirectory.

gc00:
Then if you like the change in the name for the plugin, then it looks good to me.

MikeDacre:
I am trying to run this myself, but I am getting the error:

    ERROR: ld.so: object '/home/dacre/dmtcp/dmtcp-2.4.4/contrib/slurm-ft/slurm-ft.o' from LD_PRELOAD cannot be preloaded: ignored.

My submission script is here: https://gist.github.com/MikeDacre/c2875bcff4ccae8771a3

gc00:
@rohgarg, could you update the PR to resolve conflicts and then take a look at the comment of @MikeDacre .  This PR seems to have been orphaned, and we should bring it back to life.  Thanks.

gc00:
@rohgarg, I have the same comment as @MikeDacre:  This PR seems to have been orphaned, and we should bring it back to life. Thanks.

rohgarg:
@MikeDacre: I looked at your submission script. The `--with-plugin` takes in a path to the *.so file. Could you please re-try after changing your command to:
  `srun dmtcp_launch --with-plugin /path/to/libdmtcp_slurm-ft.so --checkpoint-open-files python test.py`?

(Your script was referring to the *.o file instead of the *.so file.)

PR #330, opened 2016-02-27
fixup! threadwrappers: Fix segfault when clone call fails

rohgarg:
I accidentally pushed in my changes and closed PR #316.


gc00:
This fixup looks good to me.

PR #337, opened 2016-03-01
dmtcp_coordinator: Fix checkpoint timing when using --exit-on-checkpoint

rohgarg:
This patch fixes an issue with the checkpoint time reported
by the coordinator when using the --exit-on-checkpoint flag.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -422,6 +422,7 @@ void DmtcpCoordinator::updateMinimumState(WorkerState oldState)
 
     if (exitAfterCkpt || exitAfterCkptOnce) {
       JNOTE("Checkpoint Done. Killing all peers.");
+      JTIMER_STOP ( checkpoint );
       broadcastMessage(DMT_KILL_PEER);
       exitAfterCkptOnce = false;
     } else {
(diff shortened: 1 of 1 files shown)


src/dmtcp_coordinator.cpp
```diff
@@ -423,6 +423,7 @@ void DmtcpCoordinator::updateMinimumState(WorkerState oldState)
     if (exitAfterCkpt || exitAfterCkptOnce) {
       JNOTE("Checkpoint Done. Killing all peers.");
       broadcastMessage(DMT_KILL_PEER);
+      JTIMER_STOP ( checkpoint );
```
karya0: Let's move this before `broadcastMessage`. That could take a long time if there are lot of peers.

src/dmtcp_coordinator.cpp
```diff
@@ -467,6 +468,7 @@ void DmtcpCoordinator::updateMinimumState(WorkerState oldState)
...
```
karya0: Same as above

karya0:
LGTM modulo the two minor comments.

src/dmtcp_coordinator.cpp
rohgarg: I followed the convention from earlier, where we'd stop the checkpoint timer after broadcasting messages to the peers. See: https://github.com/dmtcp/dmtcp/blob/2.5/src/dmtcp_coordinator.cpp#L491

gc00: I tend to agree with @karya0.  @rohgarg, the previous case that you mention stops the timer after finishing the resume barrier.  That makes sense to me.  The cost of a checkpoint should include the cost of the corresponding "resume".

However, the cost of a "checkpoint" during `--exit-on-checkpoint` should be just the cost of the `checkpoint`.  It should not include the cost of the `exit` (the cost of a barrier to tell all peers to exit).

rohgarg:
Thanks for the review, @karya0 and @gc00. I have now updated the patch based on your comments.

gc00:
LGTM.

PR #338, opened 2016-03-02
Skip saving text area to ckpt image

karya0:
@gc00 @rohgarg @jiajuncao: Can you take a look at it and see if that works for you?

The optimization is enabled by setting `DMTCP_SKIP_WRITING_TEXT_AREA=1` environment variable.

### DIFF
--- a/src/writeckpt.cpp
+++ b/src/writeckpt.cpp
@@ -30,6 +30,7 @@
 #include <sys/stat.h>
 #include <sys/fcntl.h>
 #include "dmtcp.h"
+#include "constants.h"
 #include "processinfo.h"
 #include "procmapsarea.h"
 #include "procselfmaps.h"
@@ -54,6 +55,7 @@ using namespace dmtcp;
 EXTERNC int dmtcp_infiniband_enabled(void) __attribute__((weak));
 
 static const int END_OF_NSCD_AREAS = -1;
+static bool skipWritingTextSegments = false;
 
 // FIXME:  Why do we create two global variable here?  They should at least
 //         be static (file-private), and preferably local to a function.
@@ -92,6 +94,10 @@ void mtcp_writememoryareas(int fd)
   Area area;
   //DeviceInfo dev_info;
   int stack_was_seen = 0;
+
+  if (getenv(ENV_VAR_SKIP_WRITING_TEXT_SEGMENTS) != NULL) {
+    skipWritingTextSegments = true;
+  }
 
   JTRACE("Performing checkpoint.");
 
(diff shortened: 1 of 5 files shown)


src/writeckpt.cpp
```diff
@@ -433,7 +441,15 @@ static void writememoryarea (int fd, Area *area, int stack_was_seen)
      *   implemented with backing files
      */
     JASSERT((area->flags & MAP_ANONYMOUS) || (area->flags & MAP_SHARED));
-    Util::writeAll(fd, area, sizeof(*area));
-    Util::writeAll(fd, area->addr, area->size);
+
+    if (skipWritingTextArea && (area->prot & PROT_EXEC)) {
+      area->properties = DMTCP_SKIP_WRITING_TEXT_AREA;
+      Util::writeAll(fd, area, sizeof(*area));
+      JTRACE("Skipping over text-area") (area->name) ((void*)area->addr);
```
gc00: text-area -> text segment memory

[ First, avoiding the hyphen; second, the JTRACE should use standard terminology ("text segment memory"), while the reader can quickly translate this into our internal variables with the terminology "area".  If you object to this, my fall-back suggestion is "text area in memory" ]

src/writeckpt.cpp
```diff
@@ -228,7 +234,8 @@ void mtcp_writememoryareas(int fd)
       area.name[0] = '\0';
     } else if (Util::isNscdArea(area)) {
       /* Special Case Handling: nscd is enabled*/
-      area.prot = PROT_READ | PROT_WRITE | MTCP_PROT_ZERO_PAGE;
+      area.prot = PROT_READ | PROT_WRITE;
+      area.properties = DMTCP_ZERO_PAGE;
```
gc00: There's a minor inconsistency here (and throughout this patch).

This version assumes that area.properties is equal to DMTCP_ZERO_PAGE or DMTCP_SKIP_WRITING_TEXT_AREA or 0.  Yet, you defined the macros as 0x0001 and 0x0002, which implies that these are bit masks, and that you can "or" them together to have both properties.  If they are bit properties, then `|=` is the right operator.  Otherwise, the macros should be set to `1` and `2`, rather than `0x0001` and `0x0002`.

In the future, you might switch between bit masks and enumerated types.  But for now, choose one or the other.

gc00:
I have two inline notes.  One of the inline notes is an elaboration on the comment of @rohgarg.

After the inline notes are addressed, it looks good to me.

gc00:
(Also, please check this into both the 2.5 branch and master.  As we discussed, this will be useful in the near future on both branches.)

gc00:
And I have one more comment:  I propose that we change `DMTCP_SKIP_WRITING_TEXT_AREA`
to `DMTCP_SKIP_WRITING_TEXT_SEGMENTS`.  The environment variable is visible to end users and so should use standard terms.  Also, the plural is important, since each library has its own text segment.

PR #341, opened 2016-03-04
dmtcp_coordinator: Fix restart timing

rohgarg:
This patch fixes an issue with the coordinator where
it would fail to record the restart timing information when
configured with --enable-timing.

The issue was due to recent changes in the barrier API. The
new implementation would first check if the coordinator had
released all the requested global barriers, and if not, it would
execute a barrier and increment a counter that keeps track of
the number of executed barriers. However, the implementation
failed to stop and record the restart timer when the global
counter (or the number of executed barriers) became equal to
the number of requested global barriers.


rohgarg:
@karya0, @gc00: Ping!

karya0:
LGTM.

PR #344, opened 2016-03-05
Private barriers

karya0:
The core idea is to provide a un-synchronized barrier which essentially provides an event notification mechanism.


rohgarg:
LGTM.

Issue #346, opened 2016-03-10
--ckpt-open-files in 2.4.4 does not checkpoint or restore unlinked open files correctly


kwharrigan:
Consider the example in under tests in: https://github.com/kwharrigan/dmtcp/tree/unlink_test

It is common w/ temporary files to open and immediately unlink them, so that they are deleted once the process exits.  I'm concerned this is not being handled quite right, or perhaps I'm expecting some behavior that is not designed.

For one, the unlinked file is not stored in the ckpt*_files folder.  This means, at minimum, that in a restarted process the file contents are incorrect (or non-existent?).  

In my tests, if I pause right after the launch and  checkpoint, and inspect the /proc/<pid>/fd entry, I see the deleted file entry.  I can, of course, cat the contents of the deleted entry while the process is still running (0,1,2,3,4 as expected).

However, upon restart, the file is not restored to the unlinked / deleted (yet still open) state.  There is a new file handle open, which is empty.  I can only assume downstream processes expected to read the contents of that file will see incorrect file contents, and thus the state of the process has not been recreated faithfully.  **EDIT: The previous is kind of a silly statement.  In case of `O_WRONLY`, obviously we won't be reading this file later.  However, there is still perhaps an edge case, and clearly the argument we haven't restored the process accurately to the checkpointed state.  ** 

Hopefully this all makes sense.  I'm happy to dig into it more but thought I'd report so we can work it in parallel.

rohgarg:
@kwharrigan: See: https://github.com/dmtcp/dmtcp/blob/2.5/src/plugin/ipc/file/fileconnection.cpp#L580 
The if-clause prevents an unlinked file that was opened with `O_WRONLY` flag from being checkpointed. Other unlinked files are checkpointed by default, regardless of the `--checkpoint-open-files` flag (see line #588 in `fileconnection.cpp`). The if-clause was introduced in commit: 5c3dd0aa, and seems to be intentional, but I am not sure if I understand the reason behind it. 

(FYI: the file1 test  (`test/file1.c`) in the DMTCP test-suite does the same thing as your unlink_test.)

@karya0: do you happen to remember the rationale behind this?

rohgarg:
@kwharrigan: After discussing with @karya0, I figured out problematic lines in the code. I'll create a patch for this soon.

rohgarg:
Fixed via #349.

PR #349, opened 2016-03-14
fileconnection: Fix checkpointing of unlinked files opened with O_WRONLY

rohgarg:
This patch fixes an issue where DMTCP would incorrectly restore
files that were unlinked after opening with O_WRONLY.

The policy for checkpoint-restore of such files was first defined
in the commit 5c3dd0a as follows:

 a) DMTCP should not checkpoint such files, regardless of the
    --checkpoint-open-files flags; and
 b) On restart, DMTCP should open such files with (O_WRONLY|O_CREAT)
    and unlink them.

However, the implementaiton never achieved these goals. We did end up
doing the first part correctly, but never really got the second path right.

With the above commit, files would be opened with (O_WRONLY|O_CREAT)
on restart but were never unlinked because the commit overlooked an
if-clause in the `FileConnection::resume()` method. The if-clause would
call unlink() only if a file had the `_ckpted_file` flag set to true.

Later, the commit: 8514737 worsened the situation. It introduced a bug
in the `FileConnection::refill()` method that would prevent us from even
opening write-only, unlinked files on restart.

### DIFF
--- a/src/plugin/ipc/file/fileconnection.cpp
+++ b/src/plugin/ipc/file/fileconnection.cpp
@@ -264,7 +264,7 @@ void FileConnection::refill(bool isRestart)
 
   if (!_ckpted_file) {
     int tempfd;
-    if (_type == FILE_DELETED && (_flags & (O_WRONLY | O_RDWR))) {
+    if (_type == FILE_DELETED && ((_flags & O_WRONLY) || (_flags & O_RDWR))) {
       tempfd = _real_open(_path.c_str(), _fcntlFlags | O_CREAT, 0600);
       JASSERT(tempfd != -1) (_path) (JASSERT_ERRNO) .Text("open() failed");
       JASSERT(truncate(_path.c_str(), _st_size) ==  0)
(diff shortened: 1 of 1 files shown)


rohgarg:
@kwharrigan: Could you please try this patch and see if it fixes the issue #346 for you? I have this targeted for the master branch now, but this will be also be pushed to the 2.5 branch.

rohgarg:
@karya0: Could you please review this change?

kwharrigan:
Thanks for the update. I will have a look but it may be a few days.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -261,7 +261,7 @@ void FileConnection::refill(bool isRestart)
 
   if (!_ckpted_file) {
     int tempfd;
-    if (_type == FILE_DELETED && (_flags & (O_WRONLY | O_RDWR))) {
+    if (_type == FILE_DELETED && ((_flags & O_WRONLY) || (_flags & O_RDWR))) {
```
karya0: Let's remove `(_flags & O_RDWR)` until we have the full context and commit it. Please create an issue to address 8514737.

PR #350, opened 2016-03-16
pid_miscwrappers: Fix CMA wrappers

rohgarg:
This patch adds a EXTERNC to correctly compile the CMA wrappers
as C functions and also adds cases for CMA system calls in the
syscall() wrapper.


rohgarg:
This should also be pushed to 2.5.

karya0:
LGTM.

PR #351, opened 2016-03-16
pid-plugin: Fix incorrect initialization of pid map

rohgarg:


rohgarg:
This should also be pushed to 2.5.

karya0:
LGTM. Can you add a comment explaining the context.

rohgarg:
I have now added a description of the issue in the commit.

PR #353, opened 2016-03-17
pathvirt-plugin: Fix pathvirt not enabled for exec post-restart

rohgarg:


rohgarg:
@ankitcse07: I'll push this to 2.5 if/when you can confirm this works for you.

@mossberg, @gc00: Could you please review this?

ankitcse07:
Test provided to you is working fine, but somehow in my system there seems to be some issue.  I need to first isolate the issue and then get back to you.

rohgarg:
@ankitcse07: Okay, let me know what you find.

ankitcse07:
@rohgarg I found one more issue,  DMTCP_NEW_PATH_PREFIX environment variable is not available to processes launched on remote host during restart, so remote host process are not able to do path virtualization.

rohgarg:
@ankitcse07: Could you explain how you invoke the restart process on a remote host?

ankitcse07:
What I mean restart process fork a new process on remote machine via SSH/RSH and this new process is not undergoing path virtualization

rohgarg:
That's correct. Environment variables from a local host are not propagated to a remote host over ssh by default. (I don't know about rsh.) One possibility is to use the `-o SendEnv= DMTCP_NEW_PATH_PREFIX ` ssh option, but that'll work only if the remote's sshd is configured to permit environment variables over ssh (see `PermitUserEnvironment` in sshd config). For a more general fix, it seems like we'll need a new flag for dmtcp_restart.

ankitcse07:
Why do we need to have switch, should n't this be automatically done for all child processes irrespective of the fact they are local or remote processes

rohgarg:
I'm assuming you restart your program as follows:

`ssh remote-host " dmtcp_restart ckpt_a.out.dmtcp"`

If that's correct, then as I mentioned before, environment variables will not be propagated over ssh. You need to either:
  - use "ssh -o SendEnv" and modify the sshd config on the remote host to allow propagation of env. vars.; or
  - add a dmtcp_restart flag to handle new-path-prefix.

ankitcse07:
This is not what I meant, restart script is launched on the local host setting env variable for the new mapping, but then one of the restarted process launches a new process on remote machine. This new process on remote machine is not performing path virtualization

rohgarg:
How is the new program exec-ed on the remote host? Is it via ssh? Then it will have the same issue as I mentioned above.

ankitcse07:
Yes it is via SSH, can't this be handled in SSH plugin  which automatically sets this variable on remote host.

rohgarg:
Possibly, but I haven't looked at it very carefully to be sure. Either way, supporting this requires modifications to dmtcp_restart and/or the ssh plugin.

ankitcse07:
Ok let me see the code of ssh, I am now little familiar and possibly propose a new PR

rohgarg:
You wrote:

>Test provided to you is working fine, but somehow in my system there seems to be some issue. I need to first isolate the issue and then get back to you.

Is this still valid? If not, I'd like to close this PR and track the ssh issue separately.

ankitcse07:
With this change, new process is performing path virtualization on frest run itself  in case DMTCP_NEW_PATH_PREFIX and DMTCP_ORIGINAL_PATH_PREFIX are both set, is that expected ?

ankitcse07:
[pathvirt_issue1 (1).tar.gz](https://github.com/dmtcp/dmtcp/files/248669/pathvirt_issue1.1.tar.gz)

See the attached test and just invoke run, you will see that file restart_process.txt will be created in the newdir itself

rohgarg:
Yes, a new process that's forked off of a restarted process will see the new paths.

ankitcse07:
I am saying this is not a restarted process but the first run in which checkpoint was taken, new process was forked and even this new process is seeing new paths is this expected ?

rohgarg:
Unless a user sets both the path prefixes -- original and new -- it shouldn't be happening. The current implementation does not have a safety check (which it probably should) to preclude a child process forked prior to the first checkpoint from referring to the new paths.

ankitcse07:
Yes both are set but this means all child process will have this property but not the parent process

rohgarg:
You wrote:

> Yes both are set ...

Why? Is this a new requirement or merely for testing? If it's just for testing, I'll probably add this as a caveat in the documentation, and add a JASSERT to catch this. However, if there is a new requirement, I'd say it's not relevant to the current pull request, and that it should be tracked separately. 

Anyway, this discussion has left me wondering if there is a more flexible interface that will allow a user to turn pathvirt on and off based on their needs.

ankitcse07:
Just it was left in shell settings and I relauched the test again, its not a requirement but a assert will be helpful if user is doing something wrong

rohgarg:
Okay, then we have converged. Thanks for pointing this out! I'll add a check in the code, update the documentation, and push it upstream. :-)

Please feel free to open a new issue to track the progress on propagation of env. vars. over ssh. That still needs to be fixed, but it's a separate concern, in my opinion.

ankitcse07:
I had already done fix for SSH in SSH plugin folder itself and it's working fine now if you want I want create a PR with the commit

PR #357, opened 2016-03-30
Fix case when mtcp_sys_brk() causes munmap()

gc00:
In `mtcp_restart.c`, the function `mtcp_sys_brk()` can be called to set an earlier end-of-data.  In this case, the `brk()` kernel call will munmap the leftover memory.  But the `rinfo` struct can be stored near the end-of-data, which is then munmap'ed inside `restore_brk()`.  This fixes it by delaying the `brk()` call in this case, until later in `restorememoryareas()`, after we're done using the global struct, `rinfo`.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -333,7 +333,15 @@ static void restore_brk(VA saved_brk, VA restore_begin, VA restore_end)
     mtcp_abort ();
   }
 
-  new_brk = mtcp_sys_brk (saved_brk);
+  if (current_brk <= saved_brk) {
+    new_brk = mtcp_sys_brk (saved_brk);
+    rinfo.saved_brk = NULL; // We no longer need the value of saved_brk.
+  } else {
+    new_brk = saved_brk;
+    // If saved_brk < current_brk, then brk() does munmap; we can lose rinfo.
+    // So, keep the value rinfo.saved_brk, and call mtcp_sys_brk() later.
+    return;
+  }
   if (new_brk == (VA)-1) {
     MTCP_PRINTF("sbrk(%p): errno: %d (bad heap)\n",
 		 saved_brk, mtcp_sys_errno );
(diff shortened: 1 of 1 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -302,7 +303,8 @@ MTCP_PRINTF("Attach for debugging.");
 }
 
 NO_OPTIMIZE
-static void restore_brk(VA saved_brk, VA restore_begin, VA restore_end)
+static void restore_brk(VA saved_brk, VA restore_begin, VA restore_end,
+                        RestoreInfo * rinfo_ptr)
```
rohgarg: It seems like we never use this new extra argument -- `rinfo_ptr` -- in the `restore_brk()` function. It's aliased to the global `rinfo` object, anyway. I think we should remove the extra argument.

rohgarg:
LGTM modulo one minor issue (see my inline comment) and some testing. Also, I think this should be pushed to both the master branch and the 2.5 branch.

src/mtcp/mtcp_restart.c
```diff
@@ -333,7 +335,14 @@ static void restore_brk(VA saved_brk, VA restore_begin, VA restore_end)
     mtcp_abort ();
   }
 
-  new_brk = mtcp_sys_brk (saved_brk);
+  if (current_brk <= saved_brk) {
+    new_brk = mtcp_sys_brk (saved_brk);
+    rinfo.saved_brk = NULL; // We no longer need the value of saved_brk.
+  } else {
+    new_brk = saved_brk;
```
karya0: Why not explicitly return from here?

src/mtcp/mtcp_restart.c
```diff
@@ -552,6 +561,11 @@ static void restorememoryareas(RestoreInfo *rinfo_ptr)
 
   RestoreInfo restore_info;
   mtcp_memcpy(&restore_info, rinfo_ptr, sizeof (restore_info));
+  if (rinfo_ptr->saved_brk != NULL); {
```
karya0: I am wondering if we should call mtcp_sys_brk() from here only. Why should we even bother about calling it before rinfo has been relocated?

gc00: Basically, I wanted to make the minimal change possible.  I didn't want to do a new analysis of the code for catching errors in `restore_brk()`.

src/mtcp/mtcp_restart.c
gc00: Good point.  Done in the next version.

PR #360, opened 2016-04-15
Enhancing Ckptfile plugin with overwrite functionality for ckpt-files

ankitcse07:
Ckptfile plugin allows user to checkpoint its own files. Now its quite possible
that these checkpointed files have changed prior to restart. Currently this will
result in error saying unequal files. However this scenario cannot be avoided
for log files which will always change.

This commit allows user to allow pre-existing files to be overwritten with the
checkpointed copy at restart rather than resulting in error.  Simple text
file used as database has been enhanced to point which files are allowed
to be overwritten. User can provide SAVE_AND_RESTORE=2 to point file will
be checkpointed and allowed to be overwritten at restart time.

This also targets   ckptfile-plugin: make it useful #9

### DIFF
--- a/src/plugin/ipc/file/fileconnection.cpp
+++ b/src/plugin/ipc/file/fileconnection.cpp
@@ -649,7 +649,13 @@ void FileConnection::refill(bool isRestart)
   if (strstr(_path.c_str(), "infiniband/uverbs") ||
       strstr(_path.c_str(), "uverbs-event")) return;
 
-  if (_ckpted_file && _fileAlreadyExists) {
+  bool overwrite_file = false;
+  if(dmtcp_overwrite_ckptfile_on_restart && 
+      dmtcp_overwrite_ckptfile_on_restart(_path.c_str())) {
+    overwrite_file = true;
+  }
+
+  if (_ckpted_file && _fileAlreadyExists && !overwrite_file) {
     string savedFilePath = getSavedFilePath(_path);
     int savedFd = _real_open(savedFilePath.c_str(), O_RDONLY, 0);
     JASSERT(savedFd != -1) (JASSERT_ERRNO) (savedFilePath);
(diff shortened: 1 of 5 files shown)


ankitcse07:
@rohgarg What do you think for this new API.

rohgarg:
@ankitcse07: Would it make sense to add a `dmtcp_restart` flag to allow a user to force overwriting of existing files with their saved versions? This is probably a less flexible approach than what you are proposing but keeps the interface cleaner. However, if you feel that there's a strong use case for allowing selective overwrites, I'm okay with what you have (except for the code formatting :-) ).

ankitcse07:
I won't do that as my plugin has no control over user files which are checkpointed, so I don't really want to force all files to overwrite including user files and really want to have control over which files to overwrite. For my case I would only touch those files which are created by my software stack.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -702,15 +708,17 @@ void FileConnection::refill(bool isRestart)
...
-  if (jalib::Filesystem::FileExists(_path) &&
-      stat(_path.c_str() ,&statbuf) == 0 && S_ISREG(statbuf.st_mode)) {
-    if (_offset <= statbuf.st_size && _offset <= _st_size) {
-      JASSERT(lseek(_fds[0], _offset, SEEK_SET) == _offset)
-        (_path) (_offset) (JASSERT_ERRNO);
-      //JTRACE("lseek(_fds[0], _offset, SEEK_SET)") (_fds[0]) (_offset);
-    } else if (_offset > statbuf.st_size || _offset > _st_size) {
-      JWARNING(false) (_path) (_offset) (_st_size) (statbuf.st_size)
-        .Text("No lseek done:  offset is larger than min of old and new size.");
+  if(!_ckpted_file || (_ckpted_file && !overwrite_file)) {
```
rohgarg: Two issues:
 - No need for such a complex expression; the expression you have is equivalent to: `!(ckpted_file && overwrite_file)`
 - Unless I'm reading this incorrectly, this new expression disallows seeking to the original offset when a file was ckpt-ed and overwritten by its ckpt-ed copy. I don't think you want that.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -821,6 +829,20 @@ void FileConnection::postRestart()
 
     if (fd == -1) {
       _fileAlreadyExists = true;
+      bool overwrite_file = false;
+      if(dmtcp_overwrite_ckptfile_on_restart &&
+          dmtcp_overwrite_ckptfile_on_restart(_path.c_str())) {
+        overwrite_file = true;
+      }
+      if(overwrite_file) {
```
rohgarg: This is probably not the best place to overwrite a file with its ckpt-ed copy; you need to defer it to the `refill()` method. See the comment above the `_real_open()` call.

The only guarantee provided in a distributed system with barriers is that a process will wait until all its peers reach the same barrier, but it is free to process its callbacks for the barrier independent of the others. And so, one reason for the open() call returning a -1 is simply that some other restarting process succeeded in creating the file before the current process got to this point, and hence, it's possible that there's data in flight.

rohgarg:
Okay, I looked at your patch a bit more closely. I have added some comments/observations.

FYI:
I am preparing a patch to allow DMTCP to overwrite a file with its ckpt-ed version. This will be similar to the current `--ckpt-open-files` option. I'll put it up for review soon. Your patch will become much simpler once we have that.

ankitcse07:
@rohgarg I can see that your PR for overwriting ckpt files is doing same thing but for all files, can this be enhanced to choose only specific files. So can you create an API which takes filename as argument and this function can then be overridden by plugin to choose specific files, similar to what I did in this PR

rohgarg:
@ankitcse07 I just update my PR. Could you please check that and see if it makes sense?

ankitcse07:
@rohgarg yes selective API is what I want so that if plugin requires to chose specific files to overwrite plugin writer can do so.  Thanks, I will merge this in my sandbox and see if this covers all of my scenarios. I will get back to you.

ankitcse07:
Thanks @rohgarg I am closing this PR

PR #235, opened 2015-11-20
Issue is occuring using --checkpoint-open-files for ckpting open files

ankitcse07:
Consider a scenario in which file was opened for writing and however
its got deleted in between prior to taking dmtcp checkpoint. fds are
still opened for process but file is actually deleted. This is not
handled and dmtcp checkpoint exit as it tries to checkpoint the deleted
file. This was handled in previous versions within 
handleUnlinkedFile function of FileConnection class but code was cleaned.

See the attached files for the test case reproducing this issue.

[error_log.txt](https://github.com/dmtcp/dmtcp/files/40030/error_log.txt)
[run.csh.txt](https://github.com/dmtcp/dmtcp/files/40032/run.csh.txt)
[test.cxx.txt](https://github.com/dmtcp/dmtcp/files/40031/test.cxx.txt)

PR #361, opened 2016-04-15
Issue is occuring using --checkpoint-open-files for ckpting open files

ankitcse07:
Consider a scenario in which file was opened for writing and however
its got deleted in between prior to taking dmtcp checkpoint. fds are
still opened for process but file is actually deleted. This is not
handled and dmtcp checkpoint exit as it tries to checkpoint the deleted
file. This was handled in previous within handleUnlinked file function
but code was cleaned.


ankitcse07:
Creating a new pull request for #235 to push in master branch

PR #362, opened 2016-04-18
fileconnection: Fix preCkpt() for files opened using fopen-family

rohgarg:
For files opened using the fopen* family of functions, the `_flags` field
of the `FileConnection` object is initialized to -1. The field is initialized
when the object is created at the time  of opening the file and is never
set to the correct value.

This patch fixes an issue caused because the `preCkpt()` method
incorrectly referred to the `_flags` field to determine how to save a
copy of an opened file (when using the `--checkpoint-open-files` option).

This fix is to use the `_fcntlFlags` field instead of the `_flags` field
in the `preCkpt()` method. The `_fnctlFlags` field is set to the value
returned by `fcntl(F_GETFL)` in the `drain()` method, which gets called
before `preCkpt()`.


ankitcse07:
Thanks @rohgarg, bug got fixed. Let me know when you will merge this in 2.5 branch

karya0:
LGTM.

PR #364, opened 2016-04-21
Add support for logging coordinator messages when running in daemon mode

rohgarg:
This patch adds a `--coord-logfile <filename>` option to dmtcp_coordinator,
dmtcp_launch, dmtcp_restart, and the restart script. This option can be
used to specify a file where the coordinator can dump its log messages
when running in daemon mode. This is useful for debugging.


jiajuncao:
LGTM.

Issue #365, opened 2016-04-22
DMTCP masks the real errno in many cases


rohgarg:
I suspect that this affects many wrappers. I have only looked at the socket wrappers for now -- `connect()`,` bind()`, and `listen()`.

The problem is the code in a wrapper between the call to its _real_ function and the return statement, for example, JTRACE's, etc., can have a side-effect of modifying the errno. As a result, the wrapper can end up returning an incorrect errno to the user. The fix is to save the errno once a _real_ call returns and restore it just before returning.

I can create a pull request to fix some of the wrappers in `socketwrappers.cpp`, but we need to be careful when writing any wrapper functions. I wonder if there's a more general fix that'll automate this and minimize human error, perhaps by a clever modification of the `NEXT_FNC` macro.

rohgarg:
After discussions with @karya0, it seems like we most of the wrappers do correctly save and restore the errno, and the problem was specifically with the `connect()` wrapper. However, @karya0 did observe that because the JTRACE's aren't correctly saving and restoring errno, they could end up masking the errno for the actual wrapper. I'm modifying this issue to track this.

PR #366, opened 2016-04-22
socketwrappers: Fix masking of real errno

rohgarg:
This patch is an attempt to fix the issue of masking of the _real_
errno by DMTCP's socket wrappers.

The problem is that the code in a wrapper between the call to the real
function and the return statement, for example, JTRACE's, etc., can have
a side-effect of modifying the errno. As a result, the wrapper can end
up returning an incorrect errno to the user. The solution is to save
the errno once a real call returns and restore it just before returning.

Although this patch only addresses the issue for socketwrappers, the
problem is more widespread (see issue #365); I'm putting this out for
early feedback.

### DIFF
--- a/src/plugin/ipc/socket/socketwrappers.cpp
+++ b/src/plugin/ipc/socket/socketwrappers.cpp
@@ -76,6 +76,7 @@ extern "C" int connect(int sockfd, const struct sockaddr *serv_addr,
   DMTCP_PLUGIN_DISABLE_CKPT(); // The lock is released inside the macro.
 
   int ret = _real_connect(sockfd,serv_addr,addrlen);
+  int savedErrno = errno; // Save errno to prevent modifications by the following code
   if ((ret != -1 || errno == EINPROGRESS) &&
       dmtcp_is_running_state() &&
       !_doNotProcessSockets) {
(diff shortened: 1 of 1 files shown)


rohgarg:
@ankitcse07: Could you please check this patch and see if it fixes the issue for you? This will be pushed to both the master branch and the 2.5 branch.

rohgarg:
@karya0, @gc00: Could you please review this? Ideally, I would have liked to integrate the `AutoSaveErrno` class with the `NEXT_FNC` macro so that the current wrappers remains unaffected, but there seems to be no easy way of doing it without changing the structure of the macro. Having said that, there is a flaw with an integrated approach: a wrapper might have calls to multiple `_real_*` functions, in which case, the integrated approach could lead the wrapper to report an incorrect `errno`. So, I'm kind of happy that the current non-integrated approach delegates the responsibility of deciding the right place to save the errno to the wrapper author.

rohgarg:
After discussions with @karya0, it seems like we most of the wrappers do correctly save and restore the errno, and the problem was specifically with the `connect()` wrapper. So, I have updated the patch to a localized change in the `connect()` wrapper. However, @karya0 did observe that because the JTRACE's aren't correctly saving and restoring errno, they could end up masking the errno for the actual wrapper. I'm opening a separate issue to track this.

rohgarg:
@ankitcse07: Did you get a chance to test this?

ankitcse07:
@rohgarg I will let you know by today,  I will pull this branch and merge in my sandbox

PR #371, opened 2016-05-01
dmtcp_launch: Add option to allow overwriting existing file w/ ckpt-ed version

rohgarg:
This patch adds a new `dmtcp_launch` option, `--allow-file-overwrite`,
to allow overwriting of an existing file with its checkpointed version.


src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -251,12 +261,24 @@ void FileConnection::refill(bool isRestart)
                 "   Change this function and re-compile, if you want "
                 "different behavior.");
       } else {
-        const char *errMsg =
+        char *errMsg =
```
ankitcse07: @rohgarg My restart is failing with the assert failure in while calling from above code.  I guess your check for file overwrite should be moved prior to making areFilesEqual call 

static bool areFilesEqual(int fd, int savedFd, size_t size)
{
  long page_size = sysconf(_SC_PAGESIZE);
  const size_t bufSize = 1024 * page_size;
  char *buf1 =(char*)JALLOC_HELPER_MALLOC(bufSize);
  char *buf2 =(char*)JALLOC_HELPER_MALLOC(bufSize);

  off_t offset1 = _real_lseek(fd, 0, SEEK_CUR);
  off_t offset2 = _real_lseek(savedFd, 0, SEEK_CUR);
  JASSERT(_real_lseek(fd, 0, SEEK_SET) == 0) (fd) (JASSERT_ERRNO);
  JASSERT(_real_lseek(savedFd, 0, SEEK_SET) == 0) (savedFd) (JASSERT_ERRNO);

  int readBytes;
  while (size > 0) {
    readBytes = Util::readAll(savedFd, buf1, MIN(bufSize, size));
    JASSERT(readBytes != -1) (JASSERT_ERRNO) .Text("Read Failed");
    if (readBytes == 0) break;
**    JASSERT(Util::readAll(fd, buf2, readBytes) == readBytes);
**    if (memcmp(buf1, buf2, readBytes) != 0) {
      break;
    }
    size -= readBytes;
  }

rohgarg:
@ankitcse07: This issue with `areFilesEqual()` has been fixed upstream. Please see the commit: f0dba2f0a504794b699c502194edb51931f76ca7.

ankitcse07:
@rohgarg Now I got following error 
      

[237000] ERROR at fileconnection.cpp:635 in preCkpt; REASON='JASSERT(tmpfd != -1) failed'
     _path.c_str() = /vel/AGARG/demos/dmtcp_demos/c_testbench/dmtcp_veloce_examples
/hello_world_dmtcp_2apis_cm_master/veloce.log/dmslogdir/EnterpriseClient.log.bak2 (deleted)


void FileConnection::preCkpt()
{
  if (_ckpted_file) {
    ConnectionIdentifier id;
    JASSERT(_type != FILE_PROCFS && _type != FILE_INVALID);
    JASSERT(SharedData::getCkptLeaderForFile(_st_dev, _st_ino, &id));
    if (id == _id) {
      string savedFilePath = getSavedFilePath(_path);
      JASSERT(Util::createDirectoryTree(savedFilePath)) (savedFilePath)
        .Text("Unable to create directory in File Path");

      int destFd = _real_open(savedFilePath.c_str(), O_CREAT | O_WRONLY | O_TRUNC,
                              S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
      JASSERT(destFd != -1) (JASSERT_ERRNO) (_path) (savedFilePath);

      JTRACE("Saving checkpointed copy of the file") (_path) (savedFilePath);
      if (_fcntlFlags & O_WRONLY) {
        // If the file is opened() in write-only mode. Open it in readonly mode
        // to create the ckpt copy.
        int tmpfd = _real_open(_path.c_str(), O_RDONLY, 0);
**        JASSERT(tmpfd != -1)(_path.c_str());**

rohgarg:
I think this is unrelated to the current patch. Are you using the ckptfiles plugin? If yes, then I think I know the bug. It's in the `FileConnection::drain()` method.

ankitcse07:
@rohgarg I am using code similar to ckptfiles plugin in my own written plugin which is giving definition of dmtcp_must_ckpt_file to choose which files to checkpoint. If you want to debug I can provide a webex session for the same

rohgarg:
That's not required, I know the bug. Can you confirm that you can reproduce this issue without this patch?

rohgarg:
Anyway, if trying to reproduce the issue without the patch is difficult, you will need to apply the following two patches:

a)
```diff
diff --git a/src/plugin/ipc/file/fileconnection.cpp b/src/plugin/ipc/file/fileconnection.cpp
index 0b73572..815d81c 100644
--- a/src/plugin/ipc/file/fileconnection.cpp
+++ b/src/plugin/ipc/file/fileconnection.cpp
@@ -575,14 +575,15 @@ void FileConnection::drain()
     return;
   }
 
-  if (dmtcp_must_ckpt_file && dmtcp_must_ckpt_file(_path.c_str())) {
-    _ckpted_file = true;
+  if (_type == FILE_DELETED && (_flags & O_WRONLY)) {
     return;
   }
 
-  if (_type == FILE_DELETED && (_flags & O_WRONLY)) {
+  if (dmtcp_must_ckpt_file && dmtcp_must_ckpt_file(_path.c_str())) {
+    _ckpted_file = true;
     return;
   }
+
   if (_isBlacklistedFile(_path)) {
     return;
   }
```


b) PR #349

ankitcse07:
This is anyways occuring at the time of checkpoint so this patch anyways has no role here

ankitcse07:
I applied above patches, now checkpoint went through but at restart following error is flashed

[117000] ERROR at fileconnection.cpp:913 in areFilesEqual; REASON='JASSERT(_real_lseek(fd, 0, SEEK_SET) == 0) failed'
     fd = 21
     (strerror((*__errno_location ()))) = Illegal seek


904 static bool areFilesEqual(int fd, int savedFd, size_t size)
 905 {
 906   long page_size = sysconf(_SC_PAGESIZE);
 907   const size_t bufSize = 1024 * page_size;
 908   char *buf1 =(char*)JALLOC_HELPER_MALLOC(bufSize);
 909   char *buf2 =(char*)JALLOC_HELPER_MALLOC(bufSize);
 910
 911   off_t offset1 = _real_lseek(fd, 0, SEEK_CUR);
 912   off_t offset2 = _real_lseek(savedFd, 0, SEEK_CUR);
 913   JASSERT(_real_lseek(fd, 0, SEEK_SET) == 0) (fd) (JASSERT_ERRNO);
 914   JASSERT(_real_lseek(savedFd, 0, SEEK_SET) == 0) (savedFd) (JASSERT_ERRNO);
 915
 916   int readBytes;

ankitcse07:
For your information I added some logs, this is one of the files required to be overwritten so anyways checking for file equality should not be done for overwritten files

rohgarg:
I was looking through your error messages from earlier, and it seems like your intent was to set ckpt-ing to true for a file ("EnterpriseClient.log.bak2") that you knew your application might delete or had already deleted after opening.

Note that the first of the two patches I suggested earlier as a fix for your issue is some thing that's definitely required for DMTCP, but it'll break what you really intended to do (assuming that was your real intent and not an accident).

ankitcse07:
@rohgarg This file was checkpointed based on regex expression added by me on file paths which decided it should be checkpointed in my plugin, however application is free to do anything on files including deletion so yes my intent was to checkpoint file, do you know how this can be fixed ?

rohgarg:
DMTCP already handles ckpt-ing of deleted files that were opened with O_WRONLY and unlinked without closing the file-descriptor. The policy for handling such files is that:

 - DMTCP will not checkpoint such files, regardless of the
   `--checkpoint-open-files` flag; and
 - on restart, DMTCP will open such files with (O_WRONLY|O_CREAT)
   and unlink them.

This is done to present a consistent view of the environment to the application.

We're happy to discuss any proposal for changes/extensions to this policy, but I think it is not pertinent to the current patch. So, unless we have any other major roadblocks, I'd like to push this patch upstream.

The "Illegal seek" error (ESPIPE) you are getting is odd, but I think comparing the file contents (checkpointed vs. existing) is still relevant. The comparison allows us to save time on copying data from the checkpointed version to the existing version when the contents match. Anyway, the error message suggests that the fd we are seeking on is not a regular file. This shouldn't be happening. Can you print the file path in the `FileConnection::refill()` method? (Try printing `_path` before the call to `areFilesEqual()`.)

ankitcse07:
I did that, however I can see this is a regular file but still its saying illegal seek 

130000] ERROR at fileconnection.cpp:913 in areFilesEqual; REASON='JASSERT(_real_lseek(fd, 0, SEEK_SET) == 0) failed'
     fd = 20
     (strerror((*__errno_location ()))) = Illegal seek
     path = /vel/AGARG/demos/dmtcp_demos/c_testbench/dmtcp_veloce_examples/hello_world_dmtcp_2apis_cm_master/veloce.log/veloce_runtime.log/TRACE_RTS.txt
runtimeserver (130000): Terminating...

rohgarg:
Interesting! Let's track/discuss this separately. I don't think it's related to the current patch. I'm guessing you can reproduce this issue even without the current patch.

ankitcse07:
I guess issue is _fds[0] returned 21 as file descriptor in restart, however if I look at current /proc/pid/fd/21 it is pointing to some socket rather than opened file which was originally checkpointed. So lseek on this file descriptor fails in areFilesEqual. 


[agarg@abraracourcix ~]$ ls -l /proc/23654/fd/
lrwx------ 1 agarg medrd 64 May  4 19:55 21 -> socket:[74674637]

[agarg@abraracourcix ~]$ fg
/tools/linux64/gdb-7.5.1-suse10/bin/gdb /distrib/Diamond_RD/qa/v30111-1/software//amd64.linux.gnu.debug/runtimeserver 23654

$8 = (int &) @0x7f01334834c0: 21
(gdb) p _fds[0]
$9 = (int &) @0x7f01334834c0: 21
(gdb) p _path
$10 = {static npos = 18446744073709551615, _M_dataplus = {<dmtcp::DmtcpAlloc<char>> = {<No data fields>},
    _M_p = 0x7f0133478418 "/vel/AGARG/demos/dmtcp_demos/c_testbench/dmtcp_veloce_examples/hello_world_dmtcp_2apis_cm_master/veloce.log/veloce_runtime.log/TRACE_RTS.txt"}}

rohgarg:
Is this socket opened/managed by your plugin? If yes, then you need to make sure that it's a _protected fd_. In general, any file descriptors that you want to be managed (checkpointed and restored) by your plugin need to be _protected_ to ensure that there's no overlap with other plugins (and the file plugin, in particular).

Here's a theory that might explain what's going on: At checkpoint time, the fd 21 was mapped to a file. On restart, the file plugin remembers this information, restores the fd, and tries to treat it as a file. However, your plugin (or perhaps some other DMTCP plugin) maps the fd 21 back to a socket at a later stage. As a result, the call to lseek fails in `FileConnection::refill()`. Could you try to track what's going on with this fd 21 prior to checkpoint and after restart? The goal is to figure out if and when the fd's mapping changes from a file to a socket.

I'm opening a separate issue to track this bug.

ankitcse07:
@rohgarg When is the plan to commit this in upstream, with the all fixes in place, this integration is working fine with my plugin.  I had overridden the api dmtcp_must_overwrite_file in my plugin and its solving the purpose of selective overwrite. Thanks for all the help

src/plugin/ipc/file/fileconnection.cpp
karya0: Agreed! If files are to be overwritten, we shouldn't compare them.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -252,12 +262,24 @@ void FileConnection::refill(bool isRestart)
...
           "\n**** File already exists! Checkpointed copy can't be restored.\n"
           "       The Contents of checkpointed copy differ from the "
           "contents of the existing copy.\n"
           "****Delete the existing file and try again!";
-        JASSERT(false) (_path) (savedFilePath) (errMsg);
+        bool shouldAssert = true;
+        if (_allow_overwrite) {
+          JTRACE("Copying saved checkpointed file to original location")
+            (savedFilePath) (_path);
+          writeFileFromFd(savedFd, _fds[0]);
```
karya0: We must make take backup of the original user file.

src/plugin/ipc/file/fileconnection.cpp
rohgarg: I'd say that the comparison allows us to save time on copying data from the checkpointed version to the existing version when the contents match.

rohgarg:
@karya0: I have now addressed both of your comments. Let me know how this looks.

src/dmtcp_launch.cpp
```diff
@@ -292,6 +296,9 @@ static void processArgs(int *orig_argc, char ***orig_argv,
     } else if (s == "--checkpoint-open-files" || s == "--ckpt-open-files") {
       checkpointOpenFiles = true;
       shift;
+    } else if (s == "--allow-file-overwrite") {
+      setenv(ENV_VAR_ALLOW_OVERWRITE_WITH_CKPTED_FILES, "1", 0);
```
karya0: We should go over the list of cmdline flags and corresponding env vars to ensure they have "similar" strings.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -228,9 +230,52 @@ void FileConnection::preCkpt()
...
+           .Text("Error creating a backup");
+
+  /* Overwrite the existing file with the contents of the saved
+   * file. We need to open it in WRONLY mode here because the file
+   * might have been opened originally in read-only mode.
+   */
+  int destFileFd = _real_open(_path.c_str(), O_CREAT | O_WRONLY, 0640);
+  JASSERT(destFileFd > 0)(JASSERT_ERRNO)(_path)
+         .Text("Error opening file for overwriting");
+  writeFileFromFd(savedFd, destFileFd);
```
karya0: Validate savedFd

Issue #356, opened 2016-03-28
Add "dmtcp_command -l" to list current processes


karya0:

rohgarg:
Closed via PR #373.

PR #373, opened 2016-05-02
Add dmtcp_command option to print list of connected workers

rohgarg:
This fixes issue #356.

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -226,16 +226,17 @@ void CoordinatorAPI::updateSockFd()
   JASSERT(_coordinatorSocket == PROTECTED_COORD_FD);
 }
 
-void CoordinatorAPI::connectAndSendUserCommand(char c,
+char* CoordinatorAPI::connectAndSendUserCommand(char c,
                                                int *coordCmdStatus,
                                                int *numPeers,
                                                int *isRunning,
                                                int *ckptInterval)
 {
+  char *replyData = NULL;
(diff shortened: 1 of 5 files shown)


src/coordinatorapi.cpp
```diff
@@ -226,16 +226,17 @@ void CoordinatorAPI::updateSockFd()
   JASSERT(_coordinatorSocket == PROTECTED_COORD_FD);
 }
 
-void CoordinatorAPI::connectAndSendUserCommand(char c,
+char* CoordinatorAPI::connectAndSendUserCommand(char c,
```
karya0: Let's change the return type to `string` if possible.

src/coordinatorapi.cpp
```diff
@@ -257,11 +258,11 @@ void CoordinatorAPI::connectAndSendUserCommand(char c,
   //the coordinator will violently close our socket...
   if (c=='q' || c=='Q') {
     *coordCmdStatus = CoordCmdStatus::NOERROR;
-    return;
+    return replyData;
   }
 
   //receive REPLY
-  recvMsgFromCoordinator(&reply);
+  recvMsgFromCoordinator(&reply, (void**)&replyData);
```
karya0: Need to call free on the buffer.

src/coordinatorapi.cpp
rohgarg: Not doing it for now.

src/coordinatorapi.cpp
rohgarg: Done in the new version.

Issue #370, opened 2016-04-30
gzip falure on checkpoint


MrPrezident:
The environment that I'm in uses a wrapper script for a lot of the common unix utilities such as gzip. i.e. $PATH points to a version of gzip that is really a shell script wrapper that ultimately resolves to the real platform dependent gzip executable. The problem is that the wrapper script relies on some environment variables, but when gzip is called from dmtcp, all of the environment variables are gone. i.e. when I try to create a checkpoint, dmtcp calls gzip, which then fails due to undefined environment variables and the checkpoint hangs. It's as if gzip was invoked like this: env -i /path/to/gzip

Why do the env vars get wiped during the gzip call, and is there anything way to fix this?

jiajuncao:
Could you try `--no-gzip` option of dmtcp_launch as a solution? This gets rid of the call to gzip.

MrPrezident:
The App that I use generates checkpoint files that are generally multiple GB in size, so I don't think --no-gzip would be a good idea.

jiajuncao:
What's the error message?

rohgarg:
@MrPrezident: I haven't looked at our code that invokes gzip very closely, but it's likely that there's some piece there that's wiping out the environment variables. I'll try to look at it later today and get back to you.

MrPrezident:
@jiajuncao Here is the error message:

```
/p/flow/iwutilities/bin/gzip: line 55: -MFile::Basename: command not found
/p/flow/iwutilities/bin/gzip: line 67: /.linux50/gzip: No such file or directory
/p/flow/iwutilities/bin/gzip: line 67: exec: /.linux50/gzip: cannot execute: No such file or directory
```
 
When I get these errors, the checkpoint file ends up being of size zero. These are the exact same errors that I get when I do "env -i /p/flow/iwutilities/bin/gzip". Also, I tried creating a fake gzip script that wrote out all of the env vars into a file when executed. Here is the result:

```
PWD=/nfs/project/work/dmtcp-dmtcp-30d2101
SHLVL=1
_=/bin/env
```

@rohgarg Thanks!

I noticed that there is a dmtcp doc about environment variables in the git repo: dmtcp/doc/environment-vars.txt

> After initialization, environment variables are maintained solely in
> user space, and are ignored by the kernel.

Does this account for the problem?

rohgarg:
[...]

PR #375, opened 2016-05-03
ckptserializer: Fix passing a NULL environment to the gzip process

rohgarg:
The existing implementation of the `open_ckpt_to_write()` passes a
NULL environment to the gzip child process. This can be an issue (#370) , as
reported by a user on Github (Github ID: MrPrezident), if gzip (the
real binary or a wrapper script around the real binary) relies on some
environment variables.

The fix here is to let the gzip process inherit the environment of the
parent process by using `execvp(gzip, gzip_args)` instead of `execve(gzip,
gzip_args, NULL)`. Note that the code in the function unsets LD_PRELOAD
for the child process already.

### DIFF
--- a/src/ckptserializer.cpp
+++ b/src/ckptserializer.cpp
@@ -362,7 +362,7 @@ open_ckpt_to_write(int fd, int pipe_fds[2], char **extcomp_args)
       ld_preload_str[0] = '\0';
     }
 
-    _real_execve(extcomp_args[0], extcomp_args, NULL);
+    _real_execvp(extcomp_args[0], extcomp_args);
 
     /* should not arrive here */
     JASSERT(false)
--- a/src/restartscript.cpp
+++ b/src/restartscript.cpp
@@ -160,7 +160,7 @@ static const char* cmdlineArgHandler =
   "          coord_port=\"$2\"\n"
   "          shift; shift;;\n"
   "        --coord-logfile)\n"
-  "          coord_logfile=\"$2\"\n"
+  "          DMTCP_COORD_LOGFILE=\"$2\"\n"
   "          shift; shift;;\n"
   "        --hostfile)\n"
   "          hostfile=\"$2\"\n"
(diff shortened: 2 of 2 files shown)


rohgarg:
@MrPrezident: Could you please verify if this fixes the issue for you?

MrPrezident:
I'm not able to test right now. I'll have to do that later today or tomorrow. One thing I noticed though, is that you used execvp instead of execv. Was this intentional? It looks like the gzip command you are passing in already has the path on it, so I would expect to see execv used instead of execvp. Maybe it doesn't matter though.

rohgarg:
That wasn't really intentional, just personal preference. However, looking through the man page, it seems like execvp might be a bit better. I don't have a strong opinion on this though. I'm happy to change it to execv if that's better.

rohgarg:
@MrPrezident: Did you get a chance to test this?

MrPrezident:
Hi @rohgarg. Nope. Sorry was extremely busy today. I was just about to test it. I will let you know.

rohgarg:
Sure, no problem. I'll push it upstream once you can confirm that it works.

MrPrezident:
It is saving a checkpoint, but the checkpoint file is not working for me. I may be doing something stupid. I'll have to debug this later tonight.

MrPrezident:
For some reason when I run your code, the dmtcp_restart_script.sh script gives me the following error:

`[23735] NOTE at dmtcp_restart.cpp:747 in main; REASON='File doesn't have .dmtcp extension. Check Usage.'`

However, if use the dmtcp_restart command directly on the checkpoint file, it works fine.

rohgarg:
Thanks for testing this, @MrPrezident. I'll take a look at this today and get back to you.

rohgarg:
@MrPrezident: I con confirm that I can reproduce the issue locally. I missed this issue because I usually don't use the restart script, and I didn't realize that `make check` invokes `dmtcp_restart` directly. 

Anyway, I know the problem. It's because of a recent commit: https://github.com/dmtcp/dmtcp/commit/91a651eb15befa3028a29f34ebf6f32f555f42e6 (and not because of this one). I'll push a fix upstream soon.

rohgarg:
I have updated the PR (with a fix for the restart script bug).  Could you please re-try this and let me know?

MrPrezident:
sure! I will do that now.

MrPrezident:
I just tested it. It works great!

PR #379, opened 2016-05-04
Replaced CoordinatorAPI class with namespace.

karya0:
Instead of using a singleton class, we use the namespace with a couple of file-global variables (coordinatorSocket and nsSocket).

This PR is motivated by the interactions around instantiation of CoordinatorAPI singleton class during startup. Often, this has caused some weird code paths where the object instantiation calls malloc which in turn calls other functions/wrappers and so on.

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -46,23 +46,58 @@
 LIB_PRIVATE bool sem_launch_first_time = false;
 LIB_PRIVATE sem_t sem_launch;
 
-namespace dmtcp
+namespace dmtcp {
+namespace CoordinatorAPI {
+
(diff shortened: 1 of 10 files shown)


gc00:
Sorry for this delayed review.  My only worry for this is that the new namespace `CoordinatorAPI` could in principle clash with a namespace of the same name by the end user.  We could call this namespace `DMTCPcoordinatorAPI'.

We could alternatively create a sub-namespace, `dmtcp::CoordinatorAPI::...`, but that is probably a bad idea.  It's a lot to type, and so people will look for workarounds like `using dmtcp::CoordinatorAPI'.  So, I tend to prefer the first option above.

Otherwise, this LGTM.

karya0:
It's already inside the top-level `dmtcp` namespace :).

PR #380, opened 2016-05-04
fileconnection: Fix ckpt-ing of deleted files

rohgarg:
This patch fixes an issue with ckpt-ing of deleted files caused when a plugin
uses the `dmtcp_must_ckpt_file()` API to force checkpointing of a
file that might have been deleted. The fix is to return early from the
`FileConnection::drain()` method in case of a deleted file, i.e., return
without checking for the return value of the `dmtcp_must_ckpt_file()` API.


rohgarg:
@karya0 gave this a ship it!

PR #384, opened 2016-05-11
Support of new dmtcp_launch cmd-line option --with-plugin-32

ankitcse07:
Any built in plugin in dmtcp source can easily load appropriate
version(32/64 bit) automatically in multi-arch support.  However for
external plugins currently there is no way in dmtcp_launch to specify
both versions of libraries so that they get automatically loaded based
on architecture.

So user can specify both versions external plugin shared libraries
64-bit via --with-plugin option and 32-bit via --with-plugin-32.
For e.g.

   dmtcp_launch --with-plugin ./libmineplugin64.so \
          --with-plugin-32 ./libmineplugin32.so ./applic


ankitcse07:
@rohgarg Any updates on reviewing this ?

rohgarg:
LGTM. Could you squash the two commits before pushing upstream?

ankitcse07:
Closing this,  commited 15602f3

PR #385, opened 2016-05-12
This patch will enable pathvirt on process lauched via ssh/rsh.

ankitcse07:
Currently pathvirt is not enabled on processes launched on remote
machine via rsh/ssh. Also the corresponding environment variables
DMTCP_ORIGINAL_PATH_PREFIX/DMTCP_NEW_PATH_PREFIX are not propagated
on restarted process. So setting these environments in restart
within partvirt plugin. Also passing additional arguments to
processes launched on remote machine in ssh plugin.

### DIFF
--- a/plugin/pathvirt/pathvirt.cpp
+++ b/plugin/pathvirt/pathvirt.cpp
@@ -69,6 +69,8 @@ static char newPathPrefixList[MAX_ENV_VAR_SIZE];
 static bool tmpBufferModified = false;
 static pthread_rwlock_t  listRwLock;
 
+EXTERNC int dmtcp_pathvirt_enabled() { return 1; }
+
 /*
  * Helper Functions
  */
@@ -362,6 +364,25 @@ dmtcp_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
            shouldSwap = *oldPathPrefixList && *newPathPrefixList;
        }
        break;
+    }
+    case DMTCP_EVENT_RESTART:
+    {
+      char tmp[MAX_ENV_VAR_SIZE] = {0};
+
+      int ret = dmtcp_get_restart_env(ENV_ORIG_DPP, tmp, sizeof(tmp) - 1);
+      errCheckGetRestartEnv(ret);
+ 
+      if (ret == RESTART_ENV_SUCCESS) {
+        setenv(ENV_ORIG_DPP, tmp, 1);
+      }
+
+      ret = dmtcp_get_restart_env(ENV_NEW_DPP, tmp, sizeof(tmp) - 1);
+      errCheckGetRestartEnv(ret);
+ 
+      if (ret == RESTART_ENV_SUCCESS) {
+        setenv(ENV_NEW_DPP, tmp, 1);
+      }
+      break;
     }
     default:
     ;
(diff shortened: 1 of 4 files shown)


ankitcse07:
@rohgarg this commit is necessary for path virtualization usage in our system as some new process are forked after restart

plugin/pathvirt/pathvirt.cpp
```diff
@@ -363,6 +365,25 @@ dmtcp_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
        }
        break;
     }
+    case DMTCP_EVENT_RESTART:
```
rohgarg: Why do you need to add this event? It seems like what you want is already covered by:
a) call to `pathvirtInitialize()` from `dmtcp::callbackPostCheckpoint()` in `src/mtcpinterface.cpp`; and
b) cases for POST_EXEC and PRE_EXEC events in the pathvirt event hook function?

ankitcse07:
@rohgarg so you are saying that I can move the code for setting the PATH_PREFIX environments to above calls right, instead of adding new event, as ssh plugin requires to get PATH_PREFIX environment on restart.

ankitcse07:
@rohgarg one question on this, I think I won't be able to call dmtcp_get_restart_env within DMTCP_EVENT_PRE_EXEC/DMTCP_EVENT_POST_EXEC right ?

ankitcse07:
@rohgarg you are right I don't need to add RESTART event hook, i can do the same in pathvirtInitialize function, thanks for the suggestion

src/plugin/ipc/ssh/ssh.cpp
```diff
@@ -337,6 +339,23 @@ static void prepareForExec(char *const argv[], char ***newArgv)
     prefix += " --ssh-slave ";
   }
 
+  if(dmtcp_pathvirt_enabled && dmtcp_pathvirt_enabled()) {
+    if(getenv(ENV_NEW_DPP)) {
```
rohgarg: I'm a little uneasy with this solution based on passing information between plugins using environment vars. Couldn't we just use the `get_new_path_prefix_list()` API here?

src/plugin/ipc/ssh/ssh.cpp
```diff
@@ -337,6 +339,23 @@ static void prepareForExec(char *const argv[], char ***newArgv)
...
+      prefix += "/usr/bin/env ";
+      prefix += ENV_NEW_DPP;
+      prefix += "=";
+      prefix += getenv(ENV_NEW_DPP);
+      prefix += " ";
+    }
+    if(getenv(ENV_ORIG_DPP)) {
```
rohgarg: Could we use the `get_original_path_prefix_list()` API here?

ankitcse07:
But using this API broke my code, it seems that pathvirt plugin is not loaded till the time we are in code of ssh doing execv so only environment variables are accessible and so APIs will not work

rohgarg:
Ah! I misread it. It's not the ssh plugin we are passing these value to but the ssh helper. Yes, then I agree with your approach. Ship it!

ankitcse07:
@rohgarg Should I squash 2 commits in one and then push upstream

rohgarg:
Yes, please. We don't want commits that say: "Incorporating changes by XXX" on the main branch. :-)

ankitcse07:
Commit done ead858a , closing the pull request

Issue #389, opened 2016-05-20
Checkpoint crash/hang-up with application that opens more than 1024 file-descriptors


rohgarg:
An issue arises due to the use of `select()` call in `jalib::JMultiSocketProgram::monitorSockets()`, which
is called at checkpoint time. The end result is random hang-ups/crashes. The root cause is the buffer overflows caused by the `FD_{SET|CLR|ISSET}`macros in glibc. However, note that, according to the man page, it's the application's responsibility to ensure that these macros are not called for a file-descriptor that is larger than `FD_SETSIZE`, which is typically 1024 in most implementations.

> Executing FD_CLR() or FD_SET() with a value of fd that is negative or is equal to or larger than FD_SETSIZE will result in undefined behavior.

Also, note the following from the man page:

> POSIX  allows  an implementation to define an upper limit, advertised via the constant FD_SETSIZE, on the range of file descriptors that can be specified in a file descriptor set. The Linux kernel imposes no fixed limit, but the glibc implementation makes fd_set a fixed-size type, with FD_SETSIZE defined as 1024, and the FD_*() macros operating according to that limit.  

The solution is to change the implementation to use `poll()` or `epoll()`.

rohgarg:
Fixed via #390.

PR #390, opened 2016-05-21
monitorSockets: Re-implement using poll()

rohgarg:
This patch fixes an issue (#389) with the `monitorsockets()` function that
would result in a buffer overflow at checkpoint time for applications
that open many file descriptors (> 1024). The issue arises due to
the use of `select()` call.  The root cause for the buffer overflows
is the implementation of `FD_{SET|CLR|ISSET}` macros in glibc. However,
according to the man page, it's the application's responsibility to ensure
that these macros are not called for a file-descriptor that is larger than
FD_SETSIZE, which is typically 1024 in most implementations.

The fix here is to re-implement the function using `poll()`, which scales
better than `select()`.

### DIFF
--- a/jalib/jsocket.cpp
+++ b/jalib/jsocket.cpp
@@ -32,6 +32,7 @@
 #include <string.h>
 #include <unistd.h>
 #include <errno.h>
+#include <poll.h>
 #include <algorithm>
 #include <set>
 #include <typeinfo>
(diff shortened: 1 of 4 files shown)


jalib/jsocket.cpp
```diff
@@ -685,37 +553,41 @@ void jalib::JMultiSocketProgram::monitorSockets ( double dblTimeout )
     //  completion, they then return to monitorSockets() to wait for more
     //  work to do.  dmtcp_coordinator.cpp also describes some of this logic.
     //this will block till we have some work to do
-    int retval = jalib::select ( maxFd+1, &rfds, &wfds, NULL, timeout );
+    //int retval = jalib::select ( numFds+1, &rfds, &wfds, NULL, timeout );
```
karya0: Stale?

jalib/jsocket.cpp
```diff
@@ -585,34 +461,31 @@ void jalib::JMultiSocketProgram::monitorSockets ( double dblTimeout )
...
-      if ( _listenSockets[i].isValid() )
-      {
-        FD_SET ( _listenSockets[i].sockfd(), &rfds );
-        maxFd = std::max ( maxFd,_listenSockets[i].sockfd() );
-      }
-      else
-      {
+    for ( i=0; i<_listenSockets.size(); ++i ) {
+      if ( _listenSockets[i].isValid() ) {
+        numFds += 1;
```
karya0: We can get rid of numFds and use fds.size() instead.

PR #393, opened 2016-05-27
Force glibc to use virt pid/tid

rohgarg:
This is a port of Kapil's patch set for 3.0 to 2.5. This fixes a long-standing issue with checkpointing of mutexes. The earlier fix was to use mutex wrappers in the pid plugin, but it came with a whole can of worms, which forced us to disable mutex tests.

I'm also including a minor patch to remove a stray print statement introduced in commit: 2ca794a.


ankitcse07:
When is this planned to be available in 2.5 branch ?

rohgarg:
@ankitcse07: This isn't quite ready yet. There's at least one case where the current patch fails, which is why the openmp tests are failing. I know the issue and I have a hack-ish fix, but I'd like a cleaner solution.

ankitcse07:
@rohgarg Thanks for the update, what is a timeline for the complete fix

ankitcse07:
@rohgarg Ping !

gc00:
@ankitcse07, I believe that @rohgarg is still a little over-loaded.  I expect that after Wed., June 22, he will have more time to look at this in detail.

ankitcse07:
@rohgarg Could you please update the branch aligning to latest 2.5, actually I had merged this PR change in my local sandbox, I guess only addition to above is your last commit right ?

src/dmtcp_launch.cpp
```diff
@@ -591,6 +593,28 @@ int main ( int argc, char** argv )
...
+extern "C" int __clone(int (*fn) (void *arg), void *child_stack, int flags,
+                       void *arg, int *ptid,
+                       struct user_desc *tls, int *ctid)
+{
+  intptr_t tidOffset = (char*)ptid - (char*)tls;
+  char value[100] = {0};
+  snprintf(value, 100, "%d", tidOffset);
+  setenv(ENV_VAR_TID_OFFSET, value, 1);
+}
+
```
ankitcse07: @rohgarg shouldn't this overwrite of __clone be calling actual __clone definition ?

ankitcse07:
invoking 
    % dmtcp_launch sort <file name>

reproduces the above issue

rohgarg:
I'm unable to reproduce the assert with `sort` locally. Are you able to reproduce the error with other programs as well? Could you verify that the `setenv()` call in `__clone()` is being triggered before the exec?

ankitcse07:
@rohgarg Its also reproducible with other programs.  I can see the setenv happening in __clone but still the same assert. I can provide you a webex if you want to debug this. Let me know I am available now

ankitcse07:
@rohgarg I am not sure what went wrong, but it worked after clean build.  However I can see now lot of warning coming now 

Error     [RTS-20110]: : EDS Pid:2735000 host:iolas ErrorMsg:ERROR: ld.so:
                       object 'libdmtcp_alloc.so' from LD_PRELOAD cannot be
                       preloaded: ignored.
                       ERROR: ld.so: object 'libdmtcp_dl.so' from LD_PRELOAD
                       cannot be preloaded: ignored.
                       ERROR: ld.so: object 'libdmtcp_ipc.so' from LD_PRELOAD
                       cannot be preloaded: ignored.
                       ERROR: ld.so: object 'libdmtcp_svipc.so' from LD_PRELOAD
                       cannot be preloaded: ignored.
                       ERROR: ld.so: object 'libdmtcp_timer.so' from LD_PRELOAD
                       cannot be preloaded: ignored.
                       ERROR: ld.so: object 'libdmtcp.so' from LD_PRELOAD cannot
                       be preloaded: ignored.

ankitcse07:
@rohgarg Create a 32-bit mode package, or multi-arch package.  Launch any 32-bit binary under dmtcp_launch and you will hit the issue.

ankitcse07:
@rohgarg Were you able to reproduce the issue, or I can provide you a webex for the same.

gc00:
@ankitcse07, I pushed in PR #406 about 6 days ago.  I pushed it to 2.5 and master.  Are you testing with this commit in place?

ankitcse07:
@gc00 I had updated the above branch with all commits in 2.5.  Issue is still coming

rohgarg:
@ankitcse07: Yes, I can reproduce the issue locally. I didn't get time to dig into it any further though. The assert is triggered because the tid offset we compute prior to exec is different than the offset we compute after exec. It seems like we'll need a different function for 32-b but I'm not really sure yet.

ankitcse07:
@rohgarg Somehow this commit seems to screw the gdb.  It seems virtual pid/tid are now being seen by gdb and hence cannot attach to them.   This can really become trouble for debugging 


warning: Can't attach LWP 40002: No child processes
warning: Can't attach LWP 40000: No child processes
linux-thread-db.c:438: internal-error: thread_get_info_callback: Assertion `inout->thread_info != NULL' failed.
A problem internal to GDB has been detected,
further debugging may prove unreliable.

ankitcse07:
@rohgarg This commit is also messing up the signal handlers, for assert which raises SIGABRT, SIGSEGV is being received.  See the below test, just run under dmtcp_launch with this branch

rohgarg:
@ankitcse07: I have figured out the issue with signals: we need a wrapper for `raise()`, that temporarily disables virtual pids/tids in glibc, similarly to other wrappers in `pid_miscwrappers.cpp`.

For the issue with GDB, I have a feeling that we might need ptrace wrappers, but I will have to look more carefully.

rohgarg:
It seems like a wrapper for `raise()` does not resolve this issue because libc calls raise directly instead of going though PLT. The wrapper does resolve an issue where an application that calls `raise()` directly would receive an incorrect signal, but it doesn't fix this issue where a call to `assert()` causes a segfault. Anyway, I have updated the PR if you'd like to take a look.

At this point, I feel like the current patch is getting too complicated and introduces more bugs/corner cases than it fixes. I'm wondering if we should kill this idea and think of a different approach.

ankitcse07:
@rohgarg Add wrapper for assert too reset pids/tids to real one in libc, would this work

rohgarg:
@ankitcse07: A direct wrapper for assert doesn't work because assert is a macro that calls various __assert* functions. I have updated the PR to add one such wrapper. This fixes the issue you were facing. The other one that still remains is the issue with ptrace.

ankitcse07:
@rohgarg ptrace issue is the issue in gdb session right ? What about the issue that I faced in multi-arch package.

rohgarg:
@ankitcse07: Yes, ptrace is for GDB. I haven't been able to figure out the 32-bit issue yet.

rohgarg:
Closing this until we can figure out a better/cleaner approach.

PR #397, opened 2016-06-02
Remove old debugging stmt from commit 2ca794acc

gc00:
This is one line: remove an old debugging stmt that should never have been checked in.  Can someone do a quick review?  Thanks.

### DIFF
--- a/jalib/jalib.cpp
+++ b/jalib/jalib.cpp
@@ -166,6 +166,10 @@ namespace jalib {
                                         timeout);
   }
 
+  int poll(struct pollfd fds[], nfds_t nfds, int timeout) {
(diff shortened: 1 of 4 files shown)


src/dmtcp_restart.cpp
```diff
@@ -801,7 +801,6 @@ int main(int argc, char** argv)
   bool foundNonOrphan = false;
   RestoreTargetMap::iterator it;
   int size = independentProcessTreeRoots.size();
```
rohgarg: Let's remove this as well. I don't think this `size` variable is being used anywhere else.

gc00:
github claims I closed this for some reason.  I don't think I did.  Also, it's no longer recording my push'es, when I do:  `https://github.com/dmtcp/dmtcp.git`

PR #399, opened 2016-06-02
Fix commit 2590376dc (restoring ARM_HOST)

gc00:
In commit 2590376dc (Sep. 16, 2015), I accidentally removed ARM (32-bit) support while trying to add AARCH64 support.  I've tested this restoration on a BeagleBoard (ARM Cortex A8), and it seems to be working there.

The only real change is in configure.ac.  You can see what I did by looking at my buggy commit:
`git diff 2590376dc^ 2590376dc -- configure.ac`
and comparing that old commit with the diff for configure.ac in this commit request.

I'm hoping for a fast review.  :-)  This should go into all of 3.0, 2.5, and 2.4.5.

### DIFF
--- a/configure
+++ b/configure
@@ -664,6 +664,9 @@ HAS_CMA
 AARCH64_HOST
 AARCH64_HOST_FALSE
 AARCH64_HOST_TRUE
+ARM_HOST
+ARM_HOST_FALSE
+ARM_HOST_TRUE
 HAS_READLINE
 M32
 CONFIG_M32_FALSE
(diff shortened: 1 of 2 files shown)


rohgarg:
@gc00: Should we have similar changes in Makefile.am?

gc00:
@rohgarg, Thanks for that comment.  You're right.

I decided to remove all changes from any `Makefile.in` file, rather than struggle with `Makefile.am`.  We are _only_ using the `ARM_HOST` variable in `src/mtcp`, and `src/mtcp/Makefile.in` already had the code from before that recognized `ARM_HOST`.  So, updating `configure` and `configure.ac` will be enough.

Is the code okay to push in now?

rohgarg:
LGTM.

PR #400, opened 2016-06-03
Adding system call wrappers in path-virt plugin

ankitcse07:
Wrapper definitions for following system calls have been
added in path-virt plugin
  chmod
  unlink
  chdir
  remove
  rmdir
  link
  symlink
Each of these calls takes path as arguments which is required to be
converted to mapped path when path-virt plugin is enabled


rohgarg:
LGTM.

ankitcse07:
9b633e4  commit done closing the PR

PR #401, opened 2016-06-06
pid-plugin: Fix restart callback to use _real_* functions

rohgarg:
This appears to be an old bug. Usage of close()/dup2() can lead
to unexpected consequences where a fd associated with a different
plugin (socket, etc.) may be closed affecting their behavior on
restart.

I noticed this issue when a raw socket associated with the IPC plugin
was being closed unintentionally as a result of calling close() in the
pid plugin's restart callback. The temporary PIDMAP fd opened in the
restart callback happened to be the same as the socket fd that the raw
socket was associated with prior to checkpointing.

### DIFF
--- a/src/plugin/pid/pid.cpp
+++ b/src/plugin/pid/pid.cpp
@@ -147,9 +147,9 @@ static void openOriginalToCurrentMappingFiles()
   if (!Util::isValidFd(PROTECTED_PIDMAP_FD)) {
     fd = openSharedFile(pidMapFile, O_RDWR);
     JASSERT (fd != -1);
-    JASSERT (dup2 (fd, PROTECTED_PIDMAP_FD) == PROTECTED_PIDMAP_FD)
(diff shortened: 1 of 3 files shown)


jiajuncao:
This may be a digression, but is it possible to call these functions directly into libc? In general, as developers, we should know which fds, sockets, shared-memory objects, etc. are used internally, and shouldn't be exposed to any plugin. The `REAL_FUNC_PASSTHROUGH` macro is still going to invoke the lower-level plugins, which may be an issue in the future, if something similar happens.

rohgarg:
> This may be a digression, but is it possible to call these functions directly into libc?

At this point, no. PID plugin is last plugin in the library search order, so effectively, we are calling in to libc. But yes, we do need such an interface in the future.

gc00:
It looks good to me.  Go ahead and check it in.  I'm also adding a 2.4.5 label to this for the 2.4.5 ponit release.

I checked and `openSharedFile` is using _real_XXX() to create the fd.  So, we should also use _real_dup2() and _real_close() to continue processing it.

PR #402, opened 2016-06-07
socket-plugin: Add basic support for raw (netlink) sockets

rohgarg:
This patch adds some basic support for raw (netlink) sockets with some
reorganization and clean-up of the code in the socket plugin.

Future patches will make the support more robust.

### DIFF
--- a/src/plugin/ipc/socket/socketconnection.cpp
+++ b/src/plugin/ipc/socket/socketconnection.cpp
@@ -28,6 +28,7 @@
 #include <arpa/inet.h>
 #include <signal.h>
 #include <poll.h>
+#include <linux/netlink.h>
 
 #include "dmtcp.h"
 #include "shareddata.h"
@@ -75,10 +76,41 @@ SocketConnection::SocketConnection(int domain, int type, int protocol)
   , _sockType(type)
   , _sockProtocol(protocol)
   , _peerType(PEER_UNKNOWN)
+  , _listenBacklog(-1)
+  , _bindAddrlen(0)
+  , _remotePeerId(ConnectionIdentifier::null())
 { }
 
+SocketConnection::SocketConnection(int domain, int type, int protocol, ConnectionIdentifier remote)
+  : _sockDomain(domain)
+  , _sockType(type)
+  , _sockProtocol(protocol)
+  , _peerType(PEER_UNKNOWN)
+  , _listenBacklog(-1)
+  , _bindAddrlen(0)
+  , _remotePeerId(remote)
+{ }
+
+void SocketConnection::onBind(const struct sockaddr* addr, socklen_t len)
+{
+  JASSERT(false).Text("Bind called on unknown socket type");
+}
+
+void SocketConnection::onListen(int backlog)
+{
+  JASSERT(false).Text("Listen called on unknown socket type");
+}
+
+void SocketConnection::onConnect(const struct sockaddr *serv_addr,
+                                 socklen_t addrlen,
+                                 bool connectInProgress)
+{
+  JASSERT(false).Text("Connect called on unknown socket type");
+}
+
+
 void SocketConnection::addSetsockopt(int level, int option,
-                                     const char* value, int len)
+                                     const void* value, int len)
 {
   _sockOptions[level][option] = jalib::JBuffer(value, len);
 }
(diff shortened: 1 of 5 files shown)


src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -95,7 +127,7 @@ void SocketConnection::restoreSocketOptions(vector<int>& fds)
       int ret = _real_setsockopt(fds[0], lvl->first, opt->first,
                                  opt->second.buffer(),
                                  opt->second.size());
-      JASSERT(ret == 0) (JASSERT_ERRNO) (fds[0])
+      JWARNING(ret == 0) (JASSERT_ERRNO) (fds[0])
```
gc00: I assume the reason for this is that if the socket is a netlink, then we expect the call to `setsockopt()` to fail.
But then shouldn't we also add a `JASSERT` in the situation where we know that the socket is not associated with netlink?

src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -178,9 +210,6 @@ void SocketConnection::serialize(jalib::JBinarySerializer& o)
   TcpConnection::TcpConnection(int domain, int type, int protocol)
   : Connection(TCP_CREATED)
   , SocketConnection(domain, type, protocol)
```
gc00: I don't understand why we can delete things like `_bindAddrlen`.  The variable is used lower in the code.  Don't we still need to set a default value for these variables?

src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -345,10 +374,7 @@ void TcpConnection::onConnect(const struct sockaddr *addr,
 TcpConnection::TcpConnection(const TcpConnection& parent,
                              const ConnectionIdentifier& remote)
   : Connection(TCP_ACCEPT)
-  , SocketConnection(parent._sockDomain, parent._sockType, parent._sockProtocol)
-  , _listenBacklog(-1)
-  , _bindAddrlen(0)
-  , _remotePeerId(remote)
+  , SocketConnection(parent._sockDomain, parent._sockType, parent._sockProtocol, remote)
```
gc00: Same comment as above.  Don't we need the default values?

src/plugin/ipc/socket/socketconnection.h
```diff
@@ -119,23 +137,23 @@ namespace dmtcp
...
   };
 
   class RawSocketConnection : public Connection, public SocketConnection
   {
     public:
+      enum RawType
+      {
+        RAW_INVALID = RAW,
+        RAW_ERROR,
+        RAW_CREATED,
```
gc00: Can we find a more meaningful name for `RAW_CREATED`?  Does this mean raw socket that is uninitialized?  How about `RAW_UNINITIALIZED` or `RAW_UNBOUND` or `RAW_NEW_ENDPOINT`?

gc00:
@rohgarg, I have some inline comments.  The code looks reasonable.  However, I'm wondering if there is a better way to organize this code for maintainability.  Please see my next paragraph.

@karya0, Currently our ipc plugin supports five related plugins that all use some common code in `ipc.cpp` and other files.  The pull request for one of the five "sub-plugins" (for 'socket') is about to create support for a new kind of socket, netlink sockets.   Currently, we support TCP, but not DGRAM.  Eventually, we'll probably want to support TCP, DGRAM, NETLINK, and others.

This seems like a good time to think through the software architecture that we're using.  Should we create sub-sub-directories inside the socket subdirectory, to support sub-sub-plugins?  In this case, the sub-sub-plugins would share code from connectionrewirer.cpp and bufferdrainer.cpp in the socket subdirectory.

src/plugin/ipc/socket/socketconnection.cpp
rohgarg: The fields have been moved to the super class, `SocketConnection`, because I felt that these are common to all socket types. They will be initialized to the default values in the constructor of `SocketConnection`.

src/plugin/ipc/socket/socketconnection.cpp
rohgarg: Same as above.

PR #394, opened 2016-05-30
Issue in checkpointing opened file deleted in between by parent 

ankitcse07:
Consider a scenario in which file is already opened by
parent and now child process is forked. DMTCP infrastructure
doesn't populates _flags field in FileConnection class of
child process. Now at certain time file is deleted using
unlink however file pointer was not closed. Further
child process invokes the application level checkpoint.

This has resulted in issue. Fix is to use _fcntlFlags
filed

PR #403, opened 2016-06-07
File deleted bug 

ankitcse07:
Consider a scenario in which file is already opened by
parent and now child process is forked. DMTCP infrastructure
doesn't populates _flags field in FileConnection class of
child process. Now at certain time file is deleted using
unlink however file pointer was not closed. Further
child process invokes the application level checkpoint.

This has resulted in issue. Fix is to use _fcntlFlags
filed


ankitcse07:
This PR was opened to commit in master branch. For ref see #394 PR

PR #404, opened 2016-06-08
Fix an assert in case of a failed exec call

rohgarg:


src/dmtcpworker.cpp
```diff
@@ -104,7 +104,18 @@ void restoreUserLDPRELOAD()
...
-  JASSERT(userPreload == NULL || strlen(userPreload) <= strlen(preload));
+  //   NOTE: Consider the following sequence of calls:
+  //   dmtcp_launch a.out
+  //     exec(a.out) --> this function gets called from DmtcpWorker(), and the LD_PRELOAD is destroyed
+  //       exec(b.out)
+  //    If the call to exec b.out fails, this function will be called from
+  //    dmtcpProcessFailedExec(). The following JASSERT could fail if the user
+  //    had a non-null preload env var. And so, we need to be less aggressive,
+  //    report this to the user, and let the user handle the failed exec.
+  //JASSERT(userPreload == NULL || strlen(userPreload) <= strlen(preload));
```
gc00: @rohgarg, I'm not sure that I understand the reasoning.  The `JASSERT` was meant to protect us because 4 to 9 lines below (at lines 120-124), we will call 

`preload[0] = '\0'; ... strcat(preload, userPreload);`

We need to ensure that we don't overrun the buffer of preload.  Please see my analysis in the general discussion.

gc00:
@rohgarg, When we call `strcat(preload, userPreload);`, we depend on preload having allocated enough storage to store the characters of userPreload. So, we set `preload[0] = '\0';` and then re-use the space that we just freed by copying in the characters from userPreload.  (Note also that `preload` is pointing into the environment variable region itself.  So, we are playing tricks with the environment variable storage following "LD_PRELOAD=".)

If we remove the JASSERT, we _will_ overrun the storage and modify the next environment variable.
If we are to replace the JASSERT by a JWARNING, we are doing something dangerous.

An alternative workaround would be that in this case, we can invoke the original code: `setenv("LD_PRELOAD", userPreload, 1);` in the situation where JASSERT would have failed.  And then, just _before_ this call to `setenv()`, we would call

```
  JWARNING(...)
  .Text("Since 'strlen(userPreload) > strlen(preload)' for LD_PRELOAD, we're forced\n"
  "to call `setenv()`.  In unusual circumstances, this can cause deadlock.\n"
  "(See comments above this in the source code for the reaons.)");
```

rohgarg:
> An alternative workaround would be that in this case where the JASSERT would have triggered, we can instead invoke the original code: setenv("LD_PRELOAD", userPreload, 1); from an earlier version of DMTCP. And then, just before this call to setenv(), we would call ...

As noted in a comment at the beginning of the function, a call to `setenv()` can result in a call to malloc and a deadlock. I think we need a different strategy. How do you feel about this:

```C
...
  static bool okayToAppend = false;
  JWARNING(userPreload == NULL || strlen(userPreload) <= strlen(preload))
          (userPreload)(preload);
  okayToAppend = strlen(userPreload) <= strlen(preload);

  // Destructively modify environment variable "LD_PRELOAD" in place:
  preload[0] = '\0';
  if (userPreload == NULL) {
    //_dmtcp_unsetenv("LD_PRELOAD");
  } else {
    if (preload && strlen(preload) == 0 && okayToAppend) {
      strcat(preload, userPreload);
    }
  }
...
```

gc00:
As noted in a comment at the beginning of the function, a call to `setenv()` can result in a call to malloc and a deadlock. I think we need a different strategy. How do you feel about this:

```C
...
  static bool okayToAppend = false;
  JWARNING(userPreload == NULL || strlen(userPreload) <= strlen(preload))
          (userPreload)(preload);
  okayToAppend = strlen(userPreload) <= strlen(preload);

  // Destructively modify environment variable "LD_PRELOAD" in place:
  preload[0] = '\0';
  if (userPreload == NULL) {
    //_dmtcp_unsetenv("LD_PRELOAD");
  } else {
    if (preload && strlen(preload) == 0 && okayToAppend) {
      strcat(preload, userPreload);
    }
  }
...
```

First of all, in your pseudocode: `okayToAppend = strlen(userPreload) <= strlen(preload);`, it will blow up if userPreload == NULL.  So, you would need a patch.

Second, maybe we need to talk on the phone, because I think there's something we're not communicating about.

For example, your condition `preload && strlen(preload) == 0 && okayToAppend` seems to say that we will drop the LD_PRELOAD of the user permanently unless the DMTCP version of LD_PRELOAD has been set, and it is set to the null string.

From where I stand, here are the problems, just for a _single_ call to exec:

If it's not possible to destructively modify LD_PRELOAD to remove the DMTCP preload and add the userPreload, then we used to JASSERT() and quick.  We thought this was better than over-writing some other environment variables.  Your first version of the code seems to say to just send the user a JWARNING, and go ahead and over-write some environment variables, and hope that the destruction of environment variables isn't too harmful.  Your second version of the code, here, seems to say that if we can't patch LD_PRELOAD with the userPreload, then issue a warning that shows our DMTCP preload (which we intend to drop and not use), and the user's userPreload, (which we intend to drop and not use because it's too long), and hope that the user realizes that this is about an exec (even though we never said that to the user) and that we're worrying about how to set LD_PRELOAD.  The thought seems to be that removing the userPreload with no explanation to the user is preferred over using `setenv()` and warning the user that this might deadlock.  If I'm right about what you're saying, then I don't agree. :-(

If I understand your patch, you now want to make this work for two exec's in a row, where the second exec is called from the first program that was exec'ed.

If this is the case, can you modify all of the comments to clearly indicate the several cases in one unified description:
# There is a danger of deadlock if we use `setenv()`.  See the comments above in the source code for an explanation.  So, we will try to over-write the LD_PRELOAD value directly with the user's value.  _(QUESTION:  Are we doing this before the exec?  Why is it safe in DMTCP to remove the DMTCP LD_PRELOAD?)_
# If the user's LD_PRELOAD is larger than the DMTCP LD_PRELOAD, then we will do what? _(I suggest to do `setenv()` anyway and first warn the user.  But there are other slightly more complicated alternatives such as assigning a very large LD_PRELOAD environment variable in advance, and terminating it with '\0' at the end of the the actual string while leaving a larger buffer available.)_
# Clearly explain the additional problems that occur if an exec'ed program then calls another exec.  In this case, what is the proposed solution.

I would have to struggle to understand what is the exec followed by exec that generates problems.  Could you also revise those comments?  Thanks.

gc00:
@rohgarg, This PR has been around for really a long time.  Can we resolve this -- maybe by talking together?

PR #406, opened 2016-06-08
Fix mtcp_restart for 32-bit mode for Linux-2.6.18

gc00:
This should be an easy review.

First commit is a short fix for 32-bit DMTCP under Linux-2.6.18.  The 'vdso' segment is not annotated by the kernel, and a user needed DMTCP to work in this configuration.

The second commit is minor:  just changing formatting strings in `src/mtcp/mtcp_restart.c` for greater clarity.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -635,7 +635,14 @@ static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
       vdsoEnd = area.endAddr;
       DPRINTF("***INFO: vDSO found (%p-%p)\n orignal vDSO: (%p-%p)\n",
               area.addr, area.endAddr, rinfo->vdsoStart, rinfo->vdsoEnd);
-    } else if (mtcp_strcmp(area.name, "[vvar]") == 0) {
+    }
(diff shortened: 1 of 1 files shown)


jiajuncao:
Is the vdso page address always at 0xfffe0000 before kernel 2.6.18? If not, we'd have to come up with a general solution to get its address.

gc00:
@jiajuncao wrote:
> Is the vdso page address always at 0xfffe0000 before kernel 2.6.18? If not, we'd have to come up with a general solution to get its address.

In principle, you're right, Jiajun.  My memory is that in older Linux kernels, this memory region didn't exist, and that in Linux 2.6.18, it was permanently set at 0xfffe0000 (unless the user requested some very strange linking commands).  I'm sure that in five years, no one will be using 2.6.18 any more.  So, I'm thinking of this patch purely for legacy applications.  The 2.6.18 kernel will not change in the future, since people will be told to upgrade their kernel.  So, I think this is reasonable.

gc00:
@jiajuncao:  Given that this patch is intended only for the common case for older legacy code, is it okay to push this in?

jiajuncao:
@gc00: Sure, please push the change in.

PR #407, opened 2016-06-11
socket-plugin: Replace call to select() in drain() with poll()

rohgarg:
The select() call is only capable of handling 1024 fds. Usage of select()
with more than 1024 fds can lead to undefined behavior because of the
way it's implemented in libc. This patch replaces a call to select()
with poll() in the drain() method of the TcpConnection class.


gc00:
Looks good to me.

PR #413, opened 2016-06-27
DMTCP 2.4.5

karya0:


karya0:
I am not so sure about shipping pthread-mutex wrappers. We have seen several issues with them (during initialization, etc.). Would it be possible to enable them with a configure option, say `--enable-pthread-wrappers` that is disabled by default and marked as experts-only?

CC: @gc00 @jiajuncao @rohgarg

gc00:
@karya0, Sorry for the late reply.  How about if we add:  `--enable-pthread-wrappers` along with a hep message that includes the statement in the configure.ac: `"recommended when using Open MPI"`?  Could you make that modification?

@jiajuncao, My memory is that we need those pthread wrappers only for supporting Open MPI, and not for the other MPIs.  Is this what you remember?

gc00:
@karya0, can we create a 2.4.5 release?  There are people who need these fixes.  Would you like to add:  `--enable-pthread-wrappers` along with a help message that includes the statement in the configure.ac: `"recommended when using Open MPI"`?  Or should I create that?  Thanks.

PR #415, opened 2016-06-28
Creating PR to merge in 3.0.0 branch for #414 Closed PR 

ankitcse07:
Current strategy is to skip pages storing in checkpoint database
which are not anonymous and don't have any READ/WRITE permissions.
Anonymous area is identified reading /proc/<pid>/maps file with empty
string for the name. Adding only anonymous pages was necessary to checkpoint
database as user might have removed the permissions temporarily and
may later restore the permissions for later use. In case restore
of permissions happens after restart and in case we have skipped mapping
these pages to user memory then it will result in segvs if in user code.

Assumption that only anonymous areas have this property is breaking
on some kernels.

In my case if memory is allocated on heap and mprotect gives it PROT_NONE
permission then I see different behavior on SUSE12 and RHEL6 machines.

1) On Suse12(Kernel 3.0.13-0.27) after mprotect call memory area is still
marked as heap segment

2) However on RHEL6 (Kernel 2.6.32-431)  after mprotect call memory
area is marked as anonymous

1st scenario is breaking code in writeckpt.cpp which only assumes area
with non-empty string with no read/write permissions should be skipped
being stored in checkpoint image, hence causing issues in restart as
area is not mapped in process memory.

This fix now also stores non-rw pages belonging to labels [heap],[stack],
[stack:XXX] apart from anonymoys pages in checkpoint image.

### DIFF
--- a/src/writeckpt.cpp
+++ b/src/writeckpt.cpp
@@ -205,10 +205,15 @@ void mtcp_writememoryareas(int fd)
      *
      * We suspect that libpthread is using mmap() instead of mprotect to change
      * the permission from "---p" to "rw-p".
+     *
+     * Also, on SUSE 12, if this region was part of heap, the protected region 
+     * may have the label "[heap]".  So, we also save the memory region if it 
+     * has label "[heap]", "[stack]", or  "[stack:XXX]".
      */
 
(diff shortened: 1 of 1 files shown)


ankitcse07:
Creating this PR in order to commit in master branch for the issue described in #414

PR #418, opened 2016-07-04
delayresume: Add new plugin

rohgarg:
The README file describes the plugin, usage, and use-case.

### DIFF
--- a/contrib/delayresume/delayresume.cpp
+++ b/contrib/delayresume/delayresume.cpp
@@ -1,0 +1,118 @@
+#include "jassert.h"
+#include "dmtcp.h"
+#include "delayresume.h"
+
+#undef dmtcp_checkpoint_block_resume
+#undef dmtcp_checkpoint_unblock_resume
+
+/* Set to true for the user thread that invokes a checkpoint */
+__thread bool ckptInitiator = false;
+
+/*
+ * Global lock; direct accesses outside of block/unblock functions (below)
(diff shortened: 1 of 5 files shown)


contrib/delayresume/delayresume.cpp
```diff
@@ -0,0 +1,121 @@
...
+dmtcp_checkpoint_block_resume()
+{
+  int ret = DMTCP_NOT_PRESENT;
+  ckptInitiator = true;
+  blockUserThreads();
+  ret = dmtcp_checkpoint();
+  if (ret == DMTCP_NOT_PRESENT) {
+    unblockUserThreads();
+    ckptInitiator = false;
+  } else if (ret == DMTCP_AFTER_RESTART) {
```
jiajuncao: The two branches do the same thing.

jiajuncao:
Looks good.

I have one question regarding the motivation: if the block/unblock is done after resuming, all checkpoint-related events have finished, how does user's behavior at this point affect the checkpoint performance?

contrib/delayresume/delayresume.cpp
rohgarg: Good point! I'll rework this.

rohgarg:
I'm not sure what you mean by the checkpoint performance here. The goal is to allow a user thread that requested for a checkpoint to block other user threads from resuming until the checkpoint-initiator gives a go-ahead.

Issue #332, opened 2016-02-28
Compilation error with `--enable-debug` on master


rohgarg:
util_init.cpp fails to compile if I configure DMTCP with `--enable-debug`. Here's the error message:

``` bash
make[2]: Entering directory '/home/rohgarg/proj/rohgarg-github/dmtcp-4/dmtcp/src'
g++ -DHAVE_CONFIG_H -I. -I../include  -I../include -I../jalib  -fPIC -DDEBUG -g -O0 -Wall -MT util_init.o -MD -MP -MF .deps/util_init.Tpo -c -o util_init.o util_init.cpp
util_init.cpp: In function ‘void dmtcp::Util::initializeLogFile(dmtcp::string, dmtcp::string, dmtcp::string)’:
util_init.cpp:154:23: error: ‘JLOG’ was not declared in this scope
   JLOG(a.str().c_str());
```

rohgarg:
The bug was introduced in commit: d46b392. In the commit the `JLOG()` macro was deleted but all of its references were not cleaned up.

rohgarg:
Fixed via PR #460.

PR #420, opened 2016-07-05
Comment JLOG call, fails with --enable-debug

FrancoisAi:
The title makes it clear.

### DIFF
--- a/src/util_init.cpp
+++ b/src/util_init.cpp
@@ -151,7 +151,8 @@ void Util::initializeLogFile(string tmpDir, string procname, string prevLogPath)
   }
   a << "\n========================================\n";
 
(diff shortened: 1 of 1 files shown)


rohgarg:
Thanks for the patch, @FrancoisAi. There's an existing issue for tracking this. See issue #332. IIRC, @karya0 had a patch for this that works without having to comment out the call to `JLOG`. @karya0, would you be able to push that in?

gc00:
@karya0, Ping.  :-)  Should we directly push in this pull request ourselves, or do you have a preferred solution based on issue #332?

rohgarg:
Ship it!

rohgarg:
This PR was migrated to #460 and pushed in.

PR #421, opened 2016-07-05
Fix warnings

FrancoisAi:
This fixes warnings seen in gcc-5.2.

### DIFF
--- a/plugin/modify-env/modify-env.c
+++ b/plugin/modify-env/modify-env.c
@@ -113,14 +113,21 @@ char * read_dmtcp_env_file(char *file, int size) {
   }
 #else
   char pathname[512];
-  if (strlen(dmtcp_get_ckpt_dir()) > sizeof(pathname)-1-sizeof(file)) {
-    warning(__FILE__ ": Pathname of ckpt dir is too long: ",
-            dmtcp_get_ckpt_dir() /* , "\n" */ );
-    exit(1);
+  // Fix for picky compiler.  The function dmtcp_get_ckpt_dir() would only
+  //   return NULL if we were not running under DMTCP.  But the compiler
+  //   doesn't know that.
+  const char *dmtcp_ckpt_dir = dmtcp_get_ckpt_dir();
+  if (dmtcp_ckpt_dir) {
+    if (strlen(dmtcp_ckpt_dir) + strlen(file) > sizeof(pathname)-1) {
+      fprintf(stderr, "%s:%d : Pathname of ckpt dir is too long: %s\n",
+              __FILE__, __LINE__, dmtcp_ckpt_dir);
+      exit(1);
+    }
+    // This can not overwrite the pathname buffer due to the if condition.
+    strcpy(pathname, dmtcp_ckpt_dir);
+    strcpy(pathname + strlen(dmtcp_ckpt_dir), "/");
+    strcpy(pathname + strlen(dmtcp_ckpt_dir) + strlen("/"), file);
   }
-  strcpy(pathname, dmtcp_get_ckpt_dir());
-  strcpy(pathname + strlen(dmtcp_get_ckpt_dir()), "/");
-  strcpy(pathname + strlen(dmtcp_get_ckpt_dir()) + strlen("/"), file);
   int fd = open(pathname, O_RDONLY);
   if (fd < 0) {
     return NULL;
--- a/src/execwrappers.cpp
+++ b/src/execwrappers.cpp
@@ -232,7 +232,7 @@ int daemon(int nochdir, int noclose)
   }
 
   if (!nochdir) {
-    (void)chdir("/");
+    int dummy = chdir("/");
   }
 
   if (!noclose) {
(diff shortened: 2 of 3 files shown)


plugin/modify-env/modify-env.c
```diff
@@ -113,14 +113,20 @@ char * read_dmtcp_env_file(char *file, int size) {
...
+  //   return NULL if we were not running under DMTCP.  But the compiler
+  //   doesn't know that.
+  const char *dmtcp_ckpt_dir = dmtcp_get_ckpt_dir();
+  if (dmtcp_ckpt_dir) {
+    if (strlen(dmtcp_ckpt_dir) > sizeof(pathname)-1-sizeof(file)) {
+      fprintf(stderr, "%s:%d : Pathname of ckpt dir is too long: %s\n",
+              __FILE__, __LINE__, dmtcp_ckpt_dir);
+      exit(1);
+    }
+    strcpy(pathname, dmtcp_ckpt_dir);
```
rohgarg: I guess since you are verifying the length of the string above, it's okay to use strcpy here. I'd still prefer using strncpy instead.

plugin/modify-env/modify-env.c
```diff
@@ -113,14 +113,20 @@ char * read_dmtcp_env_file(char *file, int size) {
...
-  if (strlen(dmtcp_get_ckpt_dir()) > sizeof(pathname)-1-sizeof(file)) {
-    warning(__FILE__ ": Pathname of ckpt dir is too long: ",
-            dmtcp_get_ckpt_dir() /* , "\n" */ );
-    exit(1);
+  // Fix for picky compiler.  The function dmtcp_get_ckpt_dir() would only
+  //   return NULL if we were not running under DMTCP.  But the compiler
+  //   doesn't know that.
+  const char *dmtcp_ckpt_dir = dmtcp_get_ckpt_dir();
+  if (dmtcp_ckpt_dir) {
+    if (strlen(dmtcp_ckpt_dir) > sizeof(pathname)-1-sizeof(file)) {
```
rohgarg: Are you sure it should be `sizeof(file)` here? It seems like the variable `file` is a pointer.

rohgarg:
LGTM modulo two minor comments.

plugin/modify-env/modify-env.c
FrancoisAi: In the old lines 121, 122, 123 (see below) the code was already using strcpy. We didn't change that part.

rohgarg: Yep, I saw that. My point was to make it a little more robust for the future. I don't have a strong opinion here though, as we are already testing for the length of the string in the if clause above.

FrancoisAi:
We fixes the bug with `sizeof(file)`. We added a comment to note that strcpy is safe.
Is the pull request okay right now?

rohgarg:
Ship it!

(It's interesting that the old code was using `sizeof(file)` and we never encountered an issue.)

gc00:
@rohgarg, My memory is that you said that a different PR has now made these same fixes.  If so, could you write that in a comment (referring to the other PR), and then select "Close and comment"?  Thanks.

plugin/modify-env/modify-env.c
```diff
@@ -113,14 +113,21 @@ char * read_dmtcp_env_file(char *file, int size) {
...
-  if (strlen(dmtcp_get_ckpt_dir()) > sizeof(pathname)-1-sizeof(file)) {
-    warning(__FILE__ ": Pathname of ckpt dir is too long: ",
-            dmtcp_get_ckpt_dir() /* , "\n" */ );
-    exit(1);
+  // Fix for picky compiler.  The function dmtcp_get_ckpt_dir() would only
+  //   return NULL if we were not running under DMTCP.  But the compiler
+  //   doesn't know that.
+  const char *dmtcp_ckpt_dir = dmtcp_get_ckpt_dir();
+  if (dmtcp_ckpt_dir) {
+    if (strlen(dmtcp_ckpt_dir) + strlen(file) > sizeof(pathname)-1) {
```
karya0: This is still needed, for both master and 2.5

plugin/modify-env/modify-env.c
```diff
@@ -113,14 +113,21 @@ char * read_dmtcp_env_file(char *file, int size) {
...
+      fprintf(stderr, "%s:%d : Pathname of ckpt dir is too long: %s\n",
+              __FILE__, __LINE__, dmtcp_ckpt_dir);
+      exit(1);
+    }
+    // This can not overwrite the pathname buffer due to the if condition.
+    strcpy(pathname, dmtcp_ckpt_dir);
```
karya0: Let's replace strcpy with strncpy.

plugin/modify-env/modify-env.c
```diff
@@ -113,14 +113,21 @@ char * read_dmtcp_env_file(char *file, int size) {
...
+    strcpy(pathname + strlen(dmtcp_ckpt_dir), "/");
```
karya0: Let's consider strncat here and below.

src/execwrappers.cpp
```diff
@@ -232,7 +232,7 @@ int daemon(int nochdir, int noclose)
   }
 
   if (!nochdir) {
-    (void)chdir("/");
+    int dummy = chdir("/");
```
karya0: Consider JASSERT for suppressing the warning.

rohgarg:
This patch is no longer needed. PR #457 supercedes this one.

Issue #29, opened 2014-12-25
timer_create with SIGEV_THREAD causes restart to fail


karya0:
Libc handles SIGEV_THREAD by creating a helper thread and uses the tid of the helper thread along with SIGEV_THREAD_ID when creating the timer (i.e., libc emulates SIGEV_THREAD).

Thus, on timer expiration, the kernel signals the helper thread which in turn spawns a new user thread that calls sigev_notify_function as asked by the user when creating the timer (with SIGEV_THREAD).

The problems happens on restart, when libc tries to use the pre-ckpt helper-tid which is invalid now.  Hence, any call to timer_create with SIGEV_THREAD fails with EINVAL.

karya0:
Fixed in 374bd885.

Issue #28, opened 2014-12-11
Restore zombie state on restart so that the parent can get the correct status.


karya0:
If a child process exits but has not been waited by the parent before checkpoint, on restart, wait() will fail.

A potential solution is to collect status of all zombie processes during checkpoint and return them on restart when requested by the parent (through wait, waitpid, etc.).

Issue #27, opened 2014-12-11
Nameservice socket to coordinator invalid after restart.


karya0:
Description as received from @jiajuncao 

In coordinatorAPI.cpp, we have 2 sockets: _coordinatorSocket and _nsSock. When DMTCP is not in running state, _coordinatorSocket is used for the publish/subscribe service (_nsSock and _coordinatorSocket are the same). To support publish/subscribe during running state, an extra socket is created (_nsSock = createNewSocketToCoordinator(COORD_ANY)). The IB plugin takes use of this feature after a restart: it needs to subscribe information about the remote memory region key on running. The first restart is OK. But when we try to checkpoint after the first restart and then restart the second time: _nsSock is a bad file descriptor, it remains the same fd as the one created after last restart, but it's not valid anymore. I think it's either closed or not successfully restored. 

Here is a simple plugin doing just publish/subscribe. It's a stand-alone version based on example-db. You can test it by running dmtcp1. 
Again, the test procedure is ckpt, restart, ckpt, restart. It'll fail after the second restart.

``` C++
#include <stdlib.h>
#include <sys/types.h>
#include <unistd.h>
#include "dmtcp.h"

struct keyPid {
  int key;
  pid_t pid;
} mystruct, mystruct_other;

static int is_restart = 0;
uint32_t sizeofPid;

void dmtcp_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
{

  /* NOTE:  See warning in plugin/README about calls to printf here. */
  switch (event) {
  case DMTCP_EVENT_RESTART:
    is_restart = 1;
    break;
  case DMTCP_EVENT_REGISTER_NAME_SERVICE_DATA:
    if (is_restart) {
      mystruct.key = 1;
      mystruct.pid = getpid();
      dmtcp_send_key_val_pair_to_coordinator("ex-db",
                                             &(mystruct.key),
                                             sizeof(mystruct.key),
                                             &(mystruct.pid),
                                             sizeof(mystruct.pid));
    }
    break;
  default:
    break;
  }
  DMTCP_NEXT_EVENT_HOOK(event, data);
}

unsigned int sleep(unsigned int seconds)
{
  unsigned int result = NEXT_FNC(sleep)(seconds);
  if (is_restart) {
    sizeofPid = sizeof(mystruct_other.pid);
    mystruct_other.key = 1;
    dmtcp_send_query_to_coordinator("ex-db",
                                    &(mystruct_other.key),
[...]

Issue #25, opened 2014-11-17
Move apache plugin from test/plugin to contrib


karya0:
Shouldn't apache plugin be inside contrib directory?

PR #422, opened 2016-07-08
pid-plugin: Use pre-allocated buffer for mutex map

rohgarg:
The pid plugin uses the `mapMutexVirtTid` STL map to track mutexes used
by an application. This is required to patch the libc internal mutex
structs on restart. The problem with the current approach is that calls
to allocate and initiate the map (using the new operator) can lead to
a deadlock when an application defines uses a custom memory allocator
library that ends up calling other wrappers. The fix is to force the
allocation of the map from the DMTCP alloc arena.

Here's an excerpt from a stacktrace demonstrating the bug:

    #15 pthread_mutex_lock ()
    #16 dmtcp::ConnectionList::_lock_tbl ()
    #17 dmtcp::ConnectionList::add ()
    #18 dmtcp::FileConnList::processFileConnection ()
    #19 _open_open64_work ()
    #20 open (path=0x5809fc "/dev/zero", flags=<optimized out>)
    ...
    #23 operator new (req=48)
    #24 mapMutexVirtTid ()
    #25 pthread_mutex_lock ()
    #26 lock_threads ()
    #27 dmtcp::ThreadList::getNewThread ()
    #28 dmtcp::ThreadList::init ()
    #29 dmtcp_initialize ()

Although, this patch fixes the problem for now, a proper fix would be
to change the DMTCP internal locks to use custom locking primitives.

### DIFF
--- a/src/plugin/pid/pid.cpp
+++ b/src/plugin/pid/pid.cpp
@@ -36,11 +36,12 @@ using namespace dmtcp;
 extern "C" pid_t dmtcp_update_ppid();
 
 static string pidMapFile;
-map<pthread_mutex_t*, pid_t>& mapMutexVirtTid()
+dmtcp::map<pthread_mutex_t*, pid_t>& mapMutexVirtTid()
 {
-  static map<pthread_mutex_t*, pid_t> *instance = NULL;
+  static dmtcp::map<pthread_mutex_t*, pid_t> *instance = NULL;
   if (instance == NULL) {
-    instance = new map<pthread_mutex_t*, pid_t>();
+    void *buffer = JALLOC_MALLOC(1024*1024);
+    instance = new (buffer) dmtcp::map<pthread_mutex_t*, pid_t>();
   }
   return *instance;
 }


src/plugin/pid/pid.cpp
```diff
@@ -36,11 +36,12 @@ using namespace dmtcp;
...
 static string pidMapFile;
-map<pthread_mutex_t*, pid_t>& mapMutexVirtTid()
+dmtcp::map<pthread_mutex_t*, pid_t>& mapMutexVirtTid()
 {
-  static map<pthread_mutex_t*, pid_t> *instance = NULL;
+  static dmtcp::map<pthread_mutex_t*, pid_t> *instance = NULL;
   if (instance == NULL) {
-    instance = new map<pthread_mutex_t*, pid_t>();
+    void *buffer = JALLOC_MALLOC(1024*1024);
+    instance = new (buffer) dmtcp::map<pthread_mutex_t*, pid_t>();
```
jiajuncao: Is this equal to
`instance = reinterpret_cast<dmtcp::map<pthread_mutex_t*, pid_t> *> (buffer)` ?
I'm no C++ expert, and that's why it takes me a while to understand what it does. @karya0 should be the judge here.

rohgarg: > Is this equal to ... ?

No, the two are not equal.

Placement new forces the object to be allocated in the pre-allocated region and calls the constructor. See https://isocpp.org/wiki/faq/dtors#placement-new

PR #427, opened 2016-07-25
Fix misc. issues pointed out by static analysis

rohgarg:


rohgarg:
The issues and the fixes should be obvious in most places. In places where the code's intent wasn't clear to me, I have added comments.

PR #428, opened 2016-07-25
Fix misc. issues pointed out by static analysis v2

rohgarg:
The issues and the fixes should be obvious in most places. In places where the code's intent wasn't clear to me, I have added comments.


rohgarg:
@karya0: There seems to be an error with the Travis CI/Github integration. Do you know how to resolve this?

PR #429, opened 2016-07-26
Plugins for DMTCP 3.0 don't require config.h now

gc00:
User plugins in version 3.0 required `#include config.h`.  This was a problem for two reasons:
1.  It would seem weird to the user.  The reason is that the new interface was using PACKAGE_VERSION, a configure variable from autoconf.
2.  `config.h` was defining lots of macros like PACKAGE_VERSION in the user's namespace.  This is bad, since the user might have defined their own macro, PACKAGE_VERSION.

The solution was to rename `include/dmtcp.h` to `include/dmtcp.h.in`, to add to it
`#define DMTCP_PACKAGE_VERSION "@PACKAGE_VERSION@"`
and then modify `configure` and `configure.ac` to auto-generate `include/dmtcp.h`.

I tested this on 3.0.  I assume we don't need this for 2.5, since 2.5 doesn't use PACKAGE_VERSION or `config.h` for plugins.

### DIFF
--- a/configure
+++ b/configure
@@ -4930,7 +4930,7 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
 ac_config_headers="$ac_config_headers include/config.h"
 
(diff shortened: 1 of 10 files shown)


rohgarg:
LGTM. Please remove the config.h from the contrib plugins.

gc00:
@rohgarg, I had to push in an updated PR that includes the commits from master that had been made in the interim.  Now, github wants this to be reviewed again, but there should be no essential changes.

Could you review it once more?  Thanks.

PR #432, opened 2016-09-09
fixup! Fix misc. issues pointed out by static analysis

rohgarg:
A few more fixes:
 - Fix a case of use-after-free in `getCoordHostAndPort()`
 - Fix a case of use of incorrect function for freeing up resources (allocated by `popen()`)

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -78,24 +78,24 @@ static int _cachedPort = 0;
 
 void
 CoordinatorAPI::getCoordHostAndPort(CoordinatorMode mode,
-                                    const char **host,
+                                    string &host,
                                     int *port)
 {
   if (SharedData::initialized()) {
-    *host = SharedData::coordHost().c_str();
+    host = SharedData::coordHost();
     *port = SharedData::coordPort();
     return;
   }
 
   if (_firstTime) {
     // Set host to cmd line (if --cord-host) or env var or DEFAULT_HOST
-    if (*host == NULL) {
+    if (host == "") {
       if (getenv(ENV_VAR_NAME_HOST)) {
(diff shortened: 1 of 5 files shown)


src/coordinatorapi.cpp
```diff
@@ -78,24 +78,24 @@ static int _cachedPort = 0;
 
 void
 CoordinatorAPI::getCoordHostAndPort(CoordinatorMode mode,
-                                    const char **host,
+                                    string &host,
                                     int *port)
 {
   if (SharedData::initialized()) {
-    *host = SharedData::coordHost().c_str();
```
rohgarg: This is a use-after-free error.

`getCoordHostAndPort()` gets a copy of the string returned by `coordHost()`. We save a pointer to this string in `*host`. However, the string object is de-allocated as soon as we return from `getCoordHostAndPort()`.

src/dmtcp_restart.cpp
```diff
@@ -437,12 +437,14 @@ runMtcpRestart(int is32bitElf, int fd, ProcessInfo *pInfo)
...
         rc = read(debugPipe[0], &currentDebugLevel, sizeof(currentDebugLevel));
         if (rc < 0) break;
-        rc = write(debugPipe[0], &requestedDebugLevel, sizeof(currentDebugLevel));
+        rc = write(debugPipe[0], &requestedDebugLevel,
+                   sizeof(currentDebugLevel));
         if (rc < 0) break;
       } while (currentDebugLevel != requestedDebugLevel);
       if (rc < 0) {
         JASSERT(false)
-            .Text("Unable to set up debug connection with the restarted process");
```
rohgarg: Same as above.

rohgarg:
@gc00 looked at this and agreed.

PR #433, opened 2016-09-12
fixup! delayresume: Add new plugin

rohgarg:
This patch brings in the changes to the configure script to ensure that a makefile is generated for the delayresume plugin (introduced in commit: 80c49ff).


karya0:
LGTM

rohgarg:
Merged.

Issue #342, opened 2016-03-04
DMTCP fails to compile with Intel C compiler


rohgarg:
It seems like we accidentally deleted Intel compiler specific code at some point. As a result, `mtcp_restart.c` fails to compile with `icc`. Here's the error message:

``` bash
icc -I../../include -I.  -g -O2 -c -o mtcp_restart.o -DPIC -fPIC -fno-stack-protector -g -O0 mtcp_restart.c
icc -g -O2  -o ../../bin/mtcp_restart -fPIC -g -O0 -nodefaultlibs mtcp_restart.o
mtcp_restart.o: In function `restart_fast_path':
src/mtcp/mtcp_restart.c:389: undefined reference to `memfence'
```

Note that this affects both the master branch and the 2.5 branch.

rohgarg:
Fixed via PRs #435 and #436.

PR #435, opened 2016-09-15
Fix mtcp_restart for Intel C compiler

rohgarg:
This patch fixes issue #342. I have tested this with ICC-16.0.3.
There were two issues:
- Missing definition for the memfence() function; and
- Incorrect offset for the restorememoryareas_fptr in the rinfo
  struct.

The second bug was introduced in the commit: dec2c26 when new fields
were added to the rinfo struct.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -98,6 +98,8 @@ typedef struct RestoreInfo {
   VA vvarStart;
   VA vvarEnd;
   fnptr_t post_restart;
+  // NOTE: Update the offset when adding fields to the RestoreInfo struct
+  // See note below in the restart_fast_path() function.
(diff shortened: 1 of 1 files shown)


gc00:
In src/dmtcpplugin.cpp, I see this code:

```
#if defined(__i386__) || defined(__x86_64__)
static inline void memfence(){  asm volatile ("mfence" ::: "memory"); }
#elif defined(__arm__)
static inline void memfence(){  asm volatile ("dmb" ::: "memory"); }
#elif defined(__aarch64__)
# include "membarrier.h"
static inline void memfence(){  RMB; WMB; }
#else
# define memfence() __sync_synchronize()
#endif
```

Should we try to reuse some of this, or at least include a comment in this file pointing to dmtcpplugin.cpp:memfence(), which handles multiple architectures?

PR #441, opened 2016-10-21
Fixed the case where two QPs are not connected.

jiajuncao:
The original code assumes that for RC connections, two QPs are always connected, which is not true in certain cases (the bug was found in Manuel's cluster using Mvapich2 2.2).

### DIFF
--- a/contrib/infiniband/ibv_internal.h
+++ b/contrib/infiniband/ibv_internal.h
@@ -118,8 +118,13 @@ struct internal_ibv_qp {
   struct ibv_qp user_qp;
   uint64_t magic1;
   uint64_t magic2;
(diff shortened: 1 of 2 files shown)


gc00:
@rohgarg, would you mind pushing in this PR #441 for master/3.0, 2.5, and 2.4?  It has already been approved.  @jiajuncao is away, and I'd like to get more of the PR's with bug fixes ready for the next release.  (Note that there is a conflict to resolve, but I assume that it will be easy to resolve.)  Thanks.

gc00:
@rohgarg or @jiajuncao, Could you please push in this PR for all three branches, as indicated?  We'll need this PR as part of the upcoming 2.5 release, in particular.  Thanks.

PR #347, opened 2016-03-12
Fix dlsym_default to not rely on dlsym and dlopen

wmvsilva:
The old DLSYM_DEFAULT implementation works exactly as expected
when given an actual handle for a library. However, potential
problems arose with the use of the RTLD_DEFAULT and RTLD_NEXT
pseudo-handles. In these cases, the function relied on calls
to dlopen and dlsym which could cause an infinite loop if called
before the complete initialization of libc.
This commit replaces the usage of dlopen and dlsym with dladdr1 which
simplifies the process greatly and should fix this problem.

Please let me know if there is anything I should change before this is merged.

PR #443, opened 2016-11-11
Fix dlsym_default to not rely on dlsym and dlopen [duplicate of PR #347]

rohgarg:
The old DLSYM_DEFAULT implementation works exactly as expected
when given an actual handle for a library. However, potential
problems arose with the use of the RTLD_DEFAULT and RTLD_NEXT
pseudo-handles. In these cases, the function relied on calls
to dlopen and dlsym which could cause an infinite loop if called
before the complete initialization of libc.
This commit replaces the usage of dlopen and dlsym with dladdr1 which
simplifies the process greatly and should fix this problem.

### DIFF
--- a/include/dlsym_default.h
+++ b/include/dlsym_default.h
@@ -32,73 +32,36 @@
 # undef __USE_GNU
 #endif
 
-/************************************************************************
- * IMPORTANT CAVEATS:
- *   DLSYM_DEFAULT() is effective when called from a library, but not when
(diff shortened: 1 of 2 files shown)


rohgarg:
@wmvsilva I have created this PR that includes your patch and some minor cosmetic changes. I'm planning to merge this to the master branch and then a few more fixes on top of this.

Issue #411, opened 2016-06-21
Segfault when using ltdl


kwharrigan:
I started a mailing list topic on this w/r to ImageMagick's "convert" crashing, which I've reduced down to a trivial test case involving use of libtool / ltdl.  Apologize for the delay in getting this submitted. 

@rohgarg suggested running with disable-alloc-plugin, and that works, so presumably the issue is in there somewhere. 

Have a look at https://github.com/kwharrigan/dl_vs_ltdl.   You can run `make dmtcp_test` and should get a segfault when trying to run the ./ltdl under dmtcp_launch

rohgarg:
@kwharrigan: Sorry about the extremely long delay on this. I think we have a solution to this problem now. The PR #347 and a few other patches on top of that will address this issue.

rohgarg:
Fixed via PRs #444 and #456.

PR #444, opened 2016-11-11
Clean up dlsym_default

rohgarg:
Fixes issues with the latest libc and also issue #411

### DIFF
--- a/doc/dmtcp_dlsym.txt
+++ b/doc/dmtcp_dlsym.txt
@@ -30,7 +30,7 @@ which will return the _default version_, which is presumably the
 desired behavior for most users calling this.
 
 These notes will help to understand the concepts and background for
-the implementations of src/dlsym_default.c.
+the implementations of src/dmtcp_dlsym.cpp.
 
 SUGGESTED READING:
   http://refspecs.linuxfoundation.org/LSB_3.2.0/LSB-Core-generic/LSB-Core-generic/symversion.html#SYMRESOLUTION
@@ -68,7 +68,7 @@ BACKGROUND DETAILS OF THE ELF LINKER:
 
 The rest of this note explains how the macro DLSYM_DEFAULT()
 (and the function dlsym_default_internal()) find the default symbol.
-This is done in src/dlsym_default.c .
+This is done in src/dmtcp_dlsym.cpp.
 
 The versym ELF section defines the symbol version strings of that file (and it
 states if a symbol version depends on a different one).  The versym ELF
--- a/include/dlsym_default.h
+++ b/include/dlsym_default.h
@@ -22,6 +22,8 @@
 #include <pthread.h>
 #include <stdio.h>
 
+#include "dmtcp.h"
+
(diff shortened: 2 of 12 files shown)


gc00:
@rohgarg, Thanks for the discussion.  Here is my memory of the conclusions that we reached, plus one additional suggestion.

1.  Minor style suggestion:  change "!version" to "version == NULL" (easier to read the code's intention).

2.  As you suggested, let's change "dlsym_default" to "dmtcp_dlsym" everywhere (including where it appears as a prefix, such as in "dlvsym_default" -> "dmtcp_dlvsym").

3.  I think there was a third thing that we agreed upon, but if so, I've forgotten it now.

And finally, one extra suggestion.  In comments above the definition of "dlsym_default", let's add a comment like:
// This implementation tries to mimic the behavior of the linking-loader, as opposed to dlsym().
// In particular, if no versioned symbol exists, then the standard symbol is returned.
// If more than one versioned symbol exists, and all but one have the hidden bit set,
//  then the version without the hidden bit is returned.
// If only one versioned symbol exists, then it is returned whether the hidden bit is set or not.
//   From examples in various libraries, it seems that when only one versioned symbol exists,
//   it has the hidden bit set.
// If two or more versions of the symbol exist, and the hidden bit is set in all cases, then
//    the newest version is returned.
// If two or more versions of the symbol exist in which the hidden bit is not set, then
//    the behavior is undefined.  [ OR DO WE HAVE A FIXED BEHAVIOR HERE? I THINK THIS CASE DOESN'T OCCUR IN THE ACTUAL LIBRARIES. ]
// If the unversioned symbol and the versioned symbol both exist, then the versioned
//   symbol is preferred.  [LET'S CHECK THE CORRECTNESS OF THIS LAST RULE.]
// Note that dlsym() in libdl.so seems to follow the unusual rule of ignoring the hidden bit,
//   and choosing a somewhat arbtrary version that is often the oldest version.

gc00:
@rohgarg,  This looks almost ready to go  Two minor things that we talked about are:
1.  git mv doc/dlsym_default.txt doc/dmtcp_dlsym.txt
    [ And maybe do the same edit, dlsym_default->dmtcp_dlsym inside the .txt file. ]
2.  Add some kind of comment in front of the definition of dmtcp_dlsym, such as the example that I gave in the previous comment.  We need a comment to explain what dmtcp_dlsym does.

gc00:
I see in the above comments. that I was wrong about doc/dlsym_default.txt.  It was renamed.  But dlsym_default.h has not been changed.  So, here is my updated list of TODOs.  We're almost there.

So, we still have:
1.  git mv dlsym_default.h dmtcp_dlsym.h
2.Edit, dlsym_default->dmtcp_dlsym inside the doc/dlsym_default.txt file.
3.  Add some kind of comment in front of the definition of dmtcp_dlsym, such as the example that I gave in the previous comment. We need a comment to explain what dmtcp_dlsym does.

PR #449, opened 2016-11-16
Improved logging mechanism

rohgarg:
This PR adds a new option: `--debug-logs` to `dmtcp_launch` ,
`dmtcp_command`, and `dmtcp_restart`. Using this option a user can specify
a comma-separated list of plugins. For the plugins that are specified,
debug logs will be displayed. Note the logs can be enabled independently
at checkpoint time, or at restart time, or at runtime.

### DIFF
--- a/include/shareddata.h
+++ b/include/shareddata.h
@@ -122,6 +122,8 @@ struct Header {
   uint32_t numIncomingConMaps;
   uint32_t numInodeConnIdMaps;
 
+  uint32_t logMask;
+
   union {
     struct BarrierInfo barrierInfo;
     char pad[128];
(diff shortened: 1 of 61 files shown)


gc00:
@rohgarg, Thanks for helping with the joint review.  The comments that I remember are that:

1.  We should use colon-separated lists (similar to how PATH or LD_PRELOAD is specified), rather than a comma-separated list.

2.  We discussed an alternative implementation of the functionality of FOREACH_LOGSOURCE, using a different macro, resulting in:  `mask |= ENUM_LOGSOURCE_TO_MASK(logSrc)` 

3.  In `enum logSource`, let's change `JTRACE` to `ALWAYS`, and update other uses of this in the code.

4.  Above the code for `enum logSource`, let's add a `FIXME` comment with something like:
// In 3.0, we should extend the new style of plugin so that `JTRACE()` can infer the plugin by calling something like `get_my_name()`, a libprivate function in each plugin library.  However, as 3.0 evolves, the plugins will be combined into one big library, and so this design will have to be revisited.

gc00:
@rohgarg, Just pinging.  When this is ready, I can review it face to face, if that will make it faster to do the fixes and push this in.

PR #450, opened 2016-11-22
pathvirt-plugin: Fix several corner cases.

jiajuncao:
The corner cases include empty paths, and paths that share the
same prefix do not match, e.g., /dir1 and /dir12


jiajuncao:
Diffs updated.

Issue #453, opened 2016-12-02
dmtcp restart fails to work under PBSPro


subhasisb:
dmtcp restart code attempts to increase the stack limit max to the ARCH_MAX ie, 256 MB. This fails to succeed when dmtcp is run under PBS (mom). PBS sets the stack max to whatever the system ulimit max was at startup (ie, value of ulimit -s). If the ulimit -s was unlimited, then pbs_mom leaves it as unlimited and that is the only case when dmtcp restart code works (since it is then able to increase the stack max to 256 MB). 

The code block where dmtcp attempts to set the stack max could be removed. It could simply inherit whatever the current max limits are from the environment. Attempting to set the stack max does not make much sense anyway, since if it was limited by the parent (to lower than 256 MB), the call will ALWAYS fail. And if it was not limited then its already large and dmtcp would just work.

subhasisb:
The code fix has been merged in. Closing the ticket.

PR #454, opened 2016-12-02
Remove code that attempts to set stack max to 256 MB at restart

subhasisb:
dmtcp restart code attempts to increase the stack limit max to the ARCH_MAX ie, 256 MB. This fails to succeed when dmtcp is run under PBS (mom). PBS sets the stack max to whatever the system ulimit max was at startup (ie, value of ulimit -s). If the ulimit -s was unlimited, then pbs_mom leaves it as unlimited and that is the only case when dmtcp restart code works (since it is then able to increase the stack max to 256 MB).

The code block where dmtcp attempts to set the stack max could be removed. It could simply inherit whatever the current max limits are from the environment. Attempting to set the stack max does not make much sense anyway, since if it was limited by the parent (to lower than 256 MB), the call will ALWAYS fail. And if it was not limited then its already large and dmtcp would just work.

This fixes the issue #453

### DIFF
--- a/src/mtcp/mtcp_check_vdso.ic
+++ b/src/mtcp/mtcp_check_vdso.ic
@@ -305,7 +305,6 @@ void mtcp_check_vdso(char **environ)
       int i = mtcp_sys_readlink("/proc/self/exe", runtime, PATH_MAX);
       if ( i != -1)
       { char *argv[MAX_ARGS+1];
-        struct rlimit rlim;
 
         /* "make" has the capability to raise RLIMIT_STACK to infinity.
          * This is a problem.  When the kernel (2.6.24 or later) detects this,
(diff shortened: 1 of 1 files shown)


gc00:
LGTM.  Let's push this in.  Some info for your interest follows below.

@rohgarg and I also discussed this, and it looks good to me.  I assume that you did `make check` on this and it passed.

The back story is that there was a time when we needed to force the kernel to produce a different memory layou in mtcp_restart in order to avoid address conflicts with the text and stack segments found in the checkpoint image from the user's a.out.  But we now have a better design, and this old technique is no longer needed.  (Now we leave a hole in the address space in the checkpoint image, and pass the address of that hole to mtcp_restart, so that mtcp_restart can relocate its own text segment into that hole and then map in the memory from the checkpoint image.

subhasisb:
hi @gc00 yes, I did a make check and all the tests passed

PR #455, opened 2016-12-02
ssh-plugin: Minor refactoring of the code

rohgarg:


src/plugin/ipc/ssh/ssh.cpp
```diff
@@ -486,18 +486,24 @@ static void updateCoordHost() {
   freeaddrinfo(result);
 }
 
+/*
+ * Side-effect: Modifies the global isRshProcess variable
+ */
+static bool isRshOrSshProcess(const char *filename)
+{
+  isRshProcess = (jalib::Filesystem::BaseName(filename) == "rsh");
```
gc00: This is a nice refactoring.  But shouldn't you go one step further as follows?
```
  static bool isSshProcess = ((jalib::Filesystem::BaseName(filename) == "ssh");
  return isSshProcess || isRshProcess;
```
Of course, this depends on this function always being called with the same `filename` argumnet.  But the `isRshProcess` logic seems to rely on that same assumption.  We should quickly check that assumption, and then use this cleaner factorization if it looks good.

I propose that this "more radical" re-factoring go into 3.0, but not into 2.5.0.

If you agree, then it looks good to me.

PR #457, opened 2016-12-09
Fix misc compilation errors

rohgarg:
Misc. patches from PR #437 for 2.5


include/dmtcp.h
```diff
@@ -182,13 +182,13 @@ EXTERNC const char* dmtcp_get_tmpdir(void);
 
 EXTERNC const char* dmtcp_get_ckpt_dir(void) __attribute ((weak));
 #define dmtcp_get_ckpt_dir() \
- (dmtcp_get_ckpt_dir ? dmtcp_get_ckpt_dir() : NULL)
+ (dmtcp_get_ckpt_dir ? dmtcp_get_ckpt_dir() : "")
```
rohgarg: Other users may be testing `dmtcp_get_ckpt_dir` for the NULL value. If we cast the function call to return `(char *)` (i.e., drop the const), then the else branch can again return NULL instead of the empty string.

src/mtcp/mtcp_restart.c
```diff
@@ -830,7 +830,7 @@ static int read_one_memory_area(int fd)
   mtcp_readfile(fd, &area, sizeof area);
   if (area.size == -1) return -1;
 
-  if (area.name && mtcp_strstr(area.name, "[heap]")
+  if (area.name[0] && mtcp_strstr(area.name, "[heap]")
```
rohgarg: Add `area.name &&` in the condition.

src/dmtcp_launch.cpp
```diff
@@ -35,7 +35,7 @@
 using namespace dmtcp;
 
 static void processArgs(int *orig_argc, char ***orig_argv,
-                        string *tmpDir_p, const char **host,
+                        string *tmpDir_p, string &host,
```
karya0: Shouldn't we be sending a pointer instead of a reference?

rohgarg: An earlier commit (df0fe3ff30109f16fa6e33915efab196f065b9c0) had changed the argument type to a string reference. I had missed out on fixing the argument type in the declaration. I noticed this issue when I compiled with `-Wall`.

PR #460, opened 2016-12-16
Comment JLOG call, fails with --enable-debug

gc00:


rohgarg:
This fixes issue #332 for DMTCP master and a different run-time bug for 2.5.

PR #467, opened 2016-12-18
file-plugin: Dump info. about ckpted open files

rohgarg:
This patch adds an ability to the file plugin to dump the information
about the open files that DMTCP saved in the last checkpoint. The open
files for each process in a computation are saved in its own separate
subdirectory. In each of these subdirectories, DMTCP will now also
write a text file that has all the meta-data about the files saved in
the subdirectory. This file would follow the format described below and
would be overwritten on every checkpoint.

fd-info.txt:
  filename1: /original/path/to/filename1
  filename2: /original/path/to/filename2
  filename3: /original/path/to/filename3

The file can be parsed easily by a program or manually.


gc00:
There are two minor changes.  It's optional.  You decide if you want to do them.
1. Change spelling of `buff` to the more conventional `buf`.
2. Test return value of `writeall()` to see if the return value is not the same as the size.

PR #468, opened 2016-12-19
protected-fds: Add ability to set protected fd base at runtime

rohgarg:


include/protectedfds.h
```diff
@@ -22,31 +22,72 @@
...
+#include <stdint.h>
+#include <stdlib.h>
+
+#define DEFAULT_PROTECTED_FD_BASE   820
+
+uint32_t protectedFdBase();
+
+inline uint32_t protectedFdBase()
+{
+  static uint32_t base = DEFAULT_PROTECTED_FD_BASE;
```
gc00: Let's set `DEFAULT_PROTECTGED_BASE` to either `(uint)-1` or `1<<30`, so as to create an immediate error if anyone tries to use this.  This might be more future-proof than setting it to the old `820`.

gc00:
The code is nice.  In addition, can we check for the case that the user created a large `fd` at launch, and the `fd` is invalid on the restart machine, which has a smaller MAX limit.  A simple implementation is to add one more field to the DMTCP header for the ckpt image:  `max_user_fd`.  We can then check this inside `dmtcp_restart` to see if it's compatible with the max limit on the restart machine.

gc00:
@rohgarg, I've also added a "3.0" label to this PR.  I assume that's appropriate, but please change it if it's not.

When the PR is ready, let me know, and we can quickly approve it.

rohgarg:
@gc00: This is only intended for the 2.5 branch. @karya0 and I had discussed this change, and we believe we can move to a completely different approach for 3.0.

gc00:
@karya0 and @gc00 believed there was a better approach, to be implemented for 3.0.  But it was never implemented, and we hit the same bug in 3.0/master.  So, we're now porting this to 3.0/master in PR #825, in order to have common code.  Any later improvements can be made to both branches, since they will now be in sync.

Issue #458, opened 2016-12-12
DMTCP multiarch broken out of the box


aparashk:
I have tried many combinations of Linux distributions (Redhat 5, Ubuntu 12.04, Ubuntu 14.04) and DMTCP versions (2.4.0, 2.4.3, 2.4.5, 2.5.0-rc2) and all fail for me in the same way. When I build a multiarch version checkpointing a 32-bit binary process creates a checkpoint, but restoring that causes a crash (SIGSEGV). The sequence of commands I have been using for establishing this are as follows:

```
./configure --enable-m32
make clean
make
./configure
make clean
make
make M32=1 check
```

Note that the single platform (64-bit) version of DMTCP passes the test:

```
make check
```

without problems.

Here is an example of what I get, on Ubuntu 14.04 LTS 64-bit install, when building from source and running a random 32-bit application under DMTCP:

```
# ./configure --enable-m32
[… lots of output, no errors…]
# make clean
[… lots of output, no errors …]
# make
[… lots of output, no errors …]
# ./configure
[… lots of output, no errors …]
# make clean
[… lots of output, no errors …]
# make
[… lots of output, no errors …]
# bin/dmtcp_launch <i386-program>
ERROR: ld.so: object '/work/dmtcp-2.5.0-rc2/lib/dmtcp/libdmtcp_alloc.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/work/dmtcp-2.5.0-rc2/lib/dmtcp/libdmtcp_dl.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/work/dmtcp-2.5.0-rc2/lib/dmtcp/libdmtcp_ipc.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/work/dmtcp-2.5.0-rc2/lib/dmtcp/libdmtcp_svipc.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/work/dmtcp-2.5.0-rc2/lib/dmtcp/libdmtcp_timer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
[...]

PR #469, opened 2016-12-21
VDSO fix for i386.

karya0:
Apparently, the second vvar page is not readable and so a memcpy() results in a BUS error. This PR tries to hack through it by ignoring the second vvar page.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -850,7 +850,14 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
       mtcp_abort();
     }
     MTCP_ASSERT(vvar == vvarStart);
-    mtcp_memcpy(vvarStart, rinfo->vvarStart, vvarEnd - vvarStart);
+    // On i386, only the first page is readable. Reading beyond that
+    // results in a bus error.
+    // Arguably, this is a bug in the kernel, since /proc/*/maps indicates
+    // that both pages of vvar memory have read permission.
+    // This issue arose due to a change in the Linux kernel pproximately in
+    // version 4.0
+    // TODO: Find a way to automatically detect the readable bytes.
+    mtcp_memcpy(vvarStart, rinfo->vvarStart, MTCP_PAGE_SIZE);
 #endif /* if defined(__i386__) */
   }
 }


rohgarg:
LGTM modulo some testing. This should help fix issue #458. @aparashk, could you please verify if this works for you?

src/mtcp/mtcp_restart.c
```diff
@@ -850,7 +850,10 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
       mtcp_abort();
     }
     MTCP_ASSERT(vvar == vvarStart);
-    mtcp_memcpy(vvarStart, rinfo->vvarStart, vvarEnd - vvarStart);
+    // On i386, only the first 4096 bytes are readable. Reading beyond that
+    // results in a bus error.
+    // TODO: Find a way to automatically detect the readable bytes.
+    mtcp_memcpy(vvarStart, rinfo->vvarStart, 4096);//vvarEnd - vvarStart);
```
rohgarg: Is this specific to some kernel version? If yes, should we wrap this an ifdef?

src/mtcp/mtcp_restart.c
```diff
@@ -850,7 +850,10 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
       mtcp_abort();
     }
     MTCP_ASSERT(vvar == vvarStart);
-    mtcp_memcpy(vvarStart, rinfo->vvarStart, vvarEnd - vvarStart);
+    // On i386, only the first 4096 bytes are readable. Reading beyond that
+    // results in a bus error.
```
gc00: Let's also add here:
` *     // This issue arose due to a change in the Linux kernel pproximately in version 4.0`
The rationale is that we will then know that this doesn't matter in some future case where a customer/user has not upgraded from an older kernel for the sake of robustness.

src/mtcp/mtcp_restart.c
gc00: In addition, let's change `4096` to `getpagesize()`.  This should be more future-proof.  The man page says:
> The function getpagesize() returns  ...where "page" is a fixed-length block, the unit for memory allocation and file mapping performed by mmap(2).

Note that we will then need to do:
`#define _DEFAULT_SOURCE`
`#include <unistd.h>`
at the head of the file.  The man page says `_DEFAULT_SOURCE` is valid for glibc-2.19 and beyond, which should handle the recent kernels that cause this bug.

Also, in answer to @rohgarg, I believe this is a kernel bug.  We've seen a similar bug where the kernel removes the write permission for the last page of a memory segment, even though `/proc/*/maps` says that the entire memory segment has read permission.

This also makes me wonder if there is a different fix for this bug.  We could use:
 `mprotect(vvarStart, rinfo->vvarEnd - vvarStart, PROT_READ`
@karya0, could you try out this alternative fix (mprotect instead of modifying mmap)?  If this works, we don't need `getpagesize()`, although I would still keep the additional comments that I'm proposing.

gc00:
Let's try the `mprotect` option also, mentioned in the inline comment.  After that, we can briefly review this again and push it in.

As noted in the inline comment, this seems to be about a kernel bug where `/proc/*/maps` says both vvar pages are readable, but the last page is not really readable.  We've seen this same kernel bug setting before in other places, and added a hack there to mprotect to `PROT_READ` again to _really_ get read permission.  See the comment in `processinfo.cpp:ProcessInfo::growStack()` concerning the same issue (arguably a kernel bug), as seen for the stack when using Matlab applications.

src/mtcp/mtcp_restart.c
rohgarg: @gc00, @karya0: I tried the `mprotect()` approach. It still fails on the memcpy. For reference here's what I tried:
```C
    MTCP_PRINTF("Before mprotect");
    mtcp_sys_mprotect(vvarStart, vvarEnd - vvarStart, PROT_READ);
    MTCP_PRINTF("After mprotect");
    mtcp_memcpy(vvarStart, rinfo->vvarStart, vvarEnd - vvarStart);
    //mtcp_memcpy(vvarStart, rinfo->vvarStart, 4096);//vvarEnd - vvarStart);
```

gc00:
@rohgarg reports that the `mprotect()` strategy does _not_ fix the bug.  So, all that remains is to replace the 4096 with `sysconf(_SC_PAGESIZE)`, and to then add the proposed comments.  Both of those changes are addressed in the inline icomments above.

src/mtcp/mtcp_restart.c
aparashk: I confirm that I still experience bus errors with some 32-bit binaries when multiarch is used.

gc00:
@karya0, there are two minor fixes remaining before this can be pushed in:

1. `4096` to `sysconf(_SC_PAGESIZE)`
2. adding the C comments described in the above inline comments.

karya0:
We are using `MTCP_PAGE_SIZE` everywhere else in `src/mtcp/*`, so I have used that instead of `sysconf(_SC_PAGESIZE)`.

gc00:
We could consider changing the `#define MTCP_PAGE_SIZE` to use `sysconf(...)`.  But it's a small issue.  Can we at least add a `JWARNING` (or `JASSERT`?) somewhere in the code to verify that `MTCP_PAGE_SIZE == sysconf(_SC_PAGE_SIZE)`?  That could save us some tedious debugging in the future.

But it's a small point, and so if you want to push this in without the `JWARNING`, I can go with that.

Otherwise, LGTM.

karya0:
@gc00: Could you approve this PR? I'll then merge it in.

For page size, here is the comment from #480:
  It's mtcp_restart, so we don't really have sysconf available for use (it's a libc call). I'll merge it as is.

PR #470, opened 2016-12-21
Change static object to static object pointer.

jiajuncao:
This is a temp fix for the case where dmtcp static local objects
are destroyed before application's static objects. For the long
term, we need to consider patching the correct order of calling
constructors/destructors.


gc00:
LGTM.  Let's ship it to both 2.5 and 3.0, as you propose.  For 3.0, github will ask for a review before shipping.  So, usually we create a second pull request for 3.0, and anybody except the submitter can then quickly review it so that it can also be pushed into 3.0.

In addition, could you also open an issue about the case of static local and global C++ objects that are initialized by constructors?  Please have the issue refer back to this PR so that we'll have the two textually linked together.  Thanks.

PR #472, opened 2016-12-22
Avoid recursive call to shmdt() wrapper.

jiajuncao:
This is a temp fix for OpenMPI 2.0. For the long term, we
need to think about the case where user code modifies its
own global offset table.


src/miscwrappers.cpp
```diff
@@ -463,7 +464,13 @@ extern "C" long syscall(long sys_num, ... )
     case SYS_shmdt:
     {
       SYSCALL_GET_ARG(const void*,shmaddr);
-      ret = shmdt(shmaddr);
+      if (dmtcp_svipc_inside_shmdt != NULL &&
+          dmtcp_svipc_inside_shmdt()) {
+        ret = _real_syscall(SYS_shmdt, shmaddr);
+      }
```
rohgarg: Let's change this to:
```C
if (...) {
  ...
} else {
  ...
}
```

src/plugin/pid/pid_miscwrappers.cpp
```diff
@@ -449,7 +449,13 @@ extern "C" long syscall(long sys_num, ...)
     case SYS_shmdt:
     {
       SYSCALL_GET_ARG(const void*,shmaddr);
-      ret = shmdt(shmaddr);
+      if (dmtcp_svipc_inside_shmdt != NULL &&
+          dmtcp_svipc_inside_shmdt()) {
+        ret = _real_syscall(SYS_shmdt, shmaddr);
+      }
+      else {
```
rohgarg: See above.

src/plugin/svipc/sysvipcwrappers.cpp
```diff
@@ -108,14 +122,22 @@ void *shmat(int shmid, const void *shmaddr, int shmflg)
 }
 
 extern "C"
+bool dmtcp_svipc_inside_shmdt()
```
rohgarg: Change this to:
```C
EXTERNC bool
dmtcp_svipc_inside_shmdt()
{
  ...
}
```

src/plugin/svipc/sysvipcwrappers.cpp
```diff
@@ -38,6 +38,20 @@
...
 
 static struct timespec ts_100ms = {0, 100 * 1000 * 1000};
+
+/*
+ * In OpenMPI 2.0, shmdt() is intercepted by modifying libraries' global offset
+ * table, meaning that _real_shmdt() will be redirected into OpenMPI's hook
+ * function, instead of libc's shmdt(). The hook function finally calls syscall()
+ * with the corresponding syscall number. The inside_shmdt variable indicates
+ * if the code is inside our shmdt() wrapper. If so, our syscall wrapper simply
+ * calls _real_syscall(), instead of calling the shmdt() wrapper, resulting in
```
rohgarg: Could you please rephrase this sentence? It reads as if calling `_real_syscall()` results in an infinite recursion.

src/plugin/svipc/sysvipcwrappers.cpp
jiajuncao: The rest of this file uses `extern "C"`. Do I need to make a special case for this function?

jiajuncao:
@rohgarg, diffs are updated in place.

To answer your general question: if we call `_real_syscall()` from the `shmdt()` wrapper, we may fall into the `syscall()` wrapper again. We can use your approach if we have the mechanism to call directly into libc.

src/plugin/svipc/sysvipcwrappers.cpp
rohgarg: I'd say let's use EXTERNC. This will allow us to use this function from a C file as well.

jiajuncao: Done.

gc00:
@jiajuncao, I'm just noting for the record what we discussed.  If I am reading github correctly, this PR has now been pushed into the 2.5 branch, but not yet the master (3.0) branch.  We'll want to push this also into master.

And then after that, we should open up an issue noting, as you wrote:
> This is a temp fix for OpenMPI 2.0. For the long term, we
> need to think about the case where user code modifies its
> own global offset table.

And finally, this PR and the issue should cross-reference each other.  Thanks.

gc00:
I removed the 3.0 label.  The PR for 3.0 is now part of PR #472.

PR #477, opened 2016-12-30
Improve coord logic 2.5

gc00:
This is the same as PR #238, but targeted toward the 2.5.0 branch.


gc00:
@rohgarg, I think this update of the PR satisfies the requested changes.  The flag --any-coordinator is no longer present in this PR, and this does _not_ use the new semantics of PR #238.  Specifically, if the user does not specify a port, or any other special arguments, then the old semantics is retained:  if there is a coordinator with a running computation at the default port, then the new process will join the existing computation.

src/coordinatorapi.cpp
```diff
@@ -448,10 +448,14 @@ void CoordinatorAPI::startNewCoordinator(CoordinatorMode mode)
...
                                                  port, 128);
   JASSERT(coordinatorListenerSocket.isValid())
-    (coordinatorListenerSocket.port()) (JASSERT_ERRNO)
-    .Text("Failed to create listen socket."
-          "\nIf msg is \"Address already in use\", this may be an old coordinator."
-          "\nKill other coordinators and try again in a minute or so.");
+    (coordinatorListenerSocket.port()) (JASSERT_ERRNO) (host) (port)
+    .Text("Failed to create socket to coordinator port."
+          "\nIf msg is \"Address already in use\","
+                                    " this may be an old coordinator."
```
karya0: Seems to be a weird alignment issue here.

src/restartscript.cpp
```diff
@@ -528,7 +530,8 @@ void writeScript(const string& ckptDir,
                   "        . $DMTCP_SRUN_HELPER_SYNCFILE\n"
                   "        pass_slurm_helper_contact \"$DMTCP_LAUNCH_CKPTS\"\n"
                   "        rm $DMTCP_SRUN_HELPER_SYNCFILE\n"
-                  "        dmtcp_restart --join --coord-host $DMTCP_COORD_HOST"
+                  "        dmtcp_restart --join-coordinator"
+                              " --coord-host $DMTCP_COORD_HOST"
```
karya0: Can you fix the alignment here and below too?

src/coordinatorapi.cpp
```diff
@@ -546,6 +546,17 @@ DmtcpMessage CoordinatorAPI::sendRecvHandshake(DmtcpMessage msg,
       .Text("Connection rejected by the coordinator.\n"
             " Reason: This process has a different computation group.");
   }
+  // Coordinator also prints this, but its stderr may go to /dev/null
+  if (msg.type == DMT_REJECT_NOT_RESTARTING) {
+    string coordinatorHost = ""; // C++ magic code; "" to be invisibly replaced
+    int coordinatorPort;
+    Util::getCoordHostAndPort(COORD_ANY, coordinatorHost, &coordinatorPort);
+    JNOTE ("\n\n*** Computation not in RESTARTING or CHECKPOINTED state."
+        "\n***Reject incoming computation process requesting restart."
```
karya0: Can you reword this comment? Maybe something like:

"Can't join the existing coordinator as it is serving a different computation. Consider launching a new coordinator."

src/restartscript.cpp
gc00: In this case, it doesn't make sense to change the alignment.  The issue is that the second new line is a continuation of the first new line.  So, the only (bad) options are:

1.  Shift the second line all the way to the leftto line up.  The " --coord-host .." will stick out in the left margin.  Or,

2.  Shift the second line all the way to the left to line up, and add extra spaces to the "--coord-host" appears underneath "dmtcp_restart".  But in this case, the C++ source code looks good.  But the script that is generated will have lots of spaces in the middle of that line in the script.

gc00:
For the last inline comment, see my reply, above, about why the change in alignment doesn't make sense here.  I've now addressed the other inline comments.

src/restartscript.cpp
karya0: Let's leave it as if for 2.5. For master, we can do the following:
```
"        dmtcp_restart --join-coordinator \\\n"
"          --coord-host $DMTCP_COORD_HOST \\\n"
"          --coord-port $DMTCP_COORD_PORT \\\n"
"          $DMTCP_LAUNCH_CKPTS\n"
```

This is more readable in the source as well as the script.

PR #480, opened 2016-12-31
VDSO fix for i386.

karya0:
Apparently, the second vvar page is not readable and so a memcpy()
results in a BUS error. This commit tries to hack through it by ignoring
the second vvar page.

This is a port of PR #469 from master to the 2.5 branch.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -777,7 +777,14 @@ static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
       mtcp_abort();
     }
     MTCP_ASSERT(vvar == vvarStart);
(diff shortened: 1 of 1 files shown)


gc00:
We could consider changing the `#define MTCP_PAGE_SIZE` to use `sysconf(...)`.  But it's a small issue.  Can we at least add a `JWARNING` (or `JASSERT`?) somewhere in the code to verify that `MTCP_PAGE_SIZE == sysconf(_SC_PAGE_SIZE)`?  That could save us some tedious debugging in the future.

But it's a small point, and so if you want to push this in without the `JWARNING`, I can go with that.

Otherwise, LGTM.

karya0:
It's `mtcp_restart`, so we don't really have `sysconf` available for use (it's a libc call). I'll merge it as is.

PR #482, opened 2016-12-31
Avoid recursive call to shmdt() wrapper (for master: 3.0)

jiajuncao:
Duplicate of #472


src/plugin/svipc/sysvipcwrappers.cpp
```diff
@@ -39,6 +39,21 @@ using namespace dmtcp;
 
 static struct timespec ts_100ms = { 0, 100 * 1000 * 1000 };
 
+/*
+ * In OpenMPI 2.0, shmdt() is intercepted by modifying libraries' global offset
```
gc00: OpenMPI -> Open MPI

gc00:
I have a minor comment:
  OpenMPI -> Open MPI
After that, LGTM.

PR #456, opened 2016-12-08
Fix dlsym default 2.5

rohgarg:
This is a port of PR #444 for the 2.5 branch. Fixes issues: #447 and #411.

PR #494, opened 2017-01-02
Fix dlsym default (for 2.4 branch)

gc00:
This is a port of PR #456 to the 2.4 branch (for the DMTCP-2.4.7 release).

This is needed for the 2.4 branch since in recent versions of GLIBC, libdl/dlerror calls libc/malloc, creating an infinite recursion when we interpose using libdl/dlsym.  With this PR, in most cases, we now use our own dlsym implementation.

### DIFF
--- a/doc/dmtcp_dlsym.txt
+++ b/doc/dmtcp_dlsym.txt
@@ -30,7 +30,7 @@ which will return the _default version_, which is presumably the
 desired behavior for most users calling this.
 
 These notes will help to understand the concepts and background for
-the implementations of src/dlsym_default.c.
+the implementations of src/dmtcp_dlsym.cpp.
 
 SUGGESTED READING:
   http://refspecs.linuxfoundation.org/LSB_3.2.0/LSB-Core-generic/LSB-Core-generic/symversion.html#SYMRESOLUTION
@@ -68,7 +68,7 @@ BACKGROUND DETAILS OF THE ELF LINKER:
 
 The rest of this note explains how the macro DLSYM_DEFAULT()
 (and the function dlsym_default_internal()) find the default symbol.
-This is done in src/dlsym_default.c .
+This is done in src/dmtcp_dlsym.cpp.
 
 The versym ELF section defines the symbol version strings of that file (and it
 states if a symbol version depends on a different one).  The versym ELF
--- a/include/dlsym_default.h
+++ b/include/dlsym_default.h
@@ -22,6 +22,8 @@
 #include <stdio.h>
 #include <pthread.h>
 
+#include "dmtcp.h"
+
 #ifndef __USE_GNU
 # define __USE_GNU_NOT_SET
 # define __USE_GNU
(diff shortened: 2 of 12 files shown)


gc00:
Currently, Travis is failing on `dmtcp_launch test/dmtcp1` (just for launching).  It complains that the following symbol was not found:
```
c++filt _Z24dmtcp_ProcName_EventHook11eDmtcpEventP17_DmtcpEventData_t
dmtcp_ProcName_EventHook(eDmtcpEvent, _DmtcpEventData_t*)
```

Using `nm`, it seems that this symbol is defined in dmtcpworker.cpp.  It should have been using `DmtcpEvent_t' instead of `eDmtcpEvent`.  The symbol `DmtcpEvent_t' is just a typedef of `enum eDmtcpEvent`.  So, I'm currently guessing that this is due to the special semantics of C++.  I'm still tracking this down.

gc00:
I found it.  In `src/Makefile.in`, `am___d_libdir__libdmtcp_so_OBJECTS` does not include `procname.o` for some weird reason.  It should be a member.  Probably this is some weird bug involving automake:  `Makefile.am` and `Makefile.in`.  I'll search for the exact cause later.

gc00:
In order to back-port this from the 2.5 branch to the 2.4 branch, I had to add a new commit:
> Fix dmtcp_dlvsym: call it in GET_LIBPTHREAD_...

This commit changes a `dlvsym` to a `dmtcp_dlvsym`.  Could this line have been forgotten in the original code in the PR #456?

However, when I testing the above hypothesis by cherry-picking this new commit onto PR #456, I find that the `file2' test fails in the 2.5 branch with this new commit.  However, the 2.4 branch requires this new commit -- even in order to pass the `dmtcp1` test.  This needs further analysis.

sakshi-rg:
@gc00:

I have now identified the reason for the issue you are seeing with 2.4:
the reason is the commit: 87295a8 , which is missing from the 2.4
branch.

When your commit:

 >Fix dmtcp_dlvsym: call it in GET_LIBPTHREAD_ ...

is removed from the 2.4 branch, it results in a SEGFAULT, when running any
program under DMTCP. The reason for the SEGFAULT is a stack overflow because
of the following infinite recursion.

    ...
    initialize_libpthread_wrappers ()
    dmtcp_prepare_wrappers ()
    calloc ()
    _dlerror_run ()
    initialize_libpthread_wrappers ()
    dmtcp_prepare_wrappers ()
    calloc ()
    _dlerror_run ()
    __dlvsym ()
    initialize_libpthread_wrappers ()
    dmtcp_prepare_wrappers ()
    malloc ()
    (anonymous namespace)::pool::pool ()
    __static_initialization_and_destruction_0 ()
    _GLOBAL__sub_I_eh_alloc.cc(void) ()
    call_init ()
    call_init ()

While in theory we should have your "extra" commit for the 2.5 and the
3.0 branches as well, the infinite recursion was never observed with the
two branches because the way we initialize wrappers had changed slightly.
In particular, the function: `dmtcp_initialize()` that acts as the new single
point of entry or initialization has a check at its very beginning:
```C
  dmtcp_initialize()
  {
    static bool initialized = false;
    if (initialized) {
      return;
    }
    initialized = true;
    ...
  }
```
that prevents the function from being called recursively. Also, I'm *unable*
to reproduce the issue with the file2 test after adding your commit to the
2.5 branch. I checked on my workstation and on Dekaksi.

rohgarg:
@gc00:

I have now identified the reason for the issue you are seeing with 2.4:
the reason is the commit: 87295a8 , which is missing from the 2.4
branch.

When your commit:

Fix dmtcp_dlvsym: call it in GET_LIBPTHREAD_ ...
is removed from the 2.4 branch, it results in a SEGFAULT, when running any
program under DMTCP. The reason for the SEGFAULT is a stack overflow because
of the following infinite recursion.

...
initialize_libpthread_wrappers ()
dmtcp_prepare_wrappers ()
calloc ()
_dlerror_run ()
initialize_libpthread_wrappers ()
dmtcp_prepare_wrappers ()
calloc ()
_dlerror_run ()
__dlvsym ()
initialize_libpthread_wrappers ()
dmtcp_prepare_wrappers ()
malloc ()
(anonymous namespace)::pool::pool ()
__static_initialization_and_destruction_0 ()
_GLOBAL__sub_I_eh_alloc.cc(void) ()
call_init ()
call_init ()
While in theory we should have your "extra" commit for the 2.5 and the
3.0 branches as well, the infinite recursion was never observed with the
two branches because the way we initialize wrappers had changed slightly.
In particular, the function: dmtcp_initialize() that acts as the new single
point of entry or initialization has a check at its very beginning:

  dmtcp_initialize()
  {
    static bool initialized = false;
    if (initialized) {
      return;
    }
    initialized = true;
    ...
  }
that prevents the function from being called recursively. Also, I'm unable
to reproduce the issue with the file2 test after adding your commit to the
2.5 branch. I checked on my workstation and on Dekaksi. I can also look at
the file2 test with you later today to see if we can identify the issue.

PR #502, opened 2017-01-04
Fixed tls_pid_offset for newer (>2.24) glibc.

karya0:
More information here: 
https://sourceware.org/git/gitweb.cgi?p=glibc.git;a=commit;h=c579f48edba88380635ab98cb612030e3ed8691e


src/mtcp/restore_libc.c
```diff
@@ -477,12 +490,22 @@ TLSInfo_SetThreadSysinfo(void *sysinfo)
...
+  ASSERT (major == 2);
+
   pid_t tls_pid, tls_tid;
   char *addr = (char *)get_tls_base_addr();
 
   tls_pid = *(pid_t *)(addr + TLSInfo_GetPidOffset());
   tls_tid = *(pid_t *)(addr + TLSInfo_GetTidOffset());
 
+  // For glibc > 2.24, pid field is unused.
+  if (tls_pid == 0 && minor >= 24) {
```
gc00: Don't you want the full condition `tls_pid == NULL && major == 2 && minor >= 24`, similarly to what you test for at line 283 above?  If we rely on the earlier `ASSERT (major == 2)`, then the condition to test here requires reading some code from elsewhere.

gc00:
@karya0, I have an optional inline comment.  But otherwise, this is ready to go.

PR #507, opened 2017-01-08
Enhance --enable-timing

rohgarg:
This patch allows DMTCP to dump additional information about barrier overhead, ckpt read and write times when configured with the `--enable-timing` flag. Each process of the computation writes out this information in a separate `timings.*.csv` file.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -53,6 +53,7 @@
 #include <unistd.h>
 
 #include "../membarrier.h"
+#include "config.h"
 #include "mtcp_check_vdso.ic"
 #include "mtcp_header.h"
 #include "mtcp_sys.h"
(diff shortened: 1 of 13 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -657,6 +664,14 @@ restorememoryareas(RestoreInfo *rinfo_ptr)
...
   DPRINTF("close cpfd %d\n", restore_info.fd);
   mtcp_sys_close(restore_info.fd);
+  double readTime = 0.0;
+#ifdef TIMING
+  struct timeval endValue;
+  mtcp_sys_gettimeofday(&endValue, NULL);
+  struct timeval diff;
+  timersub(&endValue, &restore_info.startValue, &diff);
+  readTime = diff.tv_sec + (diff.tv_usec / 1000000.0);
+#endif
```
gc00: This is not obligatory, but I think I see an opportunity for re-factoring here.  I'm guessing that the above pattern will happen a lot in our code.  Should we create a new function and call it inside the `ifdef`?

Issue #439, opened 2016-10-18
dmtcp with PBS scheduler


angel-smile:
Hi,

I'm wondering if anybody has got dmtcp work with PBS? I understand it is not a supported resource manger, but want to know if anybody has got luck. I have made similar set up as that for Slurm. The launch part seems to be ok, but resume is not working with either the scheduler, the generated dmtcp_restart_script.sh or the .dmtcp file. strace shows the restarted process is waiting for something:

futex(0x7fb31b83a600, FUTEX_WAIT_PRIVATE, 0, NULL

I'm using dmtcp 3.0.0.

Thanks!

rohgarg:
Could you try to attach gdb to the process that's stuck and share the backtrace? Is this an MPI application?

angel-smile:
Thank you for the response. It is not an MPI application, but one of the simplest dmtcp examples, dmtcp1.c. Here is the backtrace:

(gdb) bt
#0  0x00007f77309b3a00 in sem_wait () from /lib64/libpthread.so.0
#1  0x00007f7731e25258 in dmtcp::ThreadList::waitForAllRestored(Thread*) () at threadlist.cpp:643
#2  0x00007f7731e273c3 in stopthisthread(int) () at threadlist.cpp:598
#3  <signal handler called>
#4  0x00007f7731458bdd in nanosleep () from /lib64/libc.so.6
#5  0x00007f7731458a50 in sleep () from /lib64/libc.so.6
#6  0x0000000000400c11 in main (argc=-409088528, argv=0x7ffce79dcdf0) at dmtcp1.c:12

From what I can see, when I checkpoint and then restart without using PBS scheduler, the sockets opened by dmtcp_coordinator match those opened by the restarted application. However, when PBS is used, checkpointing runs fine, but the restarted application opens 1 socket more than dmtcp_coordinator does, so they don't match. Do you know what might be the reason?

Sockets opened by the restarted application:
$ ls -l /proc/9836/fd
821 -> socket:[58290]
834 -> socket:[58298]

Sockets opened by dmtcp_coordinator:
$ ls -l /proc/9833/fd
4 -> socket:[58261]
6 -> socket:[58291]

rohgarg:
The application can open any number of sockets; it has no relation to the number of sockets that the coordinator has. However, it sounds like the application process has a socket that it inherits from the PBS launcher process. It should be easy to recognize this socket and "blacklist" it to make sure that DMTCP doesn't try to drain it.

[...]

PR #510, opened 2017-01-09
fileconnlist: Mark /proc/pid/environ as pre-existing fd

rohgarg:
This fixes issue #439.

### DIFF
--- a/src/plugin/ipc/file/fileconnlist.cpp
+++ b/src/plugin/ipc/file/fileconnlist.cpp
@@ -419,6 +419,19 @@ FileConnList::scanForPreExisting()
                                  FileConnection::FILE_BATCH_QUEUE));
     } else if (fd <= 2) {
       add(fd, new StdioConnection(fd));
+    } else if (getenv("PBS_JOBID") &&
+               (Util::strStartsWith(device, "/proc") &&
+                Util::strEndsWith(device, "environ"))) {
+      /*
+       * This is a workaround for an issue seen with PBS at ANU-NCI.
+       *
+       * Application processes would inherit a "/proc/<pid>/environ"
+       * file-descriptor when launched under PBS. As far as I can tell,
+       * the file-descriptor is inherited from the PBS launcher process
+       * on the compute node. The workaround is to recognize this fd as
+       * a pre-existing device and ignore it for checkpoint-restart.
+       */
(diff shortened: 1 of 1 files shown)


rohgarg:
@angel-smile: Could you please test this patch (again)? This is a cleaned up version of what I had provided to you earlier.

@subhasisb: Could you please review this? This workaround is required for supporting PBS.

subhasisb:
@rohgarg the change itself looks good. One question though: Besides PBS, if any application happens to have a legitimate fd to /proc/<pid>/environ, how would checkpointing such an application be effected by this change?

rohgarg:
@subhasisb: Since the case only triggers when the `PBS_JOBID` env. var. is set, other applications should not be affected by this change. Also, if an application explicitly opens a fd to `/proc/<pid>/environ` during its runtime, the fd will go through the regular checkpoint-restart logic. This change only affects an application that's launched under PBS (with the `PBS_JOBID` env. var set) and that inherits the `/proc/<pid/environ` fd from its PBS launcher parent process.

subhasisb:
Ah yes, of course, you have the check for PBS_JOBID. So that covers it. A follow up question though. What is it a actual problem for dmtcp to checkpoint if the application launched under PBS inherits an fd? I mean, what we are doing here is fine code wise, but sounds like a workaround, no?

angel-smile:
@rohgarg Many thanks for it! I've tested with a bioinformatics application that it works for me under PBS with DMTCP v2.5 and v3.0.0.

gc00:
@subhasisb , The function `FileConnList::scanForPreExisting()` is intended to check for any file descriptors that were in existence prior to an application starting,  Typically, this happens because a monitor process opened a file descriptor and then forked and exec'ed into the actual application.  If DMTCP is checkpointing only the application and not the monitor process, then DMTCP has no way of knowing how to restore the fd of the monitor process during restart.  So, we punt, and we don't try to restore any pre-existing fd's.

@rohgarg, I'm approving this PR.  But I wonder if we should have an issue about better documenting all of these special cases of pre-existing fd's.  Or in general, we should document what we whiltelist and blacklist.  Do we already have an issue that covers that?

subhasisb:
Thanks for the explanation @gc00  - I understand now.

PR #512, opened 2017-01-10
DMTCP_RESTART_PAUSE=[1234] (for version 3.0)

gc00:
__I've tested this and it works now.  Please review.__

* Workflow:
>    env DMTCP_RESTART_PAUSE0=1 dmtcp_restart ckpt_*.dmtcp &
    # This stops in infinite loop in ThreadList::postRestartDebug(),
    #   which will call ThreadList::postRestart()
    # Before the infinite loop, it prints to the user to do:
    gdb PROGRAM_NAME PID
    # Full 'gdb attach' should now be working

NOTES:
* This includes postRestartDebug() as target from mtcp_restart,
  where postRestartDebug() has the infinite loop.
* In commit a8cfd6666, we had removed '-g' as default.  I added a print statement with the `gdb attach` advice that reminds the user to configure DMTCP with `--enable-debug

### DIFF
--- a/src/dmtcp_restart.cpp
+++ b/src/dmtcp_restart.cpp
@@ -488,13 +488,19 @@ runMtcpRestart(int is32bitElf, int fd, ProcessInfo *pInfo)
     }
   }
 
-  /* If DMTCP_RESTART_PAUSE0 set, mtcp_restart will loop until gdb attach.*/
+  /* If DMTCP_RESTART_PAUSE==1, mtcp_restart will loop until gdb attach.*/
   bool mtcp_restart_pause = false;
-  if (getenv("MTCP_RESTART_PAUSE0") || getenv("DMTCP_RESTART_PAUSE0")) {
+  char * pause_param = getenv("DMTCP_RESTART_PAUSE");
+  if (pause_param == NULL) {
+    pause_param = getenv("MTCP_RESTART_PAUSE");
+  }
+  if (pause_param != NULL && pause_param[0] == '1' && pause_param[1] == '\0') {
 #ifdef HAS_PR_SET_PTRACER
     prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY, 0, 0, 0); // For: gdb attach
 #endif // ifdef HAS_PR_SET_PTRACER
     mtcp_restart_pause = true;
+    // If mtcp_restart_pause == true, mtcp_restart will invoke
+    //     postRestartDebug() in the checkpoint image instead of postRestart().
   }
 
   char *const newArgs[] = {
--- a/src/dmtcp_restart.cpp
+++ b/src/dmtcp_restart.cpp
@@ -488,10 +488,21 @@ runMtcpRestart(int is32bitElf, int fd, ProcessInfo *pInfo)
     }
   }
 
+  /* If DMTCP_RESTART_PAUSE0 set, mtcp_restart will loop until gdb attach.*/
+  bool mtcp_restart_pause = false;
+  if (getenv("MTCP_RESTART_PAUSE0") || getenv("DMTCP_RESTART_PAUSE0")) {
+#ifdef HAS_PR_SET_PTRACER
+    prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY, 0, 0, 0); // For: gdb attach
+#endif // ifdef HAS_PR_SET_PTRACER
+    mtcp_restart_pause = true;
+  }
+
   char *const newArgs[] = {
     (char *)mtcprestart.c_str(),
     const_cast<char *>("--fd"), fdBuf,
     const_cast<char *>("--stderr-fd"), stderrFdBuf,
+    // This flag must be last, since it may become NULL
+    ( mtcp_restart_pause ? const_cast<char *>("--mtcp-restart-pause") : NULL ),
     NULL
   };
 
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -104,11 +104,13 @@ typedef struct RestoreInfo {
   VA vvarStart;
   VA vvarEnd;
   fnptr_t post_restart;
+  fnptr_t post_restart_debug;
   // NOTE: Update the offset when adding fields to the RestoreInfo struct
   // See note below in the restart_fast_path() function.
   fnptr_t restorememoryareas_fptr;
 
   // void (*post_restart)();
+  // void (*post_restart_debug)();
   // void (*restorememoryareas_fptr)();
   int use_gdb;
   int text_offset;
@@ -119,6 +121,7 @@ typedef struct RestoreInfo {
   struct timeval startValue;
 #endif
   MYINFO_GS_T myinfo_gs;
+  int mtcp_restart_pause;  // Used by env. var. DMTCP_RESTART_PAUSE0
 } RestoreInfo;
 static RestoreInfo rinfo;
 
@@ -235,17 +238,12 @@ main(int argc, char *argv[], char **environ)
 #endif /* ifdef ENABLE_VDSO_CHECK */
 
   rinfo.fd = -1;
+  rinfo.mtcp_restart_pause = 0; /* false */
   rinfo.use_gdb = 0;
   rinfo.text_offset = -1;
   shift;
   while (argc > 0) {
-    // Flags for standalone debugging
-    if (argc == 1) {
-      // We would use MTCP_PRINTF, but it's also for output of util/readdmtcp.sh
-      mtcp_printf("Considering '%s' as a ckpt image.\n", argv[0]);
-      ckptImage = argv[0];
-      break;
-    } else if (mtcp_strcmp(argv[0], "--use-gdb") == 0) {
+    if (mtcp_strcmp(argv[0], "--use-gdb") == 0) {
       rinfo.use_gdb = 1;
       shift;
     } else if (mtcp_strcmp(argv[0], "--text-offset") == 0) {
@@ -259,9 +257,17 @@ main(int argc, char *argv[], char **environ)
     } else if (mtcp_strcmp(argv[0], "--stderr-fd") == 0) {
       rinfo.stderr_fd = mtcp_strtol(argv[1]);
       shift; shift;
+    } else if (mtcp_strcmp(argv[0], "--mtcp-restart-pause") == 0) {
+      rinfo.mtcp_restart_pause = 1; /* true */
+      shift;
     } else if (mtcp_strcmp(argv[0], "--simulate") == 0) {
       simulate = 1;
       shift;
+    } else if (argc == 1) {
+      // We would use MTCP_PRINTF, but it's also for output of util/readdmtcp.sh
+      mtcp_printf("Considering '%s' as a ckpt image.\n", argv[0]);
+      ckptImage = argv[0];
+      break;
     } else {
       MTCP_PRINTF("MTCP Internal Error\n");
       return -1;
@@ -323,6 +329,7 @@ main(int argc, char *argv[], char **environ)
   rinfo.vvarStart = mtcpHdr.vvarStart;
   rinfo.vvarEnd = mtcpHdr.vvarEnd;
   rinfo.post_restart = mtcpHdr.post_restart;
+  rinfo.post_restart_debug = mtcpHdr.post_restart_debug;
   rinfo.motherofall_tls_info = mtcpHdr.motherofall_tls_info;
   rinfo.tls_pid_offset = mtcpHdr.tls_pid_offset;
   rinfo.tls_tid_offset = mtcpHdr.tls_tid_offset;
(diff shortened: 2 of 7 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -235,12 +238,18 @@ main(int argc, char *argv[], char **environ)
 #endif /* ifdef ENABLE_VDSO_CHECK */
 
   rinfo.fd = -1;
+  rinfo.mtcp_restart_pause = 0; /* false */
   rinfo.use_gdb = 0;
   rinfo.text_offset = -1;
   shift;
   while (argc > 0) {
+    // All parsing of flags with no arg must appear before parsing ckptImage
+    if (mtcp_strcmp(argv[0], "--mtcp-restart-pause") == 0) {
```
karya0: Please move it towards the if-else-if block.

karya0:
I think this can be simplified a bit. Instead of introducing a new env, why not use the existing `[D]MTCP_RESTART_PAUSE` and use the value `0`, `1`, `2`, etc., to reflect different modes (sleep, infinite loop, etc.)? That way, we can do everything at one place in the `postRestart` routine. It's also easier to read/maintain than having two env vars trying to do similar things.

gc00:
@karya0, It's a good point.  For 3.0, we can use [D]MTCP_RESTART_PAUSE and use the value 0, 1, 2.  For DMTCP-2.5.1, should we offer both possibilities (for backwards compatibility), or simply add DMTCP_RESTART_PAUSE0 to 2.5.1, and stop there?

karya0:
I would suggest that we add the flag but do not advertise it in the 2.5.1 branch. That way, we don't have to explicitly un-support it in 3.0.

gc00:
@karya0, If you mean to not advertise `DMTCP_RESTART_PAUSE0`, I think that's a mistake.  Right now it's painful to debug on restart, since you have to go back and generate a new ckpt image.  Debugging restart on the fly is a win and defininstely should be advertise.  (I don't believe in purposely crippling our software.  But maybe I misunderstood your point.)

Anyway, let's agree on `PAUSE0` for 2.5.1 now, so we can do the release.

For 3.0, shall we offer only `DMTCP_RESTART_PAUSE=[123]`, or also the three individual flags?

src/mtcp/mtcp_restart.c
gc00: @karya0, I think moving this to the end of the if-else would be a mistake.  It means that if we see `mtcp_restart <bunch of flags with args> --mtcp-restart-pause`, it currently reports `***MTCP Internal Error`.  In your revised ordering, it would report `***ERROR opening ckpt image (--mtcp-restart-pause); errno: No such file or directory` when it reaches lower in the code.  This type of thing can happen if DMTCP constructs an empty string for the checkpoint image file due to a logic error.

karya0: I see, our current logic is faulty. The `argc == 1` test should be moved near the last `else` statement. Otherwise, we will see the same error with `mtcp_restart <args> --simulate`, `mtcp_restart <args> --use-gdb`, and so on.

gc00: Nice point!  The `argc == 1` test should come after any flags with no arguments, and that's donce most easily if it comes last.  I'll move it to the end, and place thje flags with no arguments near the end.

rohgarg:
LGTM. Let's push this to 2.5. We can have a different patch for 3.0 as discussed.

gc00:
PR #558 has now pushed this in, for the 2.5 branch.  This PR pertains only to 3.0, and will be revised based on the discussion with @karya0.

src/mtcp/mtcp_restart.c
```diff
@@ -104,11 +104,13 @@ typedef struct RestoreInfo {
   VA vvarStart;
   VA vvarEnd;
   fnptr_t post_restart;
+  fnptr_t post_restart_debug;
```
rohgarg: Please update the offset of `restorememoryareas_fptr` in the `restart_fast_path()` function for the Intel compiler (`#if defined(__INTEL_COMPILER) ...`).

src/dmtcp_restart.cpp
```diff
@@ -471,10 +471,27 @@ runMtcpRestart(int is32bitElf, int fd, ProcessInfo *pInfo)
     }
   }
 
+  /* If DMTCP_RESTART_PAUSE==1, mtcp_restart will loop until gdb attach.*/
+  bool mtcp_restart_pause = false;
+  char * pause_param = getenv("DMTCP_RESTART_PAUSE");
+  if (pause_param == NULL) {
+    pause_param = getenv("MTCP_RESTART_PAUSE");
+  }
+  if (pause_param != NULL && pause_param[0] == '1' && pause_param[1] == '\0') {
```
rohgarg: Can't we just use `strncmp()` here?

src/threadlist.cpp
```diff
@@ -686,19 +687,59 @@ ThreadList::waitForAllRestored(Thread *thread)
...
   }
   Thread_RestoreSigState(thread);
+
+  if (thread == motherofall) {
+    /* If DMTCP_RESTART_PAUSE==4, sleep 15 seconds to allow gdb attach.*/
+    char * pause_param = getenv("DMTCP_RESTART_PAUSE");
+    if (pause_param == NULL) {
+      pause_param = getenv("MTCP_RESTART_PAUSE");
+    }
+    if (pause_param != NULL && pause_param[0] == '4' &&
```
rohgarg: `strncmp()`?

src/threadlist.cpp
```diff
@@ -686,19 +687,59 @@ ThreadList::waitForAllRestored(Thread *thread)
...
   sigset_t tmp;
 
-  /* If DMTCP_RESTART_PAUSE set, sleep 15 seconds and allow gdb attach. */
-  if (getenv("MTCP_RESTART_PAUSE") || getenv("DMTCP_RESTART_PAUSE")) {
+  /* If DMTCP_RESTART_PAUSE==2, sleep 15 seconds and allow gdb attach. */
+  char * pause_param = getenv("DMTCP_RESTART_PAUSE");
+  if (pause_param == NULL) {
+    pause_param = getenv("MTCP_RESTART_PAUSE");
+  }
+  if (pause_param != NULL && pause_param[0] == '2' && pause_param[1] == '\0') {
```
rohgarg: `strncmp()`?

src/threadlist.cpp
```diff
@@ -782,6 +823,27 @@ restarthread(void *threadv)
     TLSInfo_SetThreadSysinfo(saved_sysinfo);
   }
 
+  if (thread == motherofall) { // if this is a user thread
+    /* If DMTCP_RESTART_PAUSE==3, sleep 15 seconds to allow gdb attach.*/
+    char * pause_param = getenv("DMTCP_RESTART_PAUSE");
+    if (pause_param == NULL) {
+      pause_param = getenv("MTCP_RESTART_PAUSE");
+    }
+    if (pause_param != NULL && pause_param[0] == '3' &&
```
rohgarg: `strncmp()`?

src/threadlist.cpp
```diff
@@ -686,19 +687,59 @@ ThreadList::waitForAllRestored(Thread *thread)
...
+    char * pause_param = getenv("DMTCP_RESTART_PAUSE");
+    if (pause_param == NULL) {
+      pause_param = getenv("MTCP_RESTART_PAUSE");
+    }
+    if (pause_param != NULL && pause_param[0] == '4' &&
+        pause_param[1] == '\0') {
+#ifdef HAS_PR_SET_PTRACER
+      prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY, 0, 0, 0); // For: gdb attach
+#endif // ifdef HAS_PR_SET_PTRACER
+      struct timespec delay = { 15, 0 }; /* 15 seconds */
```
rohgarg: Should we move this entire block in the ifdef? It makes no sense to sleep if HAS_PR_SET_PTRACER is not set. We won't be able to attach gdb anyway.

src/threadlist.cpp
```diff
@@ -686,19 +687,59 @@ ThreadList::waitForAllRestored(Thread *thread)
...
+{ // Don't try to print before debugging.  Who knows what is working yet?
+  int dummy = 1;
+#ifndef DEBUG
+  // printf may fail, but we'll risk it to let user know this:
+  printf("\n** DMTCP: It appears DMTCP not configured with '--enable-debug'\n");
+  printf("**        If GDB doesn't show source, re-configure and re-compile\n");
+#endif
+  while (dummy);
+  // User should have done GDB attach if we're here.
+  prctl(PR_SET_PTRACER, 0, 0, 0, 0); // Revert permission to default: no ptracer
```
rohgarg: We need to check if HAS_PR_SET_PTRACER is set. I would suggest moving the entire block, including the infinite while loop in the ifdef. We won't be able to attach GDB anyway.

```C
#ifdef HAS_PR_SET_PTRACER
  int dummy = 1;
  ...
  while (dummy);
  prctl(PR_SET_PTRACER, 0, 0, 0, 0);
#endif
  postRestart();
```

src/dmtcp_restart.cpp
gc00: I suppose we could write `&& strncmp(pause_param, "1", 2)` instead of `&& pause_param[0] == '1' && pause_param[1] == '\0'`.  But I find the use of `strncmp()` less readable.  It seems to say to compare the first two characters of a string of length one: "1".  It works because of a trick based on knowing that a string of length one has an internal representation of length two: `{'1', '\0'}`.
Normally, people use `strncmp` when they want to use `strcmp` and are worried about running off the end of a large buffer.  We could avoid relying on knowledge of the internal string representation by writing `strcmp(pause_param, "1")` since we're not worried about the end of some buffer.  But I find the C primitives more intuitive for a string of length one.

src/threadlist.cpp
gc00: Same comment as above.  If we use `strncmp', then we should use the simpler `strcmp` since there is no safety issue.  But I prefer the C primitives.

src/threadlist.cpp
gc00: Same comment as above.  If we use `strncmp', then we should use the simpler `strcmp` since there is no safety issue.  But I prefer the C primitives.

src/mtcp/mtcp_restart.c
gc00: Could you be more explicit about the requested change?

src/threadlist.cpp
gc00: We may still be able to attach with gdb even if `HAS_PR_SET_PTRACER` has not been set.  It depends on things like if the user has done `echo 0 > /proc/sys/kernel/yama/ptrace_scope` as we did on our 16-core test machine, or if the user has used some other trick to get around it.  So, the delay for gdb attach is still valid.

src/threadlist.cpp
gc00: See the earlier comment: We may still be able to attach with gdb even if HAS_PR_SET_PTRACER has not been set.

src/threadlist.cpp
gc00: Same comment as above.  If we use `strncmp', then we should use the simpler `strcmp` since there is no safety issue.  But I prefer the C primitives.

src/mtcp/mtcp_restart.c
rohgarg: For the Intel compiler, we use inline assembly to compute the offset of `restorememoryareas_fptr` and jump into the function. See the code in `restart_fast_path()`:
```C
      asm volatile(
      // 96 = offsetof(RestoreInfo, rinfo.restorememoryareas_fptr)
      // NOTE: Update the offset when adding fields to the RestoreInfo struct
      "movq    96+rinfo(%%rip), %%rdx;" /* rinfo.restorememoryareas_fptr */
      "leaq    rinfo(%%rip), %%rdi;"    /* &rinfo */
      "movl    $0, %%eax;"
      "call    *%%rdx"
      : : );
```
However, because of the additional field `fnptr_t post_restart_debug;` before `fnptr_t restorememoryareas_fptr;` the offset of `restorememoryareas_fptr` within the struct will change, and so we need to update the code in `restart_fast_path()` to correspond.

Also, see commit: 7074b3a3, where we last fixed this bug for the Intel compiler.

gc00:
@rohgarg, I had never looked closely at your `--gdb` flag.  If there is an easy way to combine some of this, let's sit down together and do it.  If that would take more time, then I propose that we push this in, and open an issue about re-factoring the code to combine the two ideas.  I'd like to avoid cases where PRs are under review for a long time.

src/dmtcp_restart.cpp
rohgarg: Okay, I agree.

src/threadlist.cpp
rohgarg: Okay, I agree.

src/threadlist.cpp
rohgarg: Okay, I agree.

src/threadlist.cpp
rohgarg: Sure, but we still need to wrap the call to prctl in `#ifdef HAS_PR_SET_PTRACER`.

src/threadlist.cpp
rohgarg: Okay, I agree.

src/threadlist.cpp
rohgarg: Okay, I agree.

src/threadlist.cpp
gc00: @rohgarg, You're right!  Sorry that I missed that.  I'll push in a version that fixes this.  Thanks.

src/mtcp/mtcp_restart.c
rohgarg: Please see my latest PR #564. Apparently, we had missed this in the review of the patch that went into 2.5.

gc00: Thanks.  This is now fixed in this PR, and I've updated the PR to reflect that.

gc00:
@rohgarg, I think I've now addressed the inline comments.  The remaining issue is re-factoring to unify `--gdb` and `DMTCP_RESTART_PAUSE=0`.  Since that would have to be done for both 2.5 and 3.0, I propose that we do that as a separate PR.  We can create an issue as a reminder (or create an empty PR as a reminder), if you want to be sure that this happens.

PR #513, opened 2017-01-10
dmtcp_coordinator: Use higher resolution timer for coordTimeStamp

rohgarg:
An issue was observed where multiple parallel launches/restarts of independent computations on a single node would up using the incorrect shared-data files. The reason was that since a shared-data file's name is generated using the following tuple: _<hostid, virt. pid, coordinator timestamp>_, multiple independent coordinators would end up generating the same values when they were launched in parallel.

The fix is to use a nanosecond resolution timer for generating the coordinator timestamp. Although this doesn't fix the issue entirely, it does reduce the chances of two independently launched computations on a node mapping the same shared-data file.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -891,6 +891,20 @@ DmtcpCoordinator::processDmtUserCmd(DmtcpMessage &hello_remote,
   }
 }
 
+/*
+ * Returns the current timestamp with nanosecond resolution
+ */
+static uint64_t
+getCurrTimestamp()
+{
+  struct timespec value;
+  uint64_t nsecs = 0;
+  JASSERT(clock_gettime(CLOCK_MONOTONIC, &value) == 0);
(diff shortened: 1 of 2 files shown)


src/dmtcp_coordinator.cpp
```diff
@@ -891,14 +891,27 @@ DmtcpCoordinator::processDmtUserCmd(DmtcpMessage &hello_remote,
...
+/*
+ * Returns the current timestamp with nanosecond resolution
+ */
+static uint64_t
+getCurrTimestamp()
+{
+  struct timespec value;
+  uint64_t nsecs = 0;
+  JASSERT(clock_gettime(CLOCK_MONOTONIC, &value) == 0);
+  nsecs = value.tv_sec*100000000L + value.tv_nsec;
```
karya0: A missing `0`?

rohgarg: Good catch! I have updated the PR now and fixed a similar bug in `UniquePid::ThisProcess()` where we were using the same code.

PR #514, opened 2017-01-11
Reset the coordinator socket on restart.

jiajuncao:
This solves the issue where on restart there is a larger open file
descriptor limit.

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -160,6 +160,16 @@ void CoordinatorAPI::resetOnFork(CoordinatorAPI& coordAPI)
   instance()._nsSock.close();
 }
 
+/* Recompute the protected coordinator fd and reset the coordiantor socket.
+ *
+ * Used to handle the case where protectedFdBase() has changed, for example,
+ * on restart.
+ * */
+void CoordinatorAPI::resetCoordSocketFd()
+{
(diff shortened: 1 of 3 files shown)


src/coordinatorapi.cpp
```diff
@@ -160,6 +160,13 @@ void CoordinatorAPI::resetOnFork(CoordinatorAPI& coordAPI)
   instance()._nsSock.close();
 }
 
+void CoordinatorAPI::resetOnRestart()
+{
+  _coordinatorSocket = jalib::JSocket(PROTECTED_COORD_FD);
+  JASSERT(_coordinatorSocket.isValid()) (JASSERT_ERRNO)
```
rohgarg: I have three main comments about this function:

 - I think the assert is not necessary. `isValid()` checks if the fd is >=0, which is trivially true after the code on the previous line. You can continue to use the assert, if you like, but please fix the associated text. This is not a listen socket. :-)

 - Could you please add a comment explaining the purpose of this function?

 - How about renaming the function to: `resetCoordSocketFd()`? I feel like that's more precise.

src/threadlist.cpp
```diff
@@ -683,6 +683,8 @@ void ThreadList::postRestart(void)
 {
   Thread *thread;
   sigset_t tmp;
+  Util::setProtectedFdBase();
```
rohgarg: Could you please write a comment explaining why we need this call here?

PR #515, opened 2017-01-11
coordinatorapi: Replace select() with poll()

rohgarg:


gc00:
I've added the 2.4.8 branch.  That branch also has mostly 'poll', and is missing this case.

PR #478, opened 2016-12-30
Support for rsh in DMTCP-3.0

gc00:
This is a port of PR #296 to master (DMTCP-3.0).

PR #518, opened 2017-01-13
Fixed SSH support.

karya0:
This PR fixes the SSH functionality broken by #478.

### DIFF
--- a/src/plugin/ipc/ssh/dmtcp_sshd.cpp
+++ b/src/plugin/ipc/ssh/dmtcp_sshd.cpp
@@ -105,31 +105,47 @@ dummySshdProcess(char *listenAddr)
 int main(int argc, char *argv[], char *envp[])
 {
   int in[2], out[2], err[2];
-  char *host;
-  int port;
+  char *host = NULL;
+  int port = 0;
 
   if (argc < 2) {
     printf("***ERROR: This program shouldn't be used directly.\n");
     exit(1);
   }
 
-  if (strcmp(argv[1], "--listenAddr") == 0) {
-    dummySshdProcess(argv[2]);
-    printf("ERROR: Not Implemented\n");
-    assert(0);
+  shift;
+  while (true) {
+    if (strcmp(argv[0], "--listenAddr") == 0) {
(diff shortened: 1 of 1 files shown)


rohgarg:
@karya0: What is it that we broke with #478?

karya0:
The argument parsing for dmtcp_sshd was not ported over correctly from 2.x. Check this:
https://github.com/dmtcp/dmtcp/pull/478/files#diff-efac9aa16eae3f242c03e30abf588260R172

Either it should have been `execvp(argv[6], &argv[6]);` or we should have used the `shift; shift;` trick.

karya0:
It also indicates that we are not testing `ssh` in travis :(.

src/plugin/ipc/ssh/dmtcp_sshd.cpp
```diff
@@ -105,31 +105,47 @@ dummySshdProcess(char *listenAddr)
...
+      shift;
+    } else {
+      break;
+    }
+  }
+
+  if (host == NULL) {
     printf("Missing --host argument");
-    assert(0);
+    exit(1);
```
gc00: `exit(1)` -> `exit(DMTCP_FAIL_RC)`

src/plugin/ipc/ssh/dmtcp_sshd.cpp
```diff
@@ -105,31 +105,47 @@ dummySshdProcess(char *listenAddr)
...
   }
-  host = argv[2];
 
-  if (strcmp(argv[3], "--port") != 0) {
+  if (port == 0) {
     printf("Missing --port argument");
-    exit(0);
+    exit(1);
```
gc00: `exit(1)` -> `exit(DMTCP_FAIL_RC)`

gc00:
I have two inline fixes of the form: exit(1) -> exit(DMTCP_FAIL_RC), Otherwise, LGTM.

I assume that the 2.5 branch is correct, and the fixes are only needed for 3.0. I'll put in a small PR for the 2.5 branch for adjusting the call to exit(1).

PR #519, opened 2017-01-13
IB: minor fix of blocking call wrappers.

jiajuncao:


contrib/infiniband/ibvctx.c
```diff
@@ -1116,16 +1116,26 @@ int _get_cq_event(struct ibv_comp_channel * channel,
   flags = fcntl(internal_channel->real_channel->fd, F_GETFL, NULL);
 
   // We need to change the call to non-blocking mode
-  if (flags & O_NONBLOCK == 0) {
+  // FIXME: we need to take care of the case where fd is already
+  // changed to non-blocking mode by the user.
+  if (!(flags & O_NONBLOCK)) {
+    rslt = fcntl(internal_channel->real_channel->fd,
```
rohgarg: Should you be using `_real_fcntl()` here?

contrib/infiniband/ibvctx.c
```diff
@@ -1116,16 +1116,26 @@ int _get_cq_event(struct ibv_comp_channel * channel,
...
+                 F_SETFL, flags | O_NONBLOCK);
+    if (rslt < 0) {
+      fprintf(stderr, "Failed to change file descriptor "
```
rohgarg: I thought you had some plugin-internal printf function for the IB plugin. Can we use that here?

contrib/infiniband/ibvctx.c
```diff
@@ -1116,16 +1116,26 @@ int _get_cq_event(struct ibv_comp_channel * channel,
   flags = fcntl(internal_channel->real_channel->fd, F_GETFL, NULL);
 
   // We need to change the call to non-blocking mode
-  if (flags & O_NONBLOCK == 0) {
+  // FIXME: we need to take care of the case where fd is already
+  // changed to non-blocking mode by the user.
```
rohgarg: > we need to take care of the case where fd is already changed to non-blocking mode by the user.

Why?

contrib/infiniband/ibvctx.c
```diff
@@ -1167,16 +1177,24 @@ int _get_async_event(struct ibv_context * ctx, struct ibv_async_event * event)
   int flags = fcntl(internal_ctx->real_ctx->async_fd, F_GETFL, NULL);
 
   // We need to change the call to non-blocking mode
-  if (flags & O_NONBLOCK == 0) {
+  if (!(flags & O_NONBLOCK)) {
+    rslt  = fcntl(internal_ctx->real_ctx->async_fd,
```
rohgarg: Same as above.

contrib/infiniband/ibvctx.c
```diff
@@ -1167,16 +1177,24 @@ int _get_async_event(struct ibv_context * ctx, struct ibv_async_event * event)
...
+                  F_SETFL, flags | O_NONBLOCK);
+    if (rslt < 0) {
+      fprintf(stderr, "Failed to change file descriptor "
```
rohgarg: Same as above.

contrib/infiniband/ibvctx.c
rohgarg: Could you also print the errno if it fails? It'll be useful for debugging.

rohgarg:
@jiajuncao, @gc00: Should we push this to the master branch and the 2.4 branch as well?

contrib/infiniband/ibvctx.c
jiajuncao: The user might be doing the same thing, i.e., polling then call the IB function. In that case, we may run into the infinite poll() loop.

contrib/infiniband/ibvctx.c
jiajuncao: The internal print is mainly used to print some debug info. Here it indicates an error.

jiajuncao:
I'd suggest push this into 2.5.1 and 3.0.  The non-blocking change wasn't introduced into 2.4.

contrib/infiniband/ibvctx.c
```diff
@@ -1113,21 +1113,32 @@ int _get_cq_event(struct ibv_comp_channel * channel,
...
+  flags = NEXT_FNC(fcntl)(internal_channel->real_channel->fd,
+                          F_GETFL, NULL);
 
   // We need to change the call to non-blocking mode
-  if (flags & O_NONBLOCK == 0) {
+  // FIXME: we need to take care of the case where fd is already
+  // changed to non-blocking mode by the user.
+  if (!(flags & O_NONBLOCK)) {
+    rslt = NEXT_FNC(fcntl)(internal_channel->real_channel->fd,
+                 F_SETFL, flags | O_NONBLOCK);
```
rohgarg: Could you please fix the indentation here? Align it with the first argument on the previous line.

contrib/infiniband/ibvctx.c
```diff
@@ -1113,21 +1113,32 @@ int _get_cq_event(struct ibv_comp_channel * channel,
...
+    if (rslt < 0) {
+      fprintf(stderr, "Failed to change file descriptor "
```
rohgarg: Can we use your DPRINTF here?

contrib/infiniband/ibvctx.c
```diff
@@ -1164,21 +1175,30 @@ int _get_async_event(struct ibv_context * ctx, struct ibv_async_event * event)
...
 
-  int flags = fcntl(internal_ctx->real_ctx->async_fd, F_GETFL, NULL);
+  int flags = NEXT_FNC(fcntl)(internal_ctx->real_ctx->async_fd,
+                              F_GETFL, NULL);
 
   // We need to change the call to non-blocking mode
-  if (flags & O_NONBLOCK == 0) {
+  if (!(flags & O_NONBLOCK)) {
+    rslt  = NEXT_FNC(fcntl)(internal_ctx->real_ctx->async_fd,
+                  F_SETFL, flags | O_NONBLOCK);
```
rohgarg: Fix indentation.

contrib/infiniband/ibvctx.c
```diff
@@ -1164,21 +1175,30 @@ int _get_async_event(struct ibv_context * ctx, struct ibv_async_event * event)
...
+    if (rslt < 0) {
+      fprintf(stderr, "Failed to change file descriptor "
```
rohgarg: DPRINTF?

contrib/infiniband/ibvctx.c
rohgarg: I see. So are you saying that if the user had set the fd to non-blocking mode, we shouldn't be calling `poll()`?

rohgarg:
> I'd suggest pushing this into 2.5.1 and 3.0

Okay, I have added the 3.0 label to this PR. Are there any other changes that need to go into 3.0 along with this patch? Or have they been pushed in already?

contrib/infiniband/ibvctx.c
jiajuncao: Right. But it's hard to tell when the user decides to change the mode. If we can assume the user does it before any call to the function, then I can add a static variable to indicate the mode before the first call to the function.

contrib/infiniband/ibvctx.c
jiajuncao: The internal PDEBUG prints only when the debug mode is on. I'd like this message to be printed even without the debug mode.

PR #520, opened 2017-01-13
Fix autotest.py: new coord if old one fails

gc00:
Fix long-standing bug that appears when an autotest test fails.  (joint credit w/ @rohgarg)


rohgarg:
It seems like this bug was *not* a long-standing bug. This was introduced in 5b75b050 on the master branch. It *does not* affect the other branches. :-)

PR #525, opened 2017-01-15
mtcp_restart: Remove assert introduced in 72df6746

rohgarg:
The brk call fails on Debian for bash because of alignment conflict
between heap and mtcp_restart.


src/mtcp/mtcp_restart.c
```diff
@@ -638,7 +638,7 @@ restorememoryareas(RestoreInfo *rinfo_ptr)
   if (rinfo_ptr->saved_brk != NULL) {
     // Now, we can do the pending mtcp_sys_brk(rinfo.saved_brk).
     // It's now safe to do this, even though it can munmap memory holding rinfo.
-    MTCP_ASSERT(mtcp_sys_brk(rinfo_ptr->saved_brk) == 0);
+    mtcp_sys_brk(rinfo_ptr->saved_brk);
```
karya0: Should we just check the return value anyways (to avoid compiler warning)?

PR #528, opened 2017-01-15
Better handling of vDSO

rohgarg:
We now warn the user if the vDSO layouts on the restart machine and the checkpoint machine are incompatible.

### DIFF
--- a/src/dmtcp_dlsym.cpp
+++ b/src/dmtcp_dlsym.cpp
@@ -599,3 +599,26 @@ dmtcp_dlsym_lib(const char *libname, const char *symbol)
                                                      &default_symbol_index);
   return result;
 }
+
+/*
+ * Returns the offset of the given function within the given shared library
+ * or -1 if the function does not exist in the library
+ */
+EXTERNC ptrdiff_t
+dmtcp_dlsym_lib_fnc_offset(const char *libname, const char *symbol)
+{
+  dt_tag tags;
+  ptrdiff_t ret = -1;
+  Elf32_Word default_symbol_index = 0;
+
+  // Determine where this function will return
+  void* return_address = __builtin_return_address(0);
+  void *result = dlsym_default_internal_flag_handler(NULL, libname, symbol,
+                                                     NULL,
+                                                     return_address, &tags,
+                                                     &default_symbol_index);
+  if (result) {
+    ret = (char*)result - tags.base_addr;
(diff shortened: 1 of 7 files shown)


src/dmtcp_dlsym.cpp
```diff
@@ -282,11 +282,23 @@ get_dt_tags(void *handle, dt_tag *tags)
...
   ElfW(Dyn) * cur_dyn;
 
+/*
+ * In most cases, the entries are already relocated, but for some --
+ * like vDSO -- we need to manually relocate at access time.
+ */
+
+#define ADJUST_DYN_INFO_RO(dst, map, dyn, type)    \
+  if (dyn->d_un.d_ptr > map->l_addr) {             \
+    dst = (type)dyn->d_un.d_ptr;                   \
```
rohgarg: Good point! I have updated the patch now.

src/dmtcp_dlsym.cpp
```diff
@@ -282,33 +280,59 @@ get_dt_tags(void *handle, dt_tag *tags)
...
+ * Here's an example call trace:
+ *   _start()
+ *     ...
+ *     dl_main()
+ *       ...
+ *       _dl_map_object()
+ *         _dl_map_object_from_fd()
+ *           _dl_map_segments()
+ *           elf_get_dynamic_info()
+ *
```
gc00: Let's add a small comment:  This code extends dmtcp_dlsym to work on VDSO, while in libc, they have a separate internal vdso_dlsym function for this purpose.

src/dmtcp_dlsym.cpp
```diff
@@ -575,3 +581,19 @@ dmtcp_dlvsym(void *handle, char *symbol, const char *version)
...
 }
+
+EXTERNC void *
+dmtcp_dlsym_lib(const char *libname, const char *symbol)
+{
+  dt_tag tags;
+  Elf32_Word default_symbol_index = 0;
+
+  // Determine where this function will return
+  void* return_address = __builtin_return_address(0);
```
gc00: @rohgarg observed that this is a stel comment.

gc00:
Very nice.  For the one commit, 'dmtcp_dlsym: Fix for resolving functions in vDSO', can we also cherry-pick that into the 2.4.8 release, since it's a bug fix?

PR #530, opened 2017-01-16
Fixed a bug in signal handler restoration during restart.

karya0:
The signal handlers must be restored before the restart hooks for
plugins are called because some plugins might perform tasks that could
potentially generate a signal. For example, the timer plugin may restore
a timer which will fire right away, and not having an appropriate signal
handler could kill the process.


karya0:
I'm not sure. Can you help figure that out?

rohgarg:
@karya0: It looks like the fix needs to go into the other two branches as well -- 2.5.1 and 2.4.8. I'm adding the two labels now.

PR #531, opened 2017-01-16
Support for mixed 32-/64-bit binaries

### LINKED ISSUE #524: Mixed 32-/64- bit processes on the same node fail on master
With the new plugin model, we are using `pthread_barrier` for node-local barriers. The node-local barriers are implemented by creating a `pthread_barrier_t` variable inside the shared-area that is shared by all processes belonging to a single node. Unfortunately, the underlying data structure - `pthread_barrier_t` - is different for 32- and 64- bit processes. This leads to failures when 32- and 64- bit processes try to access it.

One potential solution is to re-implement `pthread_barrier_wait` in shareddata in an arch-independent way.

karya0:
Fixes #524.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -123,7 +123,18 @@ contrib: dmtcp
 	cd contrib && $(MAKE)
 
 tests: build
+ifeq ($(M32),0)
+	if file ./test/dmtcp1 | grep -cq "ELF 32-bit"; then \
+	  cd test && $(MAKE) clean; \
+	fi
+endif
 	cd test && $(MAKE)
+
+tests-32: build
+	if file ./test/dmtcp1 | grep -cq "ELF 64-bit"; then \
+	  cd test && $(MAKE) clean; \
+	fi
+	cd test && $(MAKE) M32=1
 
 # Prevent mtcp_restart from flying out of control
 # (but Java/IcedTea6-1.9.x/RHEL-6.1 uses lots of memory,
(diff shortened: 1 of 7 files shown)


gc00:
I have concerns about the following two commits.  The remaining three are good to go.
> 	Added check-32, tests-32, etc., targets. 		
>	Reimplement local barriers using futexes. 

1. For 'Added check-32, tests-32, etc., targets', wouldn't it be better to create a separate test-32 directory and do the 32-bit tests there?  As you know, most 32-bit processes today appear as helper processes in a 64-bit computation.  In this mixed mode, it's easier if we can retain both the 32-bit and the 64-bit test directory for rapid testing.  This is analogous to our directories `lib/dmtcp` and 'lib/dmtcp/32` (I forget the exact name of the last directory).

2. For 'Reimplement local barriers using futexes', I'd like to see some kind of configure test or something that continues to make the POSIX version of DMTCP the default.  One of the strengths of DMTCP is that it follows a very stable POSIX API with good documentation so that others can quickly read and understand our code.  I don't mind adding things like 'futex' for special cases, but it should be on a special path, and not on the path that the more naive code readers are trying to understand.  Further, there are benefits to porting to other systems (e.g., BSD, Mac OS, Microsoft Azure, Windows Subsystem for Linux).  I don't trust to having a stable `futex` API on those four operating systems.

karya0:
I have split this PR into three different PRs:
* #546: 32-bit fixes
* #547: Local barriers
* #548: make check-32

PR #536, opened 2017-01-16
Renamed DEBUG to LOGGING in mtcp.

karya0:
This was leftover from cf53732.

### DIFF
--- a/src/mtcp/mtcp_util.h
+++ b/src/mtcp/mtcp_util.h
@@ -35,11 +35,11 @@
     mtcp_abort();                                       \
   }
 
-#ifdef DEBUG
+#ifdef LOGGING
 # define DPRINTF MTCP_PRINTF
-#else // ifdef DEBUG
+#else // ifdef LOGGING
 # define DPRINTF(args ...) // debug printing
-#endif // ifdef DEBUG
+#endif // ifdef LOGGING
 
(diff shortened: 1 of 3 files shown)


src/mtcp/mtcp_util.h
```diff
@@ -35,11 +35,11 @@
     mtcp_abort();                                       \
   }
 
-#ifdef DEBUG
+#ifdef LOGGING
```
rohgarg: Is the `LOGGING` macro available in mtcp? Do we need to change `Makefile.in` in the mtcp directory? Could you also see if we need to fix Makefile.debug in the mtcp directory?

karya0: Good catch! I'll take a look and update it.

karya0: It is available:

    gcc -I../../include -I. -DLOGGING -g -O2 -DLOGGING -c -o mtcp_restart.o\
      -DPIC -fPIC -fno-stack-protector -g -O0 mtcp_restart.c

I'll update Makefile.debug.

rohgarg: Great! Thank you!

karya0: Updated the PR. Should be good to go now.

Issue #537, opened 2017-01-18
dmtcp_launch hangs on RHEL5 and RHEL6


aparashk:
I have experienced problems getting a working build of DMTCP 2.5.0 final tag on Redhat/Centos Linux.

`dmtcp_launch` hangs when executed with a binary for checkpointing. This is easy to replicate on both 32-bit and 64-bit installs of RHEL6/Centos6 (Centos5 first requires a patch as DMTCP depends on a newer version of `binutils` than what Centos5 provides):

`./configure ; make ; make check`

hangs on both OSes. The same sequence works on Ubuntu 12.04 and 14.04 and all testcases pass.

rohgarg:
@aparashk: Thanks for reporting this. I'll look into this.

aparashk:
Thank you for the amazing support!

rohgarg:
Fixed via PR's #539 (for 2.4 and 2.5) and #540 (for 3.0).

Issue #476, opened 2016-12-24
Fixing the order of constructors/destructors


jiajuncao:
For global and static local objects of DMTCP, we would like our constructors to be called before any user static objects are constructed, and our destructors to be called after every user static objects are destructed. This is motivated by PR #470, where user code tries to access DMTCP static objects when their static objects are being destructed.

PR #539, opened 2017-01-18
Eliminate static C++ objects from jassert, jfilesystem, and dmtcpworker

rohgarg:
This fixes issue #537. Apparently, on older libc/libstdc++, the library
acquires a low-level lock in its constructor. This can lead to a deadlock
when the library, in its constructor, ends up calling a DMTCP wrapper,
which then calls other helper functions that have their own static
local objects that need to acquire the same lock to be initialized.
Here's a sample stacktrace:

    __lll_lock_wait_private () from /lib64/libc.so.6
    _L_lock_16 () from /lib64/libc.so.6
    __new_exitfn () from /lib64/libc.so.6
    __cxa_atexit_internal () from /lib64/libc.so.6
    theLogFilePath () at ../jalib/jassert.cpp:171
    jassert_internal::set_log_file ()
    dmtcp::Util::initializeLogFile ()
    prepareLogAndProcessdDataFromSerialFile () at dmtcpworker.cpp:231
    dmtcp_initialize () at dmtcpworker.cpp:266
    calloc (nmemb=1, size=1040) at alloc/mallocwrappers.cpp:32
    __new_exitfn ()
    __cxa_atexit_internal ()
    __static_initialization_and_destruction_0
    __do_global_ctors_aux ()
    _init ()
    _dl_init_internal ()
    _dl_start_user ()

### DIFF
--- a/jalib/jassert.cpp
+++ b/jalib/jassert.cpp
@@ -63,10 +63,28 @@ static int errConsoleFd = -1;
 // (as a lvalue) to modifying the underlying variable.
 
 static jalib::string&
-tmpDir() { static jalib::string s; return s; }
+tmpDir()
+{
+  static jalib::string *s = NULL;
+  if (s == NULL) {
+    // Technically, this is a memory leak, but s is static and so it happens
+    // only once.
+    s = new jalib::string();
+  }
+  return *s;
+}
 
 static jalib::string&
-uniquePidStr() { static jalib::string s; return s; }
+uniquePidStr()
+{
+  static jalib::string *s = NULL;
+  if (s == NULL) {
+    // Technically, this is a memory leak, but s is static and so it happens
+    // only once.
+    s = new jalib::string();
+  }
+  return *s;
+}
 
 static int
 jwrite(int fd, const char *str)
@@ -160,7 +178,15 @@ _open_log_safe(const jalib::string &s, int protectedFd)
 }
 
 static jalib::string&
-theLogFilePath() { static jalib::string s; return s; }
+theLogFilePath() {
+  static jalib::string *s = NULL;
+  if (s == NULL) {
+    // Technically, this is a memory leak, but s is static and so it happens
+    // only once.
+    s = new jalib::string();
+  }
+  return *s;
+}
 
 void
 jassert_internal::jassert_init()
--- a/jalib/jfilesystem.cpp
+++ b/jalib/jfilesystem.cpp
@@ -176,9 +176,13 @@ jalib::Filesystem::mkdir_r(const jalib::string &dir, mode_t mode)
 jalib::string
 jalib::Filesystem::GetProgramDir()
 {
-  static jalib::string value = DirName(GetProgramPath());
-
-  return value;
+  static jalib::string *value = NULL;
+  if (value == NULL) {
+    // Technically, this is a memory leak, but value is static and so it happens
+    // only once.
+    value = new jalib::string(DirName(GetProgramPath()));
+  }
+  return *value;
 }
 
 jalib::string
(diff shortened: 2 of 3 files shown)


gc00:
In addition, this PR fixes many of the cases noted in Issue #476.  It may even fix all of those cases, but we would have to go back and verify that.

jalib/jassert.cpp
```diff
@@ -63,10 +63,24 @@ static int errConsoleFd = -1;
 // (as a lvalue) to modifying the underlying variable.
 
 static jalib::string&
-tmpDir() { static jalib::string s; return s; }
+tmpDir()
+{
+  static jalib::string *s = NULL;
+  if (s == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but `s` is static and so it happens only once.

jalib/jassert.cpp
```diff
@@ -63,10 +63,24 @@ static int errConsoleFd = -1;
...
+  }
+  return *s;
+}
 
 static jalib::string&
-uniquePidStr() { static jalib::string s; return s; }
+uniquePidStr()
+{
+  static jalib::string *s = NULL;
+  if (s == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but `s` is static and so it happens only once.

jalib/jassert.cpp
```diff
@@ -160,7 +174,13 @@ _open_log_safe(const jalib::string &s, int protectedFd)
 }
 
 static jalib::string&
-theLogFilePath() { static jalib::string s; return s; }
+theLogFilePath() {
+  static jalib::string *s = NULL;
+  if (s == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but `s` is static and so it happens only once.

jalib/jfilesystem.cpp
```diff
@@ -176,9 +176,11 @@ jalib::Filesystem::mkdir_r(const jalib::string &dir, mode_t mode)
 jalib::string
 jalib::Filesystem::GetProgramDir()
 {
-  static jalib::string value = DirName(GetProgramPath());
-
-  return value;
+  static jalib::string *value = NULL;
+  if (value == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but `value` is static and so it happens only once.

jalib/jfilesystem.cpp
```diff
@@ -206,9 +208,11 @@ jalib::Filesystem::GetProgramName()
 jalib::string
 jalib::Filesystem::GetProgramPath()
 {
-  static jalib::string value = _GetProgramExe();
-
-  return value;
+  static jalib::string *value = NULL;
+  if (value == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but `s` is static and so it happens only once.

jalib/jfilesystem.cpp
```diff
@@ -259,9 +263,12 @@ jalib::Filesystem::FileExists(const jalib::string &str)
 jalib::StringVector
 jalib::Filesystem::GetProgramArgs()
 {
-  static StringVector rv;
+  static StringVector *rv = NULL;
+  if (rv == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but `rv` is static and so it happens only once.

src/dmtcpworker.cpp
```diff
@@ -276,6 +276,7 @@ dmtcp_initialize()
   static bool initialized = false;
 
   if (initialized) {
+    buf = new jalib::JBuffer(0); // To force linkage of jbuffer.cpp
```
gc00: The `buf =` should be part of teh `else` condition here.

Also, I suggest adding comment:
// Technically, this is a memory leak, but `rv` is static and so it happens only once.

PR #540, opened 2017-01-18
Eliminate static C++ objects from jassert, jfilesystem, and dmtcpworker

rohgarg:
This fixes issue #537. Apparently, on older libc/libstdc++, the library
acquires a low-level lock in its constructor. This can lead to a deadlock
when the library, in its constructor, ends up calling a DMTCP wrapper,
which then calls other helper functions that have their own static
local objects that need to acquire the same lock to be initialized.
Here's a sample stacktrace:

    __lll_lock_wait_private () from /lib64/libc.so.6
    _L_lock_16 () from /lib64/libc.so.6
    __new_exitfn () from /lib64/libc.so.6
    __cxa_atexit_internal () from /lib64/libc.so.6
    theLogFilePath () at ../jalib/jassert.cpp:171
    jassert_internal::set_log_file ()
    dmtcp::Util::initializeLogFile ()
    prepareLogAndProcessdDataFromSerialFile () at dmtcpworker.cpp:231
    dmtcp_initialize () at dmtcpworker.cpp:266
    calloc (nmemb=1, size=1040) at alloc/mallocwrappers.cpp:32
    __new_exitfn ()
    __cxa_atexit_internal ()
    __static_initialization_and_destruction_0
    __do_global_ctors_aux ()
    _init ()
    _dl_init_internal ()
    _dl_start_user ()

### DIFF
--- a/jalib/jfilesystem.cpp
+++ b/jalib/jfilesystem.cpp
@@ -158,8 +158,13 @@ int jalib::Filesystem::mkdir_r( const jalib::string& dir, mode_t mode)
 
 jalib::string jalib::Filesystem::GetProgramDir()
 {
-  static jalib::string value = DirName ( GetProgramPath() );
-  return value;
+  static jalib::string *value = NULL;
+  if (value == NULL) {
+    // Technically, this is a memory leak, but value is static and so it happens
+    // only once.
+    value = new jalib::string(DirName ( GetProgramPath() ));
+  }
+  return *value;
 }
 
 jalib::string jalib::Filesystem::GetProgramName()
@@ -183,8 +188,13 @@ jalib::string jalib::Filesystem::GetProgramName()
 
 jalib::string jalib::Filesystem::GetProgramPath()
 {
-  static jalib::string value = _GetProgramExe();
-  return value;
+  static jalib::string *value = NULL;
+  if (value == NULL) {
+    // Technically, this is a memory leak, but value is static and so it happens
+    // only once.
+     value = new jalib::string(_GetProgramExe());
+  }
+  return *value;
 }
 
 // NOTE: ResolveSymlink returns a string, buf, allocated on the stack.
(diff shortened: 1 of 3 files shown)


rohgarg:
This is a duplicate of PR #539 for the 2.5 and 2.4 branches.

gc00:
In addition, this PR fixes many of the cases noted in Issue #476. It may even fix all of those cases, but we would have to go back and verify that.

jalib/jfilesystem.cpp
```diff
@@ -183,8 +186,11 @@ jalib::string jalib::Filesystem::GetProgramName()
 
 jalib::string jalib::Filesystem::GetProgramPath()
 {
-  static jalib::string value = _GetProgramExe();
-  return value;
+  static jalib::string *value = NULL;
+  if (value == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but value is tatic and so it happens only once.

jalib/jfilesystem.cpp
```diff
@@ -229,9 +235,12 @@ bool jalib::Filesystem::FileExists ( const jalib::string& str )
 
 jalib::StringVector jalib::Filesystem::GetProgramArgs()
 {
-  static StringVector rv;
+  static StringVector *rv = NULL;
+  if (rv == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but `rv` is static and so it happens only once.

src/dmtcpworker.cpp
```diff
@@ -566,7 +566,10 @@ void dmtcp_Alarm_EventHook(DmtcpEvent_t event, DmtcpEventData_t *data);
 
 void DmtcpWorker::eventHook(DmtcpEvent_t event, DmtcpEventData_t *data)
 {
-  static jalib::JBuffer buf(0); // To force linkage of jbuffer.cpp
+  static jalib::JBuffer *buf = NULL;
+  if (buf == NULL) {
```
gc00: Suggest adding comment:
// Technically, this is a memory leak, but `buf` is static and so it happens only once.

Issue #438, opened 2016-09-29
dmtcp_get_coord_ckpt_dir() is not working.


jiajuncao:
When being called from the user thread, dmtcp_get_coord_ckpt_dir() sends a message to the coordinator. And the coordinator will send back the reply with the result. However, at the same time, the checkpoint thread is expecting the suspend message from the coordinator. It therefore will fail when it receives the message it cannot identify.

gc00:
@jiajuncao, do the commits above now allow us to close this issue?  If so, could you do that?  Thanks.

PR #544, opened 2017-01-23
Use an extra socket in dmtcp_get_coord_ckpt_dir().

jiajuncao:
Fix issue #438.


src/coordinatorapi.cpp
```diff
@@ -386,16 +386,26 @@ string CoordinatorAPI::getCoordCkptDir(void)
...
+  sock >> msg;
   msg.assertValid();
-  JASSERT(msg.type == DMT_GET_CKPT_DIR_RESULT) (msg.type);
+  JASSERT(msg.type == DMT_GET_COORD_CKPT_DIR_RESULT) (msg.type);
+
+  if (msg.extraBytes > 0) {
+    sock.readAll(buf, msg.extraBytes);
+  }
+  else {
+    JWARNING("Failed to get the coordinator checkpoint dir. "
```
rohgarg: `JWARNING()` has the same syntax as `JASSERT()`. Please change this to:
`JWARNING(false).Text("...")`

src/dmtcpplugin.cpp
```diff
@@ -199,7 +199,9 @@ EXTERNC int dmtcp_set_ckpt_dir(const char* dir)
 EXTERNC const char* dmtcp_get_coord_ckpt_dir(void)
 {
   static string dir;
+  dmtcp_disable_ckpt();
```
rohgarg: I believe we should be using the helper macros instead of calling the functions directly.

jiajuncao: You mean DMTCP_PLUGIN_DISABLE_CKPT()? I assume this is primarily used in plugins.

src/dmtcpplugin.cpp
```diff
@@ -199,7 +199,10 @@ EXTERNC int dmtcp_set_ckpt_dir(const char* dir)
 EXTERNC const char* dmtcp_get_coord_ckpt_dir(void)
```
rohgarg: Could you please document that the API returns a zero-length string, "", when a checkpoint is in progress?

src/dmtcpplugin.cpp
```diff
@@ -199,7 +199,10 @@ EXTERNC int dmtcp_set_ckpt_dir(const char* dir)
...
 {
   static string dir;
```
rohgarg: Is there any reason we want a static string object here? If not, can we get rid of this static string object?

PR #545, opened 2017-01-23
Added functionality of setting global ckpt dir.

jiajuncao:


src/coordinatorapi.cpp
```diff
@@ -409,6 +409,31 @@ void CoordinatorAPI::updateCoordCkptDir(const char *dir)
   _coordinatorSocket.writeAll(dir, strlen(dir) + 1);
 }
```
gc00: Suggest adding a comment to introduce `updateGlobalCkptDir()`.  Something like:
> //  Invoked by a plugin via `dmtcp.h:dmtcp_set_global_ckpt_dir()` and
// via `dmtcpplugin.cpp:dmtcp_set_global_ckpt_dir()`

gc00:
I suggest adding a comment, as described above to introduce `updateGlobalCkptDir()`.

Separately, we had agreed that the documentation of the behavior was complicated enough that we would add an extended comment.  It could be added in front of `dmtcp.h:dmtcp_set_global_ckpt_dir()`, or else in the doc directory.  I sort of like it in fhe `dmtcp.h` file myself.  It could be something like:

> One peer can request a change to a global ckpt dir while another requests a checkpoint.  If a peer can send a global ckpt dir request before being suspended, then the coordinator will recognize this request, and pass on to all peers the new global ckpt dir.  Upon restart, all peers will remember the current ckpt dir (whether the original local one or a change to a global one).  The ckpt dir will be sticky across fork, but _not_ across ssh.  After ssh, the new peer initializes itself to a local ckpt dir, since there is no guarantee of the global ckpt directory existing on the new host.  It is the responsibility of the user to ensure that the global ckpt directory remembered by any peer still exists at restart time, or else to again change the global ckpt directory at restart time.

gc00:
Right now, the labels for the PR say that this also goes into 3.0.  But in discussions, I think @rohgarg said that 3.0 will have a different mechanism for which this is incompatible.

@karya0 and @rohgarg, do we want a similar PR like this one to be added to the 3.0/master branch?

src/dmtcp_coordinator.cpp
```diff
@@ -755,6 +756,33 @@ void DmtcpCoordinator::onConnect()
...
+
+    if (s.minimumState == WorkerState::RUNNING && s.minimumStateUnanimous
+        && !workersRunningAndSuspendMsgSent) {
+      reply.type = DMT_UPDATE_GLOBAL_CKPT_DIR_SUCCEED;
+      globalCkptDir = extraData;
+      JTRACE("Received DMT_UPDATE_GLOBAL_CKPT_DIR msg, "
+             "updating the global checkpoint dir")
+        (hello_remote.from) (globalCkptDir);
+    }
+    else {
```
rohgarg: Change this to: `} else {`

PR #549, opened 2017-01-26
pathvirt: resolve the real path if it's a symlink.

jiajuncao:


gc00:
Let's apply this to both 2.5.1 and 3.0/master.

jiajuncao:
@gc00 path virtualization hasn't been pushed in to 3.0. We need to make some additional changes (due to the new plugin API) to make that work. Let's keep it in 2.5 for now.

PR #550, opened 2017-01-27
pathvirt: patch new prefix env var before execve().

jiajuncao:
This should be a temp fix. In the long term, we need to think about
how to inject new env vars for applications that have private copy
of env vars.


gc00:
Can we also add a new issue that refers to this PR #550, and the issue described in its first comment?  Thanks.

PR #551, opened 2017-01-31
VDSO fix for i386 (for 2.4 branch)

gc00:
__This PR (same as PR  #480) was supposed to have gone into 2.4.7.  The NEWS file said that it did.  But it didn't make it.  Let's add it now.__

Apparently, the second vvar page is not readable and so a memcpy()
results in a BUS error. This commit tries to hack through it by ignoring
the second vvar page.

This is a port of PR #469 from master to the 2.5 branch.


rohgarg:
@gc00: This change has already been pushed in. See 60bbac8a7753a665cd85e9b7ec5ec68dc2eac5c5

Issue #538, opened 2017-01-18
Document implicit dependence on binutils version


aparashk:
I have found that, somewhere between DMTCP 2.4.5 and DMTCP 2.5.0, or more specifically in [this commit](https://github.com/dmtcp/dmtcp/commit/72be59c8658d4e1fe57899e7ed42b3ed8993cecf), DMTCP started to require a newer version of binutils than some legacy Linux distributions, for example Redhat 5, come with. More specifically, this is due to the use of the STT_GNU_IFUNC macro.

It would be good for the binutils version dependence to be documented somewhere.

rohgarg:
@aparashk: Good catch! We should definitely fix this.

rohgarg:
@aparashk: I have a preliminary patch that seems to work. It's still a work-in-progress though. Could you please check if this is sort-of-correct? (This is on top of the 2.5 branch.)

```patch
diff --git a/src/dmtcp_dlsym.cpp b/src/dmtcp_dlsym.cpp
index 1485c6da..804b24b1 100644
--- a/src/dmtcp_dlsym.cpp
+++ b/src/dmtcp_dlsym.cpp
@@ -356,6 +356,8 @@ void *dlsym_default_internal_library_handler(void *handle, const char*symbol,
   *default_symbol_index_p = default_symbol_index;
 
   if (default_symbol_index) {
+#if __GLIBC_PREREQ(2, 11)
+    // See https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Function-Attributes.html
     if (ELF64_ST_TYPE(tags.symtab[default_symbol_index].st_info) ==
         STT_GNU_IFUNC) {
       typedef void* (*fnc)();
@@ -363,8 +365,8 @@ void *dlsym_default_internal_library_handler(void *handle, const char*symbol,
                      tags.symtab[default_symbol_index].st_value);
       return f();
     } else {
-      return tags.base_addr + tags.symtab[default_symbol_index].st_value;
-    }
+#endif
+    return tags.base_addr + tags.symtab[default_symbol_index].st_value;
   } else {
     return NULL;
   }
diff --git a/src/plugin/ipc/event/eventconnection.cpp b/src/plugin/ipc/event/eventconnection.cpp
index 0dec27cc..6e4e1b5c 100644
--- a/src/plugin/ipc/event/eventconnection.cpp
+++ b/src/plugin/ipc/event/eventconnection.cpp
[...]

PR #552, opened 2017-01-31
Updated NEWS and bumped revision to 2.4.8

gc00:
The title speaks for itself.  We should push this in after any pending PR's for 2.4.8.


rohgarg:
I think we'll have one more set of fixes and so the NEWS file will need to be updated. In particular, I'm working on porting PR #557 (fixes for epoll) to the 2.4 branch.

rohgarg:
I have now pushed in the changes for PR #557 to the 2.5 and 2.4 branches. I think the only thing we are missing now for the 2.4.8 release is a fix for issue #538.

gc00:
@rohgarg, Please seem my comment at issue #538 here:  https://github.com/dmtcp/dmtcp/issues/538#issuecomment-279260227
Let me know what you think of this plan.

gc00:
@rohgarg, Since you've now added the missing fix, is this ready to be pushed in?

rohgarg:
Sure, but do you also need to run `autogen.sh` to generate the updated configure script (with the new version information)?

gc00:
Good point.  I've added a commit to run autogen.sh with the new VERSION.  Next, I'll push it in.

PR #564, opened 2017-02-13
mtcp_restart: Fix a bug introduced in 71e1f6c9 for Intel compiler

rohgarg:


gc00:
The bug was in commit 71e1f6c9.  (Adding this comment to make the link clickable. :-) )

src/mtcp/mtcp_restart.c
```diff
@@ -417,7 +417,7 @@ static void restart_fast_path()
   asm volatile(
    // 96 = offsetof(RestoreInfo, rinfo.restorememoryareas_fptr)
```
gc00: Please also update the comment:  `// 104 - offsetof(...)`.
Thanks for catching this bug fix of mine.  But I thought I had tested my code and it worked.  Does the restart really fail without this?  If so, thanks again.

rohgarg: The bug would only appear when compiled with `icc` (Intel compiler).

gc00:
Aha!  It's about icc.  Thanks.  After the minor change in the comment is added, feel free to push this in.  LGTM.

PR #565, opened 2017-02-16
Fix obvious error

planetA:
Seems that merge conflict was not solved properly. Compilation fails at this step.


gc00:
Thanks, @planetA .  For others, this bad commit happened at PR #523 in commit 50ec4eb9596.

Issue #566, opened 2017-02-17
Segfault in dmtcp_dlsym_lib_fnc_offset


planetA:
When a checkpoint is about to be created, the workers crache with a segfault. Behavior is deterministic and the same for all the workers in the application.

The problem appeared for me with the commit 3dff1b3, but probably error was introduced a bit earlier, because this commit only invokes some functions.

Here is a backtrace:

    Thread 3 (Thread 0x2ad3f708c700 (LWP 6515)):
    #0  hash_first (use_gnu_hash=0, hash_table=0xffffffffff700124, name=0x2ad3f3c72c11 <dmtcp::Util::writeCoordPortToFile(int, char const*)::__FUNCTION__+353> "__vdso_clock_gettime") at dmtcp_dlsym.cpp:166
    #1  dlsym_default_internal_library_handler (handle=handle@entry=0x33dfe22718, symbol=symbol@entry=0x2ad3f3c72c11 <dmtcp::Util::writeCoordPortToFile(int, char const*)::__FUNCTION__+353> "__vdso_clock_gettime", version=version@entry=0x0, tags_p=tags_p@entry=0x2ad3f7086430, 
        default_symbol_index_p=default_symbol_index_p@entry=0x2ad3f708642c) at dmtcp_dlsym.cpp:368
    #2  0x00002ad3f3c3aeb4 in dlsym_default_internal_flag_handler (handle=handle@entry=0x0, libname=libname@entry=0x2ad3f3c72c26 <dmtcp::Util::writeCoordPortToFile(int, char const*)::__FUNCTION__+374> "linux-vdso", 
        symbol=symbol@entry=0x2ad3f3c72c11 <dmtcp::Util::writeCoordPortToFile(int, char const*)::__FUNCTION__+353> "__vdso_clock_gettime", version=version@entry=0x0, addr=<optimized out>, tags_p=tags_p@entry=0x2ad3f7086430, default_symbol_index_p=0x2ad3f708642c)
        at dmtcp_dlsym.cpp:480
    #3  0x00002ad3f3c3b07b in dmtcp_dlsym_lib_fnc_offset (libname=libname@entry=0x2ad3f3c72c26 <dmtcp::Util::writeCoordPortToFile(int, char const*)::__FUNCTION__+374> "linux-vdso", 
[...]

PR #567, opened 2017-02-17
dmtcp_dlsym: Fix a crash when link_map doesn't contain vDSO entry

rohgarg:
The commit: 3dff1b3 added a warning for the user that would be triggered
when the checkpoint and the restart machines had incompatible vDSO
layouts. The warning was added as a nice-to-have feature where earlier the
restart would have just crashed without a clue. The code would determine
the layout using a minor extension to the `dmtcp_dlsym` function.

However, the current implementation would lead to a crash if the vDSO
section was not listed as part of the link_map. We have now observed this
issue on two separate systems -- a user reported this in issue #566 and
@jiajuncao separately observed this on a CentOS 7.2 system. This patch
fixes this problem by ensuring that `dmtcp_dlsym` doesn't crash in the
corner cases.

### DIFF
--- a/src/dmtcp_dlsym.cpp
+++ b/src/dmtcp_dlsym.cpp
@@ -472,17 +472,15 @@ dlsym_default_internal_flag_handler(void *handle,
     /* If the caller specified a specific library name, only search through
      * that.
      */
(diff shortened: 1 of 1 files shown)


rohgarg:
This should help fix issue #566.

PR #571, opened 2017-02-22
socket-plugin: Add support for robust detection of external sockets

rohgarg:


src/plugin/ipc/ipc.cpp
```diff
@@ -93,6 +93,10 @@ static DmtcpBarrier socketBarriers[] = {
   { DMTCP_PRIVATE_BARRIER_PRE_CKPT, SocketConnList::saveOptions, "PRE_CKPT" },
   { DMTCP_LOCAL_BARRIER_PRE_CKPT, SocketConnList::leaderElection,
     "LEADER_ELECTION" },
+  { DMTCP_GLOBAL_BARRIER_PRE_CKPT, SocketConnList::ckptRegisterNSData,
+    "CKPT_NS_REGISTER_DATA" },
```
karya0: Can we rename this to `CKPT_REGISTER_PEER_INFO` or something like that so it's easy to decipher what it's doing?

src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -418,6 +418,57 @@ TcpConnection::TcpConnection(const TcpConnection &parent,
 }
 
 void
+TcpConnection::sendPeerInformation()
+{
+  struct sockaddr local, remote;
```
karya0: s/sockaddr/sockaddr_storage ?

src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -418,6 +418,57 @@ TcpConnection::TcpConnection(const TcpConnection &parent,
...
+  socklen_t localsz, remotesz;
+  switch (_type) {
+  case TCP_CONNECT:
+  {
+    localsz = sizeof(local);
+    JASSERT(getsockname(_fds[0],
```
karya0: Is value returned by `getsockname` name guaranteed to match with the value returned by `getpeername`?

src/plugin/ipc/socket/socketconnlist.h
```diff
@@ -51,6 +55,9 @@ class SocketConnList : public ConnectionList
     virtual void sendQueries();
     virtual void refill(bool isRestart);
 
+    virtual void preCkptRegisterNSData();
```
karya0: They don't need to be virtual, right?

karya0:
LGTM modulo some minor comments!

rohgarg:
Thanks for the review, @karya0. I have addressed most of your comments. I'm not a 100% sure about whether the result of `getpeername()` will match the result of `getsockname()` in all cases. Is there a better alternative?

gc00:
@rohgarg, I really like the goal of this code!  From our discussions, it sounds like this code will convert an external socket at checkpoint time into a dead socket at restart time, which DMTCP has been needing for a while.  I have two smaller comments now, and I'll look deeper later.

1. Can we expand the name `SocketConnList::preCkptRegisterNSData()`?  Does it mean, for example, `SocketConnList::preCkptRegisterNewSocketData()`?  If so, let's use the full name.
2. Does this change the standard treatment by DMTCP of sockets, or is this an option?  I'm guessing that it changes the standard behavior.  In either case, how do we communicate this information to the DMTCP end user?  For example, should we issue a warning about an external socket that will be restored as a dead socket on restart?  As a first step, let's at least place a comment in this code saying how the end-user behavior is affected, and maybe issue a `JWARNING`.  (User's can use `--quiet`` to turn off warnings if they don't like them.)

src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -418,6 +418,57 @@ TcpConnection::TcpConnection(const TcpConnection &parent,
...
+    int ret = dmtcp_send_query_to_coordinator("RemoteConnections",
+                                              &local, localsz,
+                                              &remote, &remotesz);
+    if (ret != 0) {
+      JASSERT(remotesz == sizeof(remote));
+    } else {
+      markExternalConnect();
+    }
+  }
+}
```
jiajuncao: What's the key used here? it seems `local` is not initialized. Also, should we do the same for accepted sockets?

rohgarg: Good catch! I have now updated the PR.

gc00:
@rohgarg,  Could you answer the two questions that I have above?  Everything else in the PR looks good.  But I'd like (a) the trivial symbol name change; and (b) to understand better if the default behavior is changed.

What I understand from discussions with you outside of this is that the default behavior is changed to detect external sockets.  This is at the cost of an extra barrier.  Could the extra barrier cost a lot when scaling to, for example, an MPI computation over 10,000 processes?  If the cost is high in that scenario, then we should add a runtime flag to disable this check.  If the cost is not high, then the PR is good as is (subject to the name change for `SocketConnList::preCkptRegisterNSData()`.  Thanks.

PR #573, opened 2017-02-25
Misc. fixes for checkpoint-restart of SysV SHM objects

rohgarg:


src/plugin/svipc/sysvipcwrappers.cpp
```diff
@@ -68,9 +68,10 @@ shmget(key_t key, size_t size, int shmflg)
   int virtId = -1;
 
   DMTCP_PLUGIN_DISABLE_CKPT();
-  realId = _real_shmget(key, size, shmflg);
+  key_t realKey = dmtcp_virtual_to_real_pid(getpid());
```
rohgarg: Thanks for catching this, @balajir1990. I'll fix this in the next iteration.

src/plugin/svipc/sysvipc.cpp
```diff
@@ -430,6 +430,37 @@ SysVIPC::serialize(jalib::JBinarySerializer &o)
...
 }
 
+int
+SysVShm::virtualToRealKey(key_t k)
+{
+  if (_keyMap.find(k) != _keyMap.end()) {
+    return _keyMap[k];
+  } else {
+    int realId = SharedData::getRealIPCId(SYSV_SHM_KEY, k);
+    updateKeyMapping(k, realId);
```
jiajuncao: Shall we update the mapping only if `realID != -1`? Just to avoid inserting a useless entry.

rohgarg:
Thanks for the feedback, @jiajuncao. I have now updated the patch to address your comments.

gc00:
Hi @rohgarg, This PR has been approved for a while now.  Could you rebase and push it in.  :-)  Thanks.

PR #574, opened 2017-02-28
Remove unused file

planetA:
This file isn't used anywhere, it just sits in the source code tree
and confuses people by appearing to represent old barrier interface.


planetA:
I added additional commit which is reasonably related, because it also purges traces of old barrier interface.

src/dmtcp_coordinator.cpp
```diff
@@ -44,16 +44,15 @@
...
- *                -> CHECKPOINTED -> NAME_SERVICE_DATA_REGISTERED           *
- *                -> DONE_QUERYING -> REFILLED -> RUNNING		    *
- * Restart:    RESTARTING -> NAME_SERVICE_DATA_REGISTERED                   *
- *                -> DONE_QUERYING -> REFILLED -> RUNNING	            *
- * If debugging, set gdb breakpoint on:					    *
- *   DmtcpCoordinator::onConnect					    *
- *   DmtcpCoordinator::onData						    *
- *   DmtcpCoordinator::handleUserCommand				    *
- *   DmtcpCoordinator::broadcastMessage					    *
+ * Checkpoint: RUNNING -> SUSPENDED -> CHECKPOINTING -> CHECKPOINTED        *
```
rohgarg: `CHECKPOINTING` --> `(Checkpoint barriers)`

planetA: Then I would change it to CHECKPOINTING -> (Checkpoint barriers) -> CHECKPOINTED, because CHECKPOINTING is a designated state.

rohgarg: Okay.

planetA: Done

PR #576, opened 2017-03-01
Fixed jalib's GetProgramArgs() to stop using JALLOC_HELPER_MALLOC

gc00:
**THIS IS READY FOR REVIEW NOW.**

Any command line longer than 4095 characters will fail in DMTCP right now.  I have a fix that works, but it's not the cleanest fix.

This PR needs to be modified before we push it in.  A bug was introduced in commit dddacca52 (March, 2014) and papered over in commit 0b1a6dd34 (May, 2014). 

Out of my own curiosity, I searched back for the origin of the combination of malloc and getdelim (which implicitly re-uses realloc).  It seems to have been here from the very earliest days:  f04e29e.  

The commit here removes the use of `JALLOC_HELPER_MALLOC`, and uses `malloc`.  Another bad possibility is to `JASSERT` and tell the user to modify the constant in the code.
<hr>

I will fix this later, to instead do the obvious C thing.  That should be compatible with C++ here.  See my next comment, below, for the origin of this decision.

### DIFF
--- a/jalib/jfilesystem.cpp
+++ b/jalib/jfilesystem.cpp
@@ -265,6 +265,88 @@ jalib::Filesystem::FileExists(const dmtcp::string &str)
   }
 }
 
+jalib::StringVector
+jalib::Filesystem::GetProgramArgs()
+{
+  // FIXME:  This static StringVector can be de-allocated during cleanup,
+  //     If the base executable or libc called a DMTCP wrapper as a utility,
+  //     it would be bad.
+  static StringVector *rv = NULL;
+  if (rv == NULL) {
+    // Technically, this is a memory leak, but rv is static and so it happens
+    // only once.
+    rv = new StringVector();
+  }
+
+  if (rv->empty()) {
+    int fd = open("/proc/self/cmdline", O_RDONLY);
+    JASSERT(fd != -1) (JASSERT_ERRNO).Text("failed to open command line");
+
+    jalib::string longword;
+    bool is_longword = false;
+    bool is_end_of_file = false;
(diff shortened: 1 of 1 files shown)


gc00:
I talked with @karya0.  I will remove`getdelim` and `JALLOC_HELPER_MALLOC`.  It suffices to call `push_back` on a word as a C string, and then advance `lineptr` by `strlen(lineptr)+1` and then conitue in  the loop until there are no more argument words to process.  I'll implement the next version.

gc00:
I now have a new implementation that is now ready for review.

I have modified the single function `jalib/jfilesystem.cpp`.  I have tested by setting a buf of only 4 bytes, and then calling `dmtcp_launch` on `echo` with arguments larger than 4 bytes.  This code should be using our own allocator (e.g., for C++ string concatenation), and only in rare cases when there is a word of length larger than 4095, for the function `GetProgramArgs`.

wpoely86:
I've hit this bug today and discovered this bug report after some searching. It would be nice to get this merged ASAP because hitting the 4096 limit is not uncommon if you use a module tool (like Lmod).

@gc00 your patch seems very complex? Can this not be done with some simpler C++ code (the `substr` method of the class `std::string`)?

karya0:
@wpoely86: I have a newer fix that simplifies the logic. I'll try to cleanup the code and generate a new PR sometime this week.

karya0:
This PR is irrelevant now because of https://github.com/dmtcp/dmtcp/pull/853. The GetProgramArgs functions is gone now.

PR #577, opened 2017-03-03
Remove meaningless barrier

planetA:
Resume refill barrier for sysvipc plugin never called, because refill
function always immediately returns, when isRestart is not true.

### DIFF
--- a/src/plugin/svipc/sysvipc.cpp
+++ b/src/plugin/svipc/sysvipc.cpp
@@ -58,12 +58,10 @@ using namespace dmtcp;
  *  8. Non ckptLeader processes unmap all shmat() addresses corresponding to
  *     the shm-object.
  *  9. BARRIER -- CHECKPOINTED
- * 10. At this point, the contents of the memory-segment have been saved.
- * 11. BARRIER -- REFILLED
- * 12. Re-map the memory-segment into each process's memory as it existed prior
+ * 10. Re-map the memory-segment into each process's memory as it existed prior
  *     to checkpoint.
- * 13. TODO: Unmap the memory that was mapped in step 4.
- * 14. BARRIER -- RESUME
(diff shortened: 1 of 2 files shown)


planetA:
If the change is OK for you I will try to update the comment about plugin logic in the top of the sysvipc.cpp.

planetA:
Actually the same applies to EventConnList. RegisterNSData and sendQueries barriers just call empty function in the end. This Conn list manages file descriptor, so it is not clear for me what is the role of NS barriers there. Shouldn't recovery of file descriptors involve no name resolution over coordinator?

rohgarg:
@planetA: On the face of it, the patch looks reasonable. It could very well be some legacy code from the 1.x days, but I'd have to look more closely before I can approve this. @karya0 do you have any suggestions/comments?

rohgarg:
@planetA: Could you please update your branch? And also please update the comment at the top of the file. I'll merge this into 3.0 once you do that.

planetA:
Hi. I'm not sure if you're getting notifications when I update the branch, so I leave a comment here.

src/plugin/svipc/sysvipc.cpp
```diff
@@ -58,12 +58,10 @@ using namespace dmtcp;
  *  8. Non ckptLeader processes unmap all shmat() addresses corresponding to
  *     the shm-object.
  *  9. BARRIER -- CHECKPOINTED
- * 10. At this point, the contents of the memory-segment have been saved.
```
rohgarg: Is there a specific reason you deleted this line? (Sorry, I'm not trying to nitpick here; it's just a genuine question.)

planetA: I removed this line because it was related to the barrier, which was not activated. You think I should merge this line with the next one?

PR #580, opened 2017-03-11
Disable mutex wrappers by default on 3.0

rohgarg:
This is a port of some patches from 2.5 for the master branch.


rohgarg:
Ping.

PR #583, opened 2017-03-26
Fixed a corner case where path and cwd are equal.

jiajuncao:


rohgarg:
@jiajuncao: Should this also go into 3.0?

PR #590, opened 2017-04-07
Fix PMI issues

jiajuncao:
1. Redirect `PMI_Init()` to our wrapper, if the pmi library is opened by `dlopen()`.
2. Remove the pmi socket as a stale connection.

### DIFF
--- a/plugin/batch-queue/rm_pmi.cpp
+++ b/plugin/batch-queue/rm_pmi.cpp
@@ -143,7 +143,7 @@ extern "C" int PMI_Init( int *spawned )
 // pmi is correctly initialized.
 extern "C" void *dlsym(void *handle, const char *symbol)
 {
-  if (strcmp(symbol, "PMI_Init") == 0) {
+  if (symbol && strcmp(symbol, "PMI_Init") == 0) {
     return (void *)PMI_Init;
   }
   return _real_dlsym(handle, symbol);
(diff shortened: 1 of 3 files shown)


plugin/batch-queue/rm_pmi.cpp
```diff
@@ -138,6 +138,17 @@ extern "C" int PMI_Init( int *spawned )
     return ret;
 }
 
+// Intel MPI uses dlopen() and dlsym() to find pmi functions.
+// We need to redirect PMI_Init() to our wrapper, so that
+// pmi is correctly initialized.
+extern "C" void *dlsym(void *handle, const char *symbol)
+{
+  if (strcmp(symbol, "PMI_Init") == 0) {
```
rohgarg: Let's change this to:
```C
if (symbol && strncmp(...) == 0) {
  ...
}
```
Do we need to add cases for other `PMI_*` functions? If yes, then how about just returning a handle to the batchqueue plugin in the `dlopen()` call?

jiajuncao: what is the advantage of `strncmp()` over `strcmp()` here?

jiajuncao: We only need to redirect `PMI_Init()` because it is the first pmi function that gets called. And the rm plugin needs to do some initialization. No other wrappers are required.

PR #597, opened 2017-04-27
[BUGFIX] dmtcp_restart ignores -j flag if there no -p flag

planetA:
The help message says that with "-j" flag, dmtcp_restart tries to join
to default port. In fact it fails to do so, unless the port is
specified.

The commit fixes this bug.

### DIFF
--- a/src/dmtcp_restart.cpp
+++ b/src/dmtcp_restart.cpp
@@ -843,7 +843,7 @@ main(int argc, char **argv)
   if ((getenv(ENV_VAR_NAME_PORT) == NULL ||
        getenv(ENV_VAR_NAME_PORT)[0]== '\0') &&
       allowedModes != COORD_NEW) {
-    allowedModes = COORD_NEW;
+    allowedModes = (allowedModes == COORD_ANY) ? COORD_NEW : allowedModes;
     setenv(ENV_VAR_NAME_PORT, STRINGIFY(DEFAULT_PORT), 1);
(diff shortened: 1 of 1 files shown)


planetA:
Here are some examples:

    ../dmtcp/bin/dmtcp_restart -j  ./ckpt_simple_55b24a7c80853818-40000-1167ef2724306.dmtcp     
    [25208] ERROR at coordinatorapi.cpp:450 in startNewCoordinator; REASON='JASSERT(coordinatorListenerSocket.isValid()) failed'
         coordinatorListenerSocket.port() = -1
         (strerror((*__errno_location ()))) = Address already in use
         host = 127.0.0.1
         port = 7779
    Message: Failed to create socket to coordinator port.
    If msg is "Address already in use", this may be an old coordinator.
    Either try again a few seconds or a minute later,
    Or kill other coordinators on this host and port:
        dmtcp_command ---coord-host XXX --coord-port XXX
    Or specify --join-coordinator if joining existing computation.
    dmtcp_restart (25208): Terminating...

To make example work I can do following:

     ../dmtcp/bin/dmtcp_restart -j -p 7779 ./ckpt_simple_55b24a7c80853818-40000-1167ef2724306.dmtcp
    [25037] mtcp_restart.c:413 restore_brk:
      error: new/current break (0x555555762000) != saved break (0x55834a034000)
    Normal exit

With the fix I do not need to pass port number.

Issue #607, opened 2017-07-21
Troubles with GS register


gweodoo:
Hi guys,

I'm still working on DMTCP integration as a C/R system for a custom MPI implementation and I'm focused to the DMTCP TLS mechanism.

TL;DR
--------

GS register does not seem to be handled in a per-thread storage, leading to inconsistent values at restart time.

Explanation
----------------

I saw that TLS save/restore are handled through a `ThreadTLSInfo`, located in a per-thread structure, and `TLSInfo_RestoreTLSState()/TLSInfo_SaveTLSState()` are responsible of updating this `tlsInfo`. `%fs` register seems to be stored in a field named `gdtentrytls`, which is a 1-entry array:

```c
typedef struct _ThreadTLSInfo {
  segreg_t fs, gs;  // thread local storage pointers
  struct user_desc gdtentrytls[1];
} ThreadTLSInfo;
```

This field is updated by `tls_get_thread_area()/tls_set_thread_area()`, the "internals" that `TLSInfo_*` functions are wrapping. Why `%gs` is not handled accordingly ? It seems that this register is stored in a per-process storage: `myinfo_gs`. However, `%gs` being a per-CPU register too, it should be stored the same way as `%fs`. The fact is that we use GS (supposed unused in 64-bit mode) to reproduce a TLS mechanism for our own user-level thread model (like the NPTL does with FS).

Is there any reason for FS and GS to be handled differently ? According to me, these registers could just be saved/restored through `arch_prctl` calls, manipulating 64-bit addresses. Then, DMTCP just needs to update PID/TID in the TCB accordingl (at least for x86_64).

Furthermore, I saw that the `struct user_desc` was used to store the FS register in the `base_addr` member (probably to transparently switch between `arch_prctl()` and `{g,s}et_thread_area()` depending on the architecture). But, on 64-bit systems, this member is an `int`, to stay compliant with the `long` definition in 32-bit mode. This leads `arch_prctl` to erase the next struct member, named `limit`. As these two fields are depicted as "not used" for 64-bit mode, it is not a problem but just to say (actually a bit confusing when debugging).

This post is based on my knowledge of TLS mechanism for x86_64 architecture and complexity introduced by DMTCP should be related to the need to support more than one (like ARM). So, maybe it is non-trivial to do it.

For now, my work-around consists in replacing the memeber with a 2-entry array, to store both the register states and an update of two "internals" accordingly (still on the 2.5 branch). But i think a proper solution would be better.

[...]

PR #609, opened 2017-07-28
Simple workaround to handle GS in a per-thread storage

gweodoo:
Work-around allowing DMTCP to save the GS register in a per-thread manner, like already done for FS (like depicted in issue #607)
This fix is valid for 2.5 branch but would need a little work for being merged with master (conflicts from reformatting stuff).

Feel free to edit if necessary.
Thanks.


gc00:
Thanks @gweodoo for this PR.  We're going to do the DMTCP-2.6.0 release soon, and so I've now pushed this in and am closing this PR.  I added some comments to the commit, but kept you as the author.
Next, I'll port it to DMTCP-3.0 (master) and create a new PR for that.

PR #606, opened 2017-07-18
pid-plugin: Save and restore CPU affinities for threads

rohgarg:
This fixes issue #203 (and possibly even #605).

PR #610, opened 2017-08-02
Allow DMTCP to handle larger SLURM allocation

gweodoo:
In cases where DMTCP relies on interactive allocations (`salloc`), the existing resources can be larger than the application needs for restart. This is not a problem for checkpoint phase but leads to issues, as all nodes are considered to be used for restarting the program.

This fix intends to downcount from available restart nodes, each node having a number of slots equal to zero (not targeted for restarting program instance). It then ensures that the remote DMTCP launcher will be executed only once per node (for each non-empty slots). A newline also needs to be added to `dmtcp_discover_rm` for bash integer evaluation.

Please, feel free to edit if needed.


src/restartscript.cpp
```diff
@@ -462,8 +462,11 @@ void writeScript(const string& ckptDir,
                   "      fi\n"
                   "      export DMTCP_REMLAUNCH_NODES=$DMTCP_REMLAUNCH_NODES\n"
                   "      bound=$(($DMTCP_REMLAUNCH_NODES - 1))\n"
+		  "      srun_ntasks=$DMTCP_REMLAUNCH_NODES\n"
```
jiajuncao: ```DMTCP_REMLAUNCH_NODES``` is the total number of nodes, not the total number of tasks, right?

src/restartscript.cpp
```diff
@@ -490,7 +493,7 @@ void writeScript(const string& ckptDir,
                   "      else\n"
                   "        DMTCP_REMLAUNCH_0_0=\"$DMTCP_REMLAUNCH_0_0"
                                                      " $DMTCP_LAUNCH_CKPTS\"\n"
-                  "        $srun_path \"$llaunch\"\n"
+                  "        $srun_path --nodes=$srun_ntasks --ntasks=$srun_ntasks \"$llaunch\"\n"
```
jiajuncao: Does this assume each node has one task only?

jiajuncao:
@gweodoo Thanks for the patch.

I have some inline comments, and one general concern:

The current rm plugin assumes that the restart allocation is the same as that of before checkpoint. That means if ```N``` nodes and ```M``` processors per node are allocated before checkpoint, it should be exactly the same after restart. The main reason is that if we allow a different combination, inter-process communication may not work correctly after restart. For example, if two MPI ranks reside in the same node before checkpoint, most likely they will communicate using shared memory. But if they are on different nodes after restart, shared memory won't work, and it is hard to change the IPC channel dynamically.

In your case, I think what you want is one task per node. Is that correct? If so, your patch works perfectly. But if the user wants to have more than one processor per node, we need to handle that case carefully as well.

src/restartscript.cpp
gweodoo: Yes and if I'm not wrong, it is the expected DMTCP behavior. At restart phase, DMTCP needs one process per node and handles the `fork()`step to recreate per-node process topology. So here, we ensure that the number of needed processes is equal to the number of nodes.

But maybe I misunderstood what DMTCP is doing at restart. Please correct me if I'm wrong

gweodoo:
Thanks for the reply @jiajuncao 

Sure, if you consider the restart allocation is the same as the one before, no need to fix it. But consider the case where the allocation is made once (before launching the application). I allocate 4 nodes and then I launch my application only on 2 of these nodes. When trying to restart, I should have to free resources to allocate exactly two nodes for the restart to process.

Maybe the `dmtcp_discover_rm`could cooperate to provide the number of root processes are existing on each node (how many processes per node have been created by the launcher). I think I misunderstood the process topology re-creation for DMTCP. It seems that only child processes are concerned, right ? Not the PM-forked ones, and my mistake is probably here.

jiajuncao:
@gweodoo In that case, I think you can slightly revise your patch to make it work: instead of skipping nodes with zero slots, you can count the total number of tasks, and use that as the parameter of srun. You can also add the total number of nodes, if needed.

And about the topology of creation:
dmtcp is concerned with the actual compute processes only. Each compute process will find the right checkpoint image to restart from.

gc00:
Hi @jiajuncao,
You wrote:
> And about the topology of creation:
dmtcp is concerned with the actual compute processes only. Each compute process will find the right checkpoint image to restart from.

    If I understand correctly, when DMTCP is used with MPI, then on
restart, it does not try to re-build a user-specified topology.
Instead, it will use the default topology of that MPI.  Am I
expressing this correctly?

Thanks, - @gc00

gweodoo:
As you suggested, @jiajuncao, I added the distinction between nodes and tasks (from SLURM's point of view). Thus, nodes are constrained by the number of `*_SLOTS` greater than 1 and for each non-zero slots, the number of tasks is added by the number of slots. This solution seems good to me for handling **larger node** allocation, detailed by the pseudo-output below, where : 
* checkpoint is made in a "**2**-node, 2-process-per-node" allocation;
* restart is made in a "**4**-node, 2-process-per-node" allocation.
```bash
RES_MANAGER=SLURM
manager_resources="*node001:2 node002:2 node003:2 node004:2 "
input_config="node101:2 node102:2 "
DMTCP_DISCOVER_PM_TYPE='UNKNOWN'
DMTCP_LAUNCH_CKPTS=''
DMTCP_REMLAUNCH_NODES=4
DMTCP_REMLAUNCH_0_SLOTS=2
DMTCP_REMLAUNCH_0_0='~/dmbuild/ckpt_sleep_7c546cd41ea30913-61000-3ed5890af2a8b.dmtcp'
DMTCP_REMLAUNCH_0_1='~/dmbuild/ckpt_sleep_7c546cd41ea30913-60000-3ed5890a2a601.dmtcp'
DMTCP_REMLAUNCH_1_SLOTS=2
DMTCP_REMLAUNCH_1_0='~/dmbuild/ckpt_sleep_7c546cd41ea30915-63000-3ed401b8798d7.dmtcp'
DMTCP_REMLAUNCH_1_1='~/dmbuild/ckpt_sleep_7c546cd41ea30915-62000-3ed401b878068.dmtcp'
DMTCP_REMLAUNCH_2_SLOTS=0
DMTCP_REMLAUNCH_3_SLOTS=0
/usr/bin/srun --nodes=2 --ntasks=4 $DMTCP_INSTALL_PREFIX/bin/dmtcp_rm_loclaunch
```

But this fix can't handle "**greater process-per-node**" allocation. For instance, if at restart, I allocate two node in both cases but with 4 processes per node at restart (instead of 2), some troubles occur:
```bash
RES_MANAGER=SLURM
manager_resources="*node001:4 node002:4 "
input_config="node101:2 node102:2 "
DMTCP_DISCOVER_PM_TYPE='UNKNOWN'
DMTCP_LAUNCH_CKPTS=''
DMTCP_REMLAUNCH_NODES=4
DMTCP_REMLAUNCH_0_SLOTS=4
DMTCP_REMLAUNCH_0_0='~/dmbuild/ckpt_sleep_7c546cd41ea30913-61000-3ed5890af2a8b.dmtcp'
DMTCP_REMLAUNCH_0_1='~/dmbuild/ckpt_sleep_7c546cd41ea30913-60000-3ed5890a2a601.dmtcp'
DMTCP_REMLAUNCH_0_2='~/dmbuild/ckpt_sleep_7c546cd41ea30915-63000-3ed401b8798d7.dmtcp'
DMTCP_REMLAUNCH_0_3='~/dmbuild/ckpt_sleep_7c546cd41ea30915-62000-3ed401b878068.dmtcp'
DMTCP_REMLAUNCH_1_SLOTS=0

/usr/bin/srun --nodes=1 --ntasks=4 $DMTCP_INSTALL_PREFIX/bin/dmtcp_rm_loclaunch
```
Processes are gathered on the first node instead of reproducing the initial topology. This should not be a terrible issue as processes should still be able to communicate (if internode communication is working, then intranode works as well), but it seems wrong to let such things happen anyway. 

Out of curiosity, I tried with a "4-node, 1-process-per-node" restart allocation (on-purpose wrong pattern), and surprisingly, it produces the same topology as the issue above, and continue the restart process (even if a clear warning lets you know that what you are doing is bad). Should DMTCP be more permissive for this kind of tricky case ?

Thanks.

jiajuncao:
@gweodoo In the second case you were describing (2 nodes * 4 processes per node), are you sure ```DMTCP_REMLAUNCH_NODES``` is ```4```? It looks like the code should produce ```2``` instead.

Anyway, I took a deep look at the code today (it was written by another former member from our team), and you should be right about the topology:

The RM plugin has the information of the topology of both before checkpoint and after restart. On restart, it iterates through the available slots, and put the original slots into the new ones. I think in this manner the plugin will consolidate the processes, if the amount of available slots is more than the original one. Just like you said, most inter-node communication should work on a single machine, whether it's TCP over Ethernet, or HPC network fabrics such as InfiniBand. The case of less processes per node however is risky, since shared memory won't work across nodes.

gweodoo:
@jiajuncao, yes, it should be `DMTCP_REMLAUNCH_NODES=2` (failed "search-and-replace").

In any case, we cannot presuppose the application behavior. In my opinion, it is hard to define a "default" behavior. For instance, with per-node process over-subscription, we cannot just suppose that because "intranode will work as internode is working", so there will be no troubles. By gathering processes on the same node, you constrain each process CPU set, leading to overlapping when the application (or linked libraries) set CPU affinities (even if nobody really check if `*_setaffinity` succeeded). At best, it will produce performance decrease and at worst, the application can crash at restart. DMTCP could detect that because of PR #606, reapplying CPU affinity for each thread.

On the contrary, with per-node process under-subscription, we cannot suppose that we are in an MPI context and that network is a requirement. For instance, if your application is a Map-Reduce-like algorithm, where data are gathered through the file system, each process could be independent and there would be no issue to scatter multiple processes over multiple nodes at restart.

I think the user has to know what it is doing by providing a different resource allocation (maybe a short tutorial could be done to help with that). But it remains a huge flexibility to not constrain DMTCP to such specific usage as MPI in distributed environments. According to me, the best DMTCP can do is a "deep" analysis of the restarted allocation and check if the initial one can be mapped with troubles. In any other cases, a clear detailed warning should be displayed to let the user know that the restart will be at its own risks.

But all of this is far beyond the PR !
Thanks,

gc00:
@jiajuncao ,  you know this code best.  Are we falling into the case where "the perfect is the enemy of the good"?  In other words, is there a version of this PR (maybe with some modifications and a FIXME comment) that could be pushed in now?  Thanks.

PR #615, opened 2017-08-11
Correctly restore virtual pty numbering and ids

rohgarg:


jiajuncao:
@rohgarg Thanks for the patch. Could you please add some comments about why this patch is required? I think a short description of the bug scenario would be enough. Apart from that, it looks good to me.

gc00:
@rohgarg, I agree with @jiajuncao.  Could you add some comments about why this patch is required?  After that, @jiajuncao or I can approve it.  Thanks.

rohgarg:
A bug was observed where a program that opens new pseudo-terminals post
restart, ends up getting assigned the incorrect virtual ptys.  The root
cause was that DMTCP wouldn't save the last virtual pty id at checkpoint
time; as a result, on restart, when the application tried to open a new
pty, it ended up with an already used virt pty. Any subsequent calls by
the application to use the virt pty would end up getting translated to
the incorrect real pty id.

src/plugin/ipc/file/ptyconnlist.cpp
```diff
@@ -61,6 +63,7 @@ PtyConnList::instance()
 void
 PtyConnList::drain()
 {
+  virtPtyId = SharedData::getVirtualPtyId();
```
jiajuncao: Do all processes on a same node return the same `virtPtyId`? Also, could you rename it to something like `nextVirtPtyId`, if that is what you intended? And the same for the function names.

rohgarg: The processes request for a new virtual pty id through existing `SharedData` API. The problem is that we don't save and restore the last virt pty id that was in-use at checkpoint time, and so on restart, we end up reallocating in-use virt pty ids for any new ptys.

The intent here is *not* to get a new virt pty id; rather, it is to save and restore the last used
virt pty id. This is to ensure that when a process requests for a new pty post restart, it must not reuse any of the pre-exiting, in-use ids.

PR #616, opened 2017-08-11
ptyconnection: Use _fcntlFlags instead of _flags

rohgarg:


jiajuncao:
It looks like the ```_flags``` member is not used at all, is it?

gc00:
There was no description.  But if my memory is correct, this is to avoid a spurious error condition.  Probably, this is needed for the pty PR that's coming.  Please correct me if i'm wrong.

rohgarg:
The `_flags` is obsolete and is not used/updated anywhere. As a result, the code that still references the field does not correctly restore the attributes of the pty object. While this has been fixed for the `FileConnection` class, the `PtyConnection` class still had references to the field.

PR #617, opened 2017-08-14
fileconnection: Correctly handle relative filepaths in checkDup()

rohgarg:


src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -505,13 +505,16 @@ FileConnection::postRestart()
...
-FileConnection::checkDup(int fd)
+FileConnection::checkDup(int fd, const char *npath)
 {
   bool retVal = false;
 
   int myfd = _fds[0];
 
-  if (lseek(myfd, 0, SEEK_CUR) == lseek(fd, 0, SEEK_CUR)) {
+  string fullPath = jalib::Filesystem::GetDeviceName(myfd);
+  string fpath = string(npath);
```
jiajuncao: Can we use ```string fpath(npath);``` instead to reduce one call to the copy constructor?

rohgarg:
@jiajuncao: `filePath()` returns `_path`. The `_path` field is set when the file connection object is created. Therefore, it equals whatever path argument was used for open, which could be a relative path.

In this case, the check `fcon->filePath() == npath`. This is because `npath`, derived from proc-fd, always refers to an absolute path.

PR #618, opened 2017-08-20
Change _cachedHost to a string object.

jiajuncao:
If ```host``` is another local variable, we have the use-after-free risk.

### DIFF
--- a/src/util_init.cpp
+++ b/src/util_init.cpp
@@ -48,7 +48,7 @@ void Util::getCoordHostAndPort(CoordinatorMode mode,
   }
 
   static bool _firstTime = true;
-  static const char *_cachedHost;
+  static string _cachedHost;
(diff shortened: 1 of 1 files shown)


gc00:
I've forgotten the motivation, but if I remember correctly, the issue was that the previous code would leave a global variable set to a string, and the string destructor would be called at the wrong time when a process exits.  Is this it?

gc00:
@jiajuncao, Ping.  :-)  (And see my note about reading my previous comment and correcting it if I am wrong on that.)

jiajuncao:
@gc00 in the previous code, the pointer `_cachedHost` points to a character string that belongs to another local scope. The string is destructed when execution runs out of the scope, and then `_cachedHost` points to a garbage value.

PR #622, opened 2017-09-04
Remove gcc warning

gc00:
This should be trivial to review.  It's only to make a gcc warning go away.


gc00:
Cancelling this.  It should have been to pull into 2.5 branch.

PR #631, opened 2017-09-26
change the include file order for Fedora

Twinklejain:
GNU Readline Library -> Programming with GNU Readline
https://cnswww.cns.cwru.edu/php/chet/readline/readline.html 
The order of the includes failed on the Fedora and the new order of the include files is consistent with the GNU manual.


gc00:
I worked with @Twinklejain on this.  The change simply re-orders the ordering of the include files.  It follows the GNU example.  So, this seems safe.  Approved.

PR #632, opened 2017-09-28
Do not reset the NS (publish-subscribe) database on checkpoint/resume.

jiajuncao:
This is necessary when one wants to use the publish/subscribe service before checkpoint, and throughout checkpoint and resume.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -470,7 +470,7 @@ void DmtcpCoordinator::updateMinimumState(WorkerState::eWorkerState oldState)
       exitAfterCkptOnce = false;
     } else {
       JNOTE ( "building name service database" );
-      lookupService.reset();
+      // lookupService.reset();
       broadcastMessage ( DMT_DO_REGISTER_NAME_SERVICE_DATA );
     }
   }


rohgarg:
@jiajuncao: I'm not sure I understand this scenario. Could you please elaborate a bit?

jiajuncao:
@rohgarg When the plugin publishes a mapping before checkpoint, i.e., at launch time in the case of the IB plugin, the plugin wants to preserve the mapping through checkpoint and resume. If we reset the database on resume, the mapping is gone. In the new design of IB plugin, IDs are virtualized at launch time, and the plugin may query the database after resume (not restart).

rohgarg:
Okay, do you happen to know the reason we were resetting the database at resume time?

jiajuncao:
No. That code has been there for several years.

gc00:
@rohgarg, Implicitly, you were managing this review.  Is the PR acceptable now?

rohgarg:
@karya0: Could you please review this change? If you have no objections, we'd like to push this in.

gc00:
The github automatic merge pushed this commit into dmtcp:2.5.  The dmtcp:2.5 branch is obsolete now.  So, I cherry-picked it and pushed it into the dmtcp:2.6 branch.  Next, I will cherry-pick it and push into dmtcp:master (assuming that the 'master' code is compatible with this).

At some point, we should do a force push `git push -f origin 2.5^:2.5`.  Otherwise, the 2.6 branch will no longer be an extension of the 2.5 branch.  But that requires temporarily disabling the github hooks.

A similar mistake had occurred with PR #632.  It was automatically pushed into the 2.5 branch by @rohgarg, and then cherry-picked into 2.6.  (And I should check if it was cherry-picked into the master branch.)

So, really, we need to do:  `git push -f origin 2.5^^:2.5`

karya0:
Let's create a tag "2.5" from current 2.5^^ and then delete the 2.5 branch. That way we'll have a reference and no more future commits.

karya0:
@gc00: The 2.5.2 and 2.4.9 tags were up to date, so I have deleted the 2.4 and 2.5 branches.

gc00:
Hi Kapil,
    Can we bring back the 2.5 branch?
    The DMTCP-2.6.0 release has not yet happened, and I think there are some people at Intel that are directly using our 2.5 branch.  That's simpler than asking them to switch to a tag for 2.5.2 (and add the one extra commit on the 2.5 branch manually), and then to switch to the 2.6 branch in a week or two.

Thanks,
- Gene

karya0:
I am confused. Didn't we want to revert the two commits that were recently merged in 2.5? Without those, it's essentially the same as 2.5.2. In any case, I have created the 2.5 branch. We'll delete it once 2.6 is released.

gc00:
> Let's create a tag "2.5" from current 2.5^^ and then delete the 2.5 branch. That way we'll have a reference and no more future commits.
Fair enough.  We're close to releasing on the 2.6 branch anyway.

gc00:
And DMTCP-2.6.0 is now released.  It's safe to delete the 2.5 branch again.  :-)
Thanks,
- Gene

PR #634, opened 2017-09-28
Ib id full virtualization

jiajuncao:
New design for IB plugin.  This PR is for the 2.x branch.  PR #664 is a port of this PR for the 3.0/master branch.

Use eager virtualization for IDs, i.e., IDs are virtualized at the time they are created. Also, unique virtual IDs are generated based on the virtual PID. Therefore, no virtual ID conflict would happen anymore, even if the application caches certain IDs before checkpoint, and registers new ones after restart.


gc00:
This PR has now been pushed into the 2.6 branch.  So, we can close the request to push into the 2.5 branch.

PR #591, opened 2017-04-08
Fixed tls_pid_offset for newer (>2.24) glibc.

karya0:
Enhances 2be93da0af12ca496b3118db8d815ff796be7965

PR #640, opened 2017-10-10
Fixed tls_pid_offset for newer (>2.24) glibc.

karya0:
Enhances 2be93d.

Based on #591.


karya0:
@rohgarg @gc00: Merging this since #591 has already been approved.

PR #645, opened 2017-10-25
pid: Fix an error introduced in 30c03f23

rohgarg:
Apparently, on some old distros, the code fails to compile without this
include file. *ALSO, which is commit 30c03f23 ?  I don't find it anywhere.  Is there a corresponding PR?*

### DIFF
--- a/src/plugin/pid/pid.cpp
+++ b/src/plugin/pid/pid.cpp
@@ -20,6 +20,12 @@
  ****************************************************************************/
 
 #include <sys/types.h>
+
+#ifndef _GNU_SOURCE
+# define _GNU_SOURCE
+#endif // ifndef _GNU_SOURCE
+#include <sched.h>
+
 #include "jalloc.h"
 #include "jassert.h"
 #include "jconvert.h"


rohgarg:
> We set `GNU_SOURCE` in pidwrappers.h, but that is included after sched.h. So, it's not set for `sched.h`. Could you take a look at that issue?

I added `#define _GNU_SOURCE` before include sched.h, but my compiler complains about redefining it. So I had assumed that it must have been inherited from an earlier header file.

> Also, do we need this only in 2.5.2, or also in 2.4.9 and 3.0?

We only need this for 2.5.2. The other branches don't have this change. (Porting this two 3.0 is actually a lot of work.)

gc00:
@rohgarg , Use your judgement, but here are two comments:

1.  If we don't have this functionality in mastter/3.0, then we need to add an issue in github.
2.  We could set:
`#ifndef _GNU_SOURCE`
`# define _GNU_SOURCE`
`#endif`<br>
before `#include <sched.h>`<br>
Then in `pidwrappers.h`, we can change the `_GNU_SOURCE` to a copy of the above excerpt.

PR #650, opened 2017-10-25
Elim bad warning: Failed to find coord IP address.

gc00:
There are now two commits.

**FIRST COMMIT:**
dmtcp_coordinator.cpp was giving a warning:  `Failed to find coordinator IP address.`, and then not setting the IP address of the coordinator.  This forced the clients to use the original hostname.

The issue is that we use `gethostname()` to get the hostname, and then `getaddrinfo()` to get the IP address, and then `getnameinfo()` to get the name again.  When name and hostname didn't match (even though they have the same IP address), we refused to set the IP address and issued a warning.

In recent distros, it's common for the same IP address to have two hostname aliases.  So, now we use `gethostname()` as the hostname, and the common IP address is used.

The same patch is applied both to `dmtcp_coordinator.cpp` and to `src/plugin/ipc/ssh/ssh.cpp`, since the same code excerpt is used in corresponding functions in those two files.

**SECOND COMMIT:**
    If hostname is missing domainname, try using getaddrinfo with hints.ai_canonname.  This is the modern way to do what we used to do long ago with the now deprecated gethostbyname().

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -1263,6 +1263,28 @@ calcLocalAddr()
   hints.ai_addr = NULL;
   hints.ai_next = NULL;
 
+  // FROM: Wikipedia:CNAME_record:
+  //  When a DNS resolver encounters a CNAME record while looking for a regular
+  //  resource record, it will restart the query using the canonical name
+  //  instead of the original name. (If the resolver is specifically told to
+  //  look for CNAME records, the canonical name (right-hand side) is returned,
+  //  rather than restarting the query.)
+  hints.ai_flags |= AI_CANONNAME;
+  error = getaddrinfo(hostname, NULL, &hints, &result);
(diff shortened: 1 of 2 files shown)


src/dmtcp_coordinator.cpp
```diff
@@ -1282,18 +1283,29 @@ calcLocalAddr()
       if (error != 0) {
         JTRACE("getnameinfo() failed.") (gai_strerror(error));
         continue;
-      }
-      if (Util::strStartsWith(name, hostname) ||
-          Util::strStartsWith(hostname, name)) {
+      } else {
+        at_least_one_match = true;
```
rohgarg: Why do we set this to true here? Shouldn't we be doing this in the if clause later?

```
if ( Util::strStartsWith(name, hostname) &&
     Util::strStartsWith(hostname, name) ) {
   ...
}
```

gc00:
The commit for 2.5 is now pushed in.

PR #374, opened 2016-05-03
dmtcp_restart: Add a debug option to enable auto gdb attach

rohgarg:

PR #652, opened 2017-10-27
Fix warnings in util_misc.cpp

gc00:
Fix a gcc warning that was caused by PR #374.


gc00:
The commit for 2.5 is now pushed in.

PR #656, opened 2017-11-15
Duplicate of PR #618 for 3.0.

jiajuncao:


gc00:
PR #757 is a copy of this PR.  The PR will be pushed in as PR #757 in order to preserve the author @jiajuncao on the original commit.  So, this PR can be closed now.

PR #657, opened 2017-11-15
Duplicate of PR #633 for 3.0.

jiajuncao:


karya0:
@jiajuncao @gc00: Could you add a little more description to the PR? I didn't get it. To me, it looks like the dlsym wrapper is not doing anything except forwarding the call to the next library. Wouldn't it have gone to the next library anyways?

PR #658, opened 2017-11-15
dmtcpplugin: Make dmtcp_get_restart_env handle env of any size

rohgarg:


src/dmtcpplugin.cpp
```diff
@@ -361,23 +361,32 @@ dmtcp_get_restart_env(const char *name,   // IN
...
     close(env_fd);
     return RESTART_ENV_NULL_PTR;
   }
 
+  struct stat statbuf;
+  size_t size = RESTART_ENV_MAXSIZE;
+  char *env_buf = NULL;
+
+  if (fstat(env_fd, &statbuf) == 0) {
+    if (statbuf.st_size > RESTART_ENV_MAXSIZE) {
```
karya0: Why even bother about `RESTART_ENV_MAXSIZE` here? We can just kill it entirely and continue with the size computation.

karya0 (approved):

PR #661, opened 2017-11-24
socket-plugin: Add support for robust detection of external sockets

rohgarg:
This is a port of commit 98002e7e from the master branch to the 2.5 branch.


src/plugin/ipc/socket/socketconnection.cpp
```diff
@@ -300,6 +300,80 @@ bool TcpConnection::isBlacklistedTcp(const sockaddr* saddr, socklen_t len)
...
+                                              &value, &vallen);
+    if (ret != 0) {
+      JASSERT(vallen == sizeof(value))(vallen)(sizeof(value));
+    } else {
+      JWARNING(false) (_fds[0])
+       .Text("DMTCP detected an \"external\" connect socket."
+             "The socket will be restored as a dead socket.");
+      markExternalConnect();
+    }
+  }
```
jiajuncao: Do we need to do the same check for TCP_ACCEPT sockets? In case some external socket connects to a local one?

gc00:
I made the change (adding the "subsection" heading in comments).  Next, I'll push in the two commits (retaining @rohgarg as author) manually.  So, I'm closing this PR now.

PR #663, opened 2017-12-07
fifoconnection: Fix serialization of vector

rohgarg:
It seems like the vector `_in_data` was using the incorrect serialization
primitive. This would cause problems when forking a 32-bit process from a
64-bit process. Thanks to R Balaji for catching this bug and the fix.


karya0:
I am not sure if I understand the change. The `serializeVector` should have been called anyways as a result of `& _in_data`.

rohgarg:
@karya0 I couldn't find where we overload the `&` operator for vectors.

karya0:
I think we can close this in lieu of https://github.com/dmtcp/dmtcp/pull/666

PR #665, opened 2017-12-29
Reduced verbosity of several test programs.

karya0:
The file tests were filling up the disk space at an alarming rate
especially when coupled with --slow and caused CI tests to fail if the
test box had lower disk space.

Similarly, reduced the frequency of unwanted printfs from some test
cases to allow for better debugging with manual launches.

### DIFF
--- a/test/dmtcp2.c
+++ b/test/dmtcp2.c
@@ -5,9 +5,14 @@ int
 main(int argc, char *argv[])
 {
   int count = 1;
+  FILE *fp = fopen("/dev/null", "w");
+  if (fp == NULL) {
+    printf("fopen(/dev/null, ...) failed! \n");
+    exit(1);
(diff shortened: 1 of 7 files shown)


test/autotest.py
```diff
@@ -739,7 +739,7 @@ def saveResultsNMI():
 # fail, causing the test to fail.
 old_ckpt_cmd = CKPT_CMD
 CKPT_CMD = 'xc'
-runTest("syscall-tester",  1, ["./test/syscall-tester"])
+#runTest("syscall-tester",  1, ["./test/syscall-tester"])
```
rohgarg: Why did you comment this test?

karya0: Thanks for the catch- fixed now.

PR #674, opened 2018-02-08
Fixed code to add MAP_GROWSDOWN flag for stack region

Twinklejain:
The stack region of user-application doesn't grow automatically on the restart in the absence of MAP_GROWSDOWN mmap flag while remapping user-applications' memory regions. The current version of DMTCP gives segmentation fault if the user application tries to extend the stack downwards by touching any address within guard page region.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -50,7 +50,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
-#include <unistd.h>
+#include <stddef.h>
 
 #include "../membarrier.h"
 #include "config.h"
@@ -102,6 +102,7 @@ typedef struct RestoreInfo {
   VA vdsoEnd;
   VA vvarStart;
   VA vvarEnd;
+  VA endOfStack;
   fnptr_t post_restart;
   fnptr_t post_restart_debug;
   // NOTE: Update the offset when adding fields to the RestoreInfo struct
@@ -130,8 +131,8 @@ typedef struct RestoreInfo {
 static RestoreInfo rinfo;
 
 /* Internal routines */
-static void readmemoryareas(int fd);
-static int read_one_memory_area(int fd);
+static void readmemoryareas(int fd, VA endOfStack);
+static int read_one_memory_area(int fd, VA endOfStack);
 #if 0
 static void adjust_for_smaller_file_size(Area *area, int fd);
 #endif /* if 0 */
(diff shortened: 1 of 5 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -975,6 +975,20 @@ read_one_memory_area(int fd)
...
   }
+  /* MAP_GROWSDOWN flag is required for stack region on restart to make
+   * stack grow automatically when application touches any address within the
+   * guard page region(usually, one page less then stack's start address).
+   *
+   * The end of stack is detected dynamically at checkpoint time. See
+   * prepareMtcpHeader() in threadlist.cpp and ProcessInfo::growStack()
+   * in processinfo.cpp.
+   */
+  if (area.name[0] && mtcp_strstr(area.name, "stack")
```
karya0: Why do we compare the names here? Isn't endOfStack a more definitive way to find stack region?

src/mtcp/mtcp_restart.c
```diff
@@ -457,9 +459,9 @@ restart_fast_path()
   // rinfo.restorememoryareas_fptr(&rinfo);
   // Intel icc-13.1.3 output uses register rbp here.  It's no longer available.
   asm volatile(
-   // 104 = offsetof(RestoreInfo, rinfo.restorememoryareas_fptr)
+   // 112 = offsetof(RestoreInfo, rinfo.restorememoryareas_fptr)
```
gc00: With the recent commit by @karya0 , you may find that this comment and assembly code has disappeared.  If so, that's fine -- not a problem.

karya0:
This PR (or a variation) is still needed to handle user stacks. PR #712 simply handles the temp stack.

We need to enhance this PR not only to add MAP_GROWSDOWN to main stack but also to stack regions of all other threads. The approach is this PR doesn’t scale to multiple threads so we might be better off putting mmap flags in Area struct itself.

JainTwinkle:
@karya0  Correct me if I am wrong but a thread's stack size is fixed at the time of thread creation. Only the main thread can dynamically grow its stack. (See [man pthread_attr_setstacksize](https://linux.die.net/man/3/pthread_attr_setstacksize)) I don't think MAP_GROWSDOWN flag is needed for all user-threads except the main thread.

gc00:
@karya0, ping!

I think @Twinklejain makes a good point.  Only the stack of the primary thread can grow.  The stack of a secondary thread should not go.  Am I missing something?  (It would be nice to resolve this old PR.)

karya0:
@JainTwinkle: Can we add  `MAP_GROWSDOWN` to `area.flags` for the stack region during checkpoint itself? I think that would simplify the restart logic. Also, can you please rebase?

gc00:
@JainTwinkle, this commit has been pending for a long time.  It's a really valuable fix.  I'm happy if you push this in now, as it is (after resolving conflicts).

I'm happy with it as is -- meaning that only the primary thread has the attribute `MAP_GROWSDOWN`.  This is the default in Linux.  You could either use @karya0's suggestion to add `MAP_GROWSDOWN' to `area.flags` at this time, or else just push in the commit now, and create an issue (a "TODO" item) to add to `area.flags` for the unusual case in which a user has artificially created a new stack with `MAP_GROWSDOWN`.

gc00:
A version of this has now been pushed into the 2.6.0 branch.  So, I'm removing the 2.6.0 label.  This now only needs to be completed for 3.0/master.

gc00:
NOTE for others:  @Twinklejain will be reviewing her PR further, before deciding to push this in.

PR #682, opened 2018-03-10
Adding Unix file-socket capability to DMTCP

pgrosu:
This add Unix file-socket capability for DMTCP when a network is not available.

Documentation detailing the reasoning: [Checkpointing_The_Initialization_Procedure.pdf](https://www.dropbox.com/s/6fl0xqeo31nnj58/Checkpointing_The_Initialization_Procedure.pdf?dl=1)

### DIFF
--- a/jalib/jsocket.cpp
+++ b/jalib/jsocket.cpp
@@ -41,6 +41,8 @@
 #include "jassert.h"
 #include "jsocket.h"
 
+#include "../src/constants.h"
+
 const jalib::JSockAddr jalib::JSockAddr::ANY(NULL);
 
 jalib::JSockAddr::JSockAddr(const char *hostname /* == NULL*/,
@@ -152,25 +154,52 @@ jalib::JSockAddr::JSockAddr(const char *hostname /* == NULL*/,
 
 jalib::JSocket::JSocket()
 {
-  _sockfd = jalib::socket(AF_INET, SOCK_STREAM, 0);
+
+	if( jalib::JSocket::isFileSocketDefined() ) {
+
+		memset( &FILE_SOCKET_ADDR, 0, sizeof(struct sockaddr_un) );
+		FILE_SOCKET_ADDR.sun_family = AF_UNIX;
+		strncpy( FILE_SOCKET_ADDR.sun_path, getenv(ENV_VAR_DMTCP_FILE_SOCKET_PATH), sizeof(FILE_SOCKET_ADDR.sun_path) - 1 );
+		_sockfd = jalib::socket(AF_UNIX, SOCK_STREAM, 0);
+	} else {
+    _sockfd = jalib::socket(AF_INET, SOCK_STREAM, 0);
+	}
 }
+
+bool jalib::JSocket::isFileSocketDefined() {
+  bool isDefined = false;
+  if ( getenv( ENV_VAR_DMTCP_FILE_SOCKET_PATH ) != NULL ) {
+        isDefined = true;
+  }
+  return isDefined;
+}
+
+char * jalib::JSocket::getFileSocketPath() {
+  return getenv( ENV_VAR_DMTCP_FILE_SOCKET_PATH ); // requires #include "../src/constants.h"
+}
+
 
 bool
 jalib::JSocket::connect(const JSockAddr &addr, int port)
 {
   bool ret = false;
 
-  // jalib::JSockAddr::JSockAddr used -2 to poison port (invalid host)
-  if (addr._addr->sin_port == (unsigned short)-2) {
-    return false;
-  }
-  for (unsigned int i = 0; i < addr._count; i++) {
-    ret = JSocket::connect((sockaddr *)(addr._addr + i),
-                           sizeof(addr._addr[0]),
-                           port);
-    if (ret || errno != ECONNREFUSED) {
-      break;
-    }
+  if( !jalib::JSocket::isFileSocketDefined() ) {
+		// jalib::JSockAddr::JSockAddr used -2 to poison port (invalid host)
+		if (addr._addr->sin_port == (unsigned short)-2) {
+			return false;
+		}
+		for (unsigned int i = 0; i < addr._count; i++) {
+			ret = JSocket::connect((sockaddr *)(addr._addr + i),
+														 sizeof(addr._addr[0]),
+														 port);
+			if (ret || errno != ECONNREFUSED) {
+				break;
+			}
+		}
+		return ret;
+  } else {
+  	return ::connect(_sockfd, (const struct sockaddr *) &FILE_SOCKET_ADDR, sizeof(struct sockaddr_un));
   }
   return ret;
 }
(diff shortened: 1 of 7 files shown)


pgrosu:
And I did this from Dekaksi following the instructions from here:  https://github.com/dmtcp/dmtcp/wiki/DMTCP-GIT-Cheatsheet :)

jalib/jsocket.cpp
```diff
@@ -152,25 +154,52 @@ jalib::JSockAddr::JSockAddr(const char *hostname /* == NULL*/,
...
+		strncpy( FILE_SOCKET_ADDR.sun_path, getenv(ENV_VAR_DMTCP_FILE_SOCKET_PATH), sizeof(FILE_SOCKET_ADDR.sun_path) - 1 );
+		_sockfd = jalib::socket(AF_UNIX, SOCK_STREAM, 0);
+	} else {
+    _sockfd = jalib::socket(AF_INET, SOCK_STREAM, 0);
+	}
+}
+
+bool jalib::JSocket::isFileSocketDefined() {
+  bool isDefined = false;
+  if ( getenv( ENV_VAR_DMTCP_FILE_SOCKET_PATH ) != NULL ) {
```
karya0: Why aren't we validating the contents of the environment variable?

jalib/jsocket.cpp
```diff
@@ -152,25 +154,52 @@ jalib::JSockAddr::JSockAddr(const char *hostname /* == NULL*/,
...
+        isDefined = true;
+  }
+  return isDefined;
+}
+
+char * jalib::JSocket::getFileSocketPath() {
+  return getenv( ENV_VAR_DMTCP_FILE_SOCKET_PATH ); // requires #include "../src/constants.h"
```
karya0: Same as above.

jalib/jsocket.cpp
```diff
@@ -152,25 +154,52 @@ jalib::JSockAddr::JSockAddr(const char *hostname /* == NULL*/,
...
+			ret = JSocket::connect((sockaddr *)(addr._addr + i),
+														 sizeof(addr._addr[0]),
+														 port);
+			if (ret || errno != ECONNREFUSED) {
+				break;
+			}
+		}
+		return ret;
+  } else {
+  	return ::connect(_sockfd, (const struct sockaddr *) &FILE_SOCKET_ADDR, sizeof(struct sockaddr_un));
```
karya0: For blocks like these (here and below), it might be better to have something like:
```if (jalib::JSocket::isFileSocketDefined()) {
  return ::connect(_sockfd, (const struct sockaddr *) &FILE_SOCKET_ADDR, sizeof(struct sockaddr_un));
}```

That way the rest of the code block stays the same.

File: jalib/jsocket.cpp [hunk 4]
```diff
@@ -41,6 +41,8 @@
 #include "jassert.h"
 #include "jsocket.h"
 
+#include "../src/constants.h"
```
karya0: s:../src/::

src/constants.h
```diff
@@ -89,6 +89,8 @@
 #define ENV_VAR_QUIET               "DMTCP_QUIET"
 #define ENV_VAR_DMTCP_DUMMY         "DMTCP_DUMMY"
 
+#define ENV_VAR_DMTCP_FILE_SOCKET_PATH         "DMTCP_FILE_SOCKET_PATH"
```
karya0: Fix the alignment.

src/coordinatorapi.cpp
```diff
@@ -348,6 +348,7 @@ sendMsgToCoordinatorRaw(int fd,
                         const void *extraData,
                         size_t len)
 {
+
```
karya0: Why this newline?

src/coordinatorapi.cpp
```diff
@@ -358,6 +359,7 @@ sendMsgToCoordinatorRaw(int fd,
   if (extraData != NULL) {
     JASSERT(Util::writeAll(fd, extraData, len) == len);
   }
+
```
karya0: Same

src/coordinatorapi.cpp
```diff
@@ -445,18 +447,21 @@ startNewCoordinator(CoordinatorMode mode)
 
   // Create a socket and bind it to an unused port.
   errno = 0;
+  //JNOTE("socket fd value")(jalib::JSocket::sockfd());
```
karya0: Delete?

src/coordinatorapi.cpp
```diff
@@ -445,18 +447,21 @@ startNewCoordinator(CoordinatorMode mode)
...
-    (coordinatorListenerSocket.port()) (JASSERT_ERRNO) (host) (port)
-    .Text("Failed to create socket to coordinator port."
-          "\nIf msg is \"Address already in use\","
-             " this may be an old coordinator."
-          "\nEither try again a few seconds or a minute later,"
-          "\nOr kill other coordinators on this host and port:"
-          "\n    dmtcp_command ---coord-host XXX --coord-port XXX"
-          "\nOr specify --join-coordinator if joining existing computation.");
-  // Now dup the sockfd to
+		(coordinatorListenerSocket.port()) (JASSERT_ERRNO) (host) (port)
```
karya0: Is there any real change or just formatting?

src/coordinatorapi.cpp
```diff
@@ -516,6 +521,7 @@ createNewConnToCoord(CoordinatorMode mode)
     JASSERT(false).Text("Not Reached");
   }
 
+
```
karya0: Kill the newline

src/dmtcp_coordinator.cpp
```diff
@@ -1629,6 +1629,7 @@ main(int argc, char **argv)
       "\n    Exit on last client: " << exitOnLast << "\n";
   }
 #else // if 0
+
```
karya0: ditto

src/dmtcp_coordinator.cpp
```diff
@@ -1645,6 +1646,7 @@ main(int argc, char **argv)
 #endif // if 0
 
   if (daemon) {
+
```
karya0: ditto

src/dmtcp_launch.cpp
```diff
@@ -272,7 +272,11 @@ processArgs(int *orig_argc,
       shift;
     }
 #endif // ifdef HBICT_DELTACOMP
-    else if (s == "--new-coordinator") {
+    else if (s == "--file-socket-path") {
```
karya0: Instead of asking the user for the file socket path, can't DMTCP generate one automatically?

src/util_misc.cpp
```diff
@@ -259,24 +260,27 @@ Util::createDirectoryTree(const string &path)
...
-      } else {
-        return rc;
-      }
-    } else if (rc == 0) {
-      break;
-    } else { // else rc > 0
-      num_written += rc;
-    }
-  } while (num_written < count);
+	do {
```
karya0: Is this intentional?

src/util_misc.cpp
```diff
@@ -259,24 +260,27 @@ Util::createDirectoryTree(const string &path)
 ssize_t
 Util::writeAll(int fd, const void *buf, size_t count)
 {
+
```
karya0: Remove newline

karya0 (changes requested):
It will be really helpful to add a top-level description of what's going on in this PR. How the file socket path is used, and so on. Without that, it's hard to review the core part of your patch.

In the meanwhile, I have left some formatting-related comments for you to address. Please follow the formatting guidelines. Did the git pre-commit hook not work for you? It should have emitted a ton of formatting errors.

jalib/jsocket.cpp
pgrosu: We are validating to see if we are using a INET or Unix socket, since each have different function signatures.

jalib/jsocket.cpp
pgrosu: Same reply as above :)

pgrosu:
Hey Kapil (@karya0) regarding your question on the pre-commit formatting-check.  So initially I did this on Windows and Rohan mentioned I should do it on Linux so that the style-check kicks in.  So I did this on Dekaksi and never got an error, which is why I mentioned that I used the Wiki page - indicating there could be an issue with how the pre-commit hook gets triggered.  Thank you for the comments, and will answer the rest later today since I have a presentation to put together quickly for the Formal Methods class, which has a 48-hour due date.

jiajuncao:
This change seems to make Unix domain socket a special case by setting an env var. If we want TCP sockets and Unix sockets to coexist, then we might have to take care of the env var each time we create a new jalib socket, which is quite painful, in my opinion. I wonder if it is possible to extend (maybe inherit?) the current jsocket API, given the fact that the only difference between the two is the type and addr format. I haven't looked much into the code, but one possible hack would be to add an additional member, which is a hint to the constructor, default to false, telling if it is a Unix domain socket. This may not be (most likely not!) the best solution, but I think it fits more naturally to the design, as the type of the socket is an attribute of the specific socket instance.

src/dmtcp_launch.cpp
pgrosu: Users might not have permission if we chose a default location.  This option provides flexibility if we scale DMTCP and want these locations to be customized.  Where would you like to store it and call it?  Now I usually test it via `/tmp/dmtcp_network.socket`

pgrosu:
So I documented my progress in the following document (see PDF link below) which led me to reach the current implementation.  The methodology I followed was to perform dataflow analysis to understand the implementation, and then determine what were the minimal necessary-and-sufficient invariants that would keep as much of the code untouched:

[Checkpointing_The_Initialization_Procedure.pdf](https://www.dropbox.com/s/6fl0xqeo31nnj58/Checkpointing_The_Initialization_Procedure.pdf?dl=1)

I'm not opposed to exploring ideas but I prefer we spec any ideas, analyze their implications and select the one for optimal integration.  If that process is too long and one satisfies an immediate need in the meantime, then we can optimize to a better one later while working with something that is operational.

What do you think?

Just my $0.02,
~p

rohgarg:
> So I did this on Dekaksi and never got an error, ...

Did you use it for creating a local commit or for pushing an existing commit upstream to your fork? The pre-commit hooks trigger only when you are doing a local commit. I tried applying this patch and doing a local commit on Dekaksi, and it does trigger the pre-commit hooks. Here's what the output looks like for me:

```bash

$ git add jalib/ src/
$ git commit
jalib/jsocket.cpp:238: space before tab in indent.
+  	return ::connect(_sockfd, (const struct sockaddr *) &FILE_SOCKET_ADDR, sizeof(struct sockaddr_un)) == 0;
jalib/jsocket.cpp:256: space before tab in indent.
+  	return ::bind(_sockfd, (const struct sockaddr *) &FILE_SOCKET_ADDR, sizeof(struct sockaddr_un));
...
```

So perhaps it has to do with some configuration settings in your account?

pgrosu:
Hi Rohan,

Yes of course a local commit first, as the wiki instructs.  Here's  a list from my Dekaksi history as to the exact steps and never got it to trigger the pre-commit hooks:

```git clone https://github.com/pgrosu/dmtcp
git pull --rebase origin master
cd dmtcp/
git pull --rebase origin master
git push origin master
git branch
git checkout -b AddUnixFileSocket
cd jalib/
cp ../../../unzipped/dmtcp-file-socket/jalib/* .
cd ..
cd src/
ls
cp ../../../unzipped/dmtcp-file-socket/src/*cpp .
cp ../../../unzipped/dmtcp-file-socket/src/*c .
cp ../../../unzipped/dmtcp-file-socket/src/*h .
cd ..
git add .
git status
git remote add dmtcp-pgrosu https://github.com/pgrosu/dmtcp
git commit
git push dmtcp-pgrosu AddUnixFileSocket
```

`p

PR #688, opened 2018-04-24
Fix checkpoint-restore of file-backed memory regions

rohgarg:
The issue is related to the fact that an application is allowed to
mmap a large region of memory even though the entire region may not be
backed by a file. For example, for a file that is just 1 MB on disk,
the application can make an mmap call asking for 2 GB, but only the
first 1 MB of the memory region will be actually made available. There
are two different behaviors of the Linux kernel when one tries to read
anything beyond the first 1 MB:

- If one tries to read any byte in user-land (load instruction) beyond the first
  1 MB, it generates a SIGBUS.

- If one makes a system call such as write() that uses the entire memory
  region (2 GB), the system call returns success, meaning that it wrote
  2 GB, but it actually only writes out 1 MB!

Now, this confuses the restart code: since we recorded that the size
of the region is 2 GB, the code expects that the memory region would be
2 GB, but the kernel only wrote 1 MB. The code then tries to read 2 GB from
the image. There are two possible outcomes:

 - Either the read fails (because of EOF), or

 - The read succeeds (if the image had sufficient data). However, for the
   subsequent memory region, we'll end up seeing garbage data.

Basically, the checkpoint image has all the checkpoint data but it's
not at the offsets that we are expecting.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -517,23 +517,21 @@ static void mtcp_simulateread(int fd, MtcpHeader *mtcpHdr)
     if (area.size == -1) break;
     if ((area.properties & DMTCP_ZERO_PAGE) == 0 &&
         (area.properties & DMTCP_SKIP_WRITING_TEXT_SEGMENTS) == 0) {
-      void *addr = mtcp_sys_mmap(0, area.size, PROT_WRITE | PROT_READ,
-                                 MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
-      if (addr == MAP_FAILED) {
-        MTCP_PRINTF("***Error: mmap failed; errno: %d\n", mtcp_sys_errno);
-        mtcp_abort();
+
+      off_t seekLen = area.size;
+      if (!(area.flags & MAP_ANONYMOUS) && area.mmapFileSize > 0) {
+        seekLen =  area.mmapFileSize;
       }
-      mtcp_readfile(fd, addr, area.size);
-      if (mtcp_sys_munmap(addr, area.size) == -1) {
-        MTCP_PRINTF("***Error: munmap failed; errno: %d\n", mtcp_sys_errno);
-        mtcp_abort();
+      if (mtcp_sys_lseek(fd, seekLen, SEEK_CUR) < 0) {
+         mtcp_printf("Could not seek!\n");
+         break;
       }
     }
 
     mtcp_printf("%p-%p %c%c%c%c "
                // "%x %u:%u %u"
                 "          %s\n",
-                area.addr, area.addr + area.size,
+                area.addr, area.endAddr,
                 ( area.prot & PROT_READ  ? 'r' : '-' ),
                 ( area.prot & PROT_WRITE ? 'w' : '-' ),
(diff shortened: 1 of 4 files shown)


rohgarg:
The fix here has three parts:

 - An extra field, `mmapFileSize`, in the `ProcMapsArea` struct.

 - Code that detects the size of the backing file at checkpoint time and populates the new field. It also only writes out the portions of memory that are actually backed by the file.

 - Code that detects that a region was in fact a file-backed, shared-memory region on restart, reads the new field, and reads in only the relevant portions from the checkpoint image.

jiajuncao:
LG. It turns out this behavior is in POSIX:
http://pubs.opengroup.org/onlinepubs/009695399/functions/mmap.html
```The mmap() function can be used to map a region of memory that is larger than the current size of the object. Memory access within the mapping but beyond the current end of the underlying objects may result in SIGBUS signals being sent to the process. The reason for this is that the size of the object can be manipulated by other processes and can change at any moment. The implementation should tell the application that a memory reference is outside the object where this can be detected; otherwise, written data may be lost and read data may not reflect actual data in the object.```

File: src/procselfmaps.cpp [hunk 1]
```diff
@@ -235,6 +235,7 @@ int ProcSelfMaps::getNextArea(ProcMapsArea* area)
     area -> flags |= MAP_ANONYMOUS;
   }
 
+  area->mmapFileSize = 0;
```
gc00: Let's initialize `area->mmapFileSize` to `-1`.  The intention of the code is then clearer.  A size of 0 could be understood as somehow mmap'ing a file of file size 0, which is not the intention.

src/mtcp/mtcp_restart.c
```diff
@@ -524,23 +524,20 @@ static void mtcp_simulateread(int fd, MtcpHeader *mtcpHdr)
...
     if ((area.properties & DMTCP_ZERO_PAGE) == 0 &&
         (area.properties & DMTCP_SKIP_WRITING_TEXT_SEGMENTS) == 0) {
-      void *addr = mtcp_sys_mmap(0, area.size, PROT_WRITE | PROT_READ,
-                                 MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
-      if (addr == MAP_FAILED) {
-        MTCP_PRINTF("***Error: mmap failed; errno: %d\n", mtcp_sys_errno);
-        mtcp_abort();
+
+      if (area.flags & MAP_SHARED) {
+        area.size =  area.mmapFileSize;
```
gc00: Suppose `area.mmapFileSize == 0` (in your version of the code) or e `area.mmapFileSize == -1` (with my proposed change).  This is the case where we issued:..
`JWARNING(false)(JASSERT_ERRNO)(area->name).Text("Error getting file info");`

We should probably just use the area.size in the ckpt image file and not try to overwrite it.

rohgarg:
@gc00, @karya0, @jiajuncao: I have now updated the patch to make it more generic. Turns out that this issue affects not just shared-memory regions, but any kind of non-anonymous, file-backed memory regions.

rohgarg:
Ping!

gc00:
I'm not sure why the call to autotest.py itself is failing.  I can also try to look into this.

PR #693, opened 2018-05-28
WSL fix: now mostly works if not using 'make check'

gc00:
This should be easy to review.  he PR has two tiny commits to fix bugs.

Now, `bin/dmtcp_launch -i 5 test/dmtcp1` works, as well as several other random tests by hand.  But for some reason, `make check` is failing.  When running `python autotest.py -v dmtcp1`, it seems to show some problems with `fcntl` and `F_SETOWN`.  But we can leave for another day digging into this remaining mystery.

### DIFF
--- a/configure
+++ b/configure
@@ -5803,7 +5803,9 @@ else
 fi
 if test "$is_wsl" == "yes"; then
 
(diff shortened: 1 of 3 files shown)


pgrosu:
Many Thanks!
Paul

PR #696, opened 2018-06-20
Fix GCC8 compiler warnings

rohgarg:


src/util_exec.cpp
```diff
@@ -576,10 +576,10 @@ Util::runMtcpRestore(int is32bitElf,
   size_t argvSizeDiff = originalArgvEnvSize - newArgvEnvSize;
   if (argvSizeDiff > 0) {
     dummyEnviron = (char *)malloc(argvSizeDiff);
-    memset(dummyEnviron, '0', argvSizeDiff - 1);
+    memset(dummyEnviron, 0, argvSizeDiff - 1);
```
karya0: Here we do want the character `'0'` instead of nul byte. The reason is that we are trying to create a big env var that will be of the form `DUMMY_ENV=00000000000` so that it takes up `argvSizeDiff-1` bytes in length..

src/util_exec.cpp
```diff
@@ -579,7 +579,7 @@ Util::runMtcpRestore(int is32bitElf,
     memset(dummyEnviron, '0', argvSizeDiff - 1);
     strncpy(dummyEnviron,
             ENV_VAR_DMTCP_DUMMY "=",
-            strlen(ENV_VAR_DMTCP_DUMMY "="));
+            argvSizeDiff);
```
karya0: We just want to copy the `"DMTCP_DUMMY_ENV="` string to the beginning of `dummyEnviron`. That way, the buffer `dummyEnviron` will contain `"DMTCP_DUMMY_ENV=00000...0"`

src/mtcp/restore_libc.c
```diff
@@ -268,10 +268,11 @@ TLSInfo_GetTidOffset(void)
 
     tid_offset = tmp - (char *)pthread_desc;
     if (tid_offset != STATIC_TLS_TID_OFFSET()) {
+      // NOTE: PRINTF can only handle strings upto 256 chars
       PRINTF("WARNING: tid_offset (%d) different from expected.\n"
              "  It is possible that DMTCP was compiled with a different\n"
              "  glibc version than the one it's dynamically linking to.\n"
-             "  Continuing anyway.  If this fails, please try again.",
+             "  If this fails, please try again.",
```
gc00: But in this change, you lose the string "continuing anyway.  ".  A better solution is to split the `PRINTF()` into to separate `PRINTF()` commands.  Each of them can be passed a string of less than 256 characters.

src/dmtcpmessagetypes.cpp
```diff
@@ -44,7 +44,8 @@ DmtcpMessage::DmtcpMessage(DmtcpMessageType t /*= DMT_NULL*/)
 {
   // struct sockaddr_storage _addr;
   // socklen_t _addrlen;
-  memset(&compGroup, 0, sizeof(compGroup));
+  // compGroup is a non-trival struct object. It cannot be memset to 0.
```
gc00: `non-trival` -> `non-trivial`

src/util_exec.cpp
gc00: I agree with @karya0.  We need a better fix here.

gc00 (approved):
LGTM.

karya0:
Closing as stale. Please reopen if still relevant.

PR #703, opened 2018-09-12
Fix for aarch64, no delay loop on restart

gc00:
Should be easy to review.  The essential changes are all in the file `src/mtcp/mtcp_restart.c`.  I removed the delay loop for __aarch64__, and added different code.  The issue is that when we mmap and copy new code into it, this is _self-modifying code_.  On the ARM and ARM64 CPUs, that operation requires the program to invalidate the previous cache conctents.  Otherwise, one waits for the cache to catch up, or one observes an ILLEGAL INSTRUCTION error.

Beyond this, one needs a data and instruction barrier, so that all instructions complete and no prefetches are attempted when switching to the new memory region.

Linux/gcc/glibc don't really provide a clean way to invalidate the cache.  So, I added somewhat hackish code with clear comments about what the issues are.

### DIFF
--- a/src/mtcp/mtcp_sys.h
+++ b/src/mtcp/mtcp_sys.h
@@ -326,6 +326,7 @@ struct linux_dirent {
                               5,                                          \
                               args)
 # define mtcp_sys_munmap(args ...)    mtcp_inline_syscall(munmap, 2, args)
+# define mtcp_sys_msync(args ...)    mtcp_inline_syscall(msync, 3, args)
 # define mtcp_sys_mprotect(args ...)  mtcp_inline_syscall(mprotect, 3, args)
 # define mtcp_sys_nanosleep(args ...) mtcp_inline_syscall(nanosleep, 2, args)
 # define mtcp_sys_brk(args ...)                                            \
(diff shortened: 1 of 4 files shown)


src/mtcp/mtcp_sys.h
```diff
@@ -326,6 +326,7 @@ struct linux_dirent {
                               5,                                          \
                               args)
 # define mtcp_sys_munmap(args ...)    mtcp_inline_syscall(munmap, 2, args)
+# define mtcp_sys_msync(args ...)    mtcp_inline_syscall(munmap, 3, args)
```
JainTwinkle: Isn't it  the ```mtcp_inline_syscall(munmap, 3, args) ``` should be ```mtcp_inline_syscall(msync, 3, args) ``` ?

src/mtcp/mtcp_sys.h
```diff
@@ -483,6 +484,13 @@ static unsigned int myinfo_gs;
   INLINE_SYSCALL_RAW(__ARM_NR_set_tls, 1, args)
 # endif // if defined(__arm__)
 
+# ifdef __arm__
+// https://elixir.bootlin.com/linux/v3.1/source/arch/arm/include/asm/unistd.h#L411
+#  define __ARM_NR_cacheflush 0x0f0002
+#  define mtcp_sys_kernel_cacheflush(args ...) \
+  INLINE_SYSCALL_RAW(__ARM_NR_cacheflush, 2, args)
```
JainTwinkle: Could you check if the number of argument is correct? (see [man cacheflush](http://man7.org/linux/man-pages/man2/cacheflush.2.html)

src/mtcp/mtcp_sys.h
gc00: Fixed in my next update to this PR.  Thanks.

PR #705, opened 2018-09-15
Fix epoll1 test (initialize addrlen for accept())

gc00:
Trivial to reviw (one-line change)

### DIFF
--- a/test/epoll1.c
+++ b/test/epoll1.c
@@ -54,7 +54,7 @@ void server()
       exit(EXIT_FAILURE);
     }
 
-    socklen_t addrlen;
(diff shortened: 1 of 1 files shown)


test/epoll1.c
```diff
@@ -54,7 +54,7 @@ void server()
       exit(EXIT_FAILURE);
     }
 
-    socklen_t addrlen;
+    socklen_t addrlen = sizeof(struct sockaddr);
```
karya0: Ideally, it should be `sizeof(sockaddr)` since `struct sockaddr` and `struct sockaddr_in` could be of different sizes but I think it's fine here for this test.

PR #706, opened 2018-09-17
Add:  dmtcp_restart --debug-restart-pause ...

gc00:
There are two unrelated commits here, but it's simpler than separating out an extra, trivial PR, and later re-merging it.

COMMIT 1:  Trivial: 
  Change some text for user that clarifies about conflict w/ old coord; Also, fix a buggy comment.

COMMIT 2:  Add a new flag, --debug-restart-pause, for dmtcp_restart.
  The user can now choose any of four locations to stop and debug very easily.
  Instead of setting an environment variable, creating a new ckpt image, etc., one can directly
    debug on restart from an old ckpt image.  Setting --debug-restart-pause to 1, 2, 3, or 4
    specifies where the code stops with `int dummy = 1; while (dummy);`.
    One then does a gdb attach, and `(gdb) print dummy = 0`
    So, one can quickly get to a chosen location and use the usual debugging trick to continue in GDB.

### DIFF
--- a/src/threadlist.cpp
+++ b/src/threadlist.cpp
@@ -85,6 +85,8 @@ static void Thread_RestoreSigState(Thread *th);
 // Without this, libdmtcp.so will depend on libdmtcp_plugin.so being loaded
 static pid_t _real_getpid(void)
 {
+  JWARNING("_real_getpid")
+          .Text("FIXME: _real_getpid returning virtual pid, not real pid.");
   // libc caches pid of the process and hence after restart, libc:getpid()
   // returns the pre-ckpt value.
   return (pid_t)_real_syscall(SYS_getpid);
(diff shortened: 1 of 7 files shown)


src/threadlist.cpp
```diff
@@ -854,10 +859,8 @@ restarthread(void *threadv)
 #ifdef HAS_PR_SET_PTRACER
       prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY, 0, 0, 0); // For: gdb attach
 #endif // ifdef HAS_PR_SET_PTRACER
-      struct timespec delay = { 15, 0 }; /* 15 seconds */
-      printf("Pausing 15 seconds. Do:  gdb <PROGNAME> %d\n",
-             dmtcp_virtual_to_real_pid(getpid()));
-      nanosleep(&delay, NULL);
+      int dummy = 1;
+      while (dummy);
```
JainTwinkle: An informative print here too. It's good to inform the user that we are waiting for gdb to be attached.

PR #711, opened 2018-10-16
Fix: Coordinator/Host IP is incorrectly calculated

ankitcse07:
Applications which launch processes on remote machine won't be able to run under DMTCP due to
incorrect evaluation of host IP as well as coordinator IP. Problem is localhostIPAddr structure
is empty and code breaks the loop without doing a memcpy.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -1309,6 +1309,7 @@ calcLocalAddr()
         JASSERT(sizeof localhostIPAddr == sizeof s->sin_addr);
         if ( strncmp( name, hostname, sizeof hostname ) == 0 ) {
           success = true;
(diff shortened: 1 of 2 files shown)


ankitcse07:
I incorrectly closed the request

ankitcse07:
Would you be merging this into other branches or this is required to be done by me

rohgarg:
@ankitcse07:
1. I'm not sure what's the right strategy for merging to 2.5.2. There's no 2.5.2 "branch"; it was a release. We can also push this to the 2.5 branch, if that's what you want, but this patch won't be a part of the 2.5.2 release.
2. It's a small enough change. I can port this and merge it into the other branches.

ankitcse07:
Please push this into 2.5 branch

rohgarg:
Done.

PR #712, opened 2018-10-19
Updated remapping of mtcp_restart and stack.

karya0:
This fixes the problem we've been seeing on modern libc/gcc versions where mtcp_restart has additional readonly-data sections that must be copies to the restore location. 

Instead of doing a memcpy which required invalidating the icache on certain ARM cpus, this patch now uses mmap to map the text/data regions at the restore address. The hope is that  we won't need to patch icache since the kernel/cpu would see the newly mmapped region as regular text.

Another change in PR copies the contents of the original mtcp_restart stack to the "safe" restore location. Next, we calculate the offset between the original stack location and the restore stack location.  This offset is then added to the stack pointer ($rsp) and framepointer ($rbp/$fp) registers. That way, we can continue to use the rbp/rsp registers without having to worry about invalid accesses.
 
I've tested it with gcc 8.2.1 and clang 6.0.1 on Linux 4.18.9 with libc 2.27. We still need to test it on ARM, x86 and other variants.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -93,8 +93,7 @@ typedef struct RestoreInfo {
   int stderr_fd;  /* FIXME:  This is never used. */
 
   // int mtcp_sys_errno;
-  VA text_addr;
-  size_t text_size;
+
   VA saved_brk;
   VA restore_addr;
   VA restore_end;
@@ -113,7 +112,7 @@ typedef struct RestoreInfo {
   // void (*post_restart_debug)();
   // void (*restorememoryareas_fptr)();
   int use_gdb;
-  int text_offset;
+  VA mtcp_restart_text_addr;
   ThreadTLSInfo motherofall_tls_info;
   int tls_pid_offset;
   int tls_tid_offset;
@@ -137,7 +136,7 @@ static void restart_fast_path(void);
 static void restart_slow_path(void);
 static int doAreasOverlap(VA addr1, size_t size1, VA addr2, size_t size2);
 static int hasOverlappingMapping(VA addr, size_t size);
-static void getTextAddr(VA *textAddr, size_t *size);
+static void remapMtcpRestartToReservedArea(RestoreInfo *rinfo);
 static void mtcp_simulateread(int fd, MtcpHeader *mtcpHdr);
 void restore_libc(ThreadTLSInfo *tlsInfo,
                   int tls_pid_offset,
@@ -240,16 +239,11 @@ main(int argc, char *argv[], char **environ)
   rinfo.fd = -1;
   rinfo.mtcp_restart_pause = 0; /* false */
   rinfo.use_gdb = 0;
-  rinfo.text_offset = -1;
   shift;
   while (argc > 0) {
     if (mtcp_strcmp(argv[0], "--use-gdb") == 0) {
       rinfo.use_gdb = 1;
       shift;
-    } else if (mtcp_strcmp(argv[0], "--text-offset") == 0) {
-      rinfo.text_offset = mtcp_strtol(argv[1]);
-      shift; shift;
-
       // Flags for call by dmtcp_restart follow here:
     } else if (mtcp_strcmp(argv[0], "--fd") == 0) {
       rinfo.fd = mtcp_strtol(argv[1]);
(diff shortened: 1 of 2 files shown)


JainTwinkle:
@karya0 There is a compile error for mtcp_restart. (see the travis log if you haven't)

karya0:
@rohgarg @Twinklejain @gc00 : This PR is now ready for review.

JainTwinkle:
Approved modulo `travis` checks.
I did the peer review with Kapil.

gc00:
I tested this PR on aarch64.  It breaks aaarch64 badly.   The first commit causes all tests in autotest.py to fail.  The second commit causes assembly errors in mtcp_restart.c, so that we never get as far as testing.

gc00:
@karya0, @Twinklejain, As far as I can tell, this commit continues to break multiple parts of DMTCP.  The PR was approved on Nov. 2, 2018, and I noted on that same day that it breaks aarch64 support.  Recently, I've been adding to DMTCP to create DMTCP-2.6.0, push into a new Debian package, etc.  But as a result of this PR, I am being forced to drop DMTCP support for:
  armhf, aarch64, i386,(and now for WSL),  etc.  The new Debian package is supporting _only_ amd64.

In the past, we have broken `src/mtcp/mtcp_restart.c` for various other architectures.  But the solution then was always to debug mtcp_restart.c in isolation:
```
make -f Makefile.debug gdb
```

But that debugging of mtcp_restart.c requires the `--text-offset`.  This PR removes `--text-offset` with no explanation on why that was removed.  I donn't understand why it was required to remove `--text-offset` in order to support recent glibc/gcc.

@karya0 , how easy would it be for you to make debugging work again for this?
```
make -f Makefile.debug gdb
```

Thanks.

gc00:
@karya0,
    I figured out how to revert PR #712.  After I reverted it, I tested it on dekaksi with gcc-8.3.0 and glibc-2.27 and Linux-4.15.  `make check` seems to work fine there.
    You had written earlier:
> I've tested it with gcc 8.2.1 and clang 6.0.1 on Linux 4.18.9 with libc 2.27. We still need to test it on ARM, x86 and other variants.

But there seems to be a conflict between what you saw and what I'm seeing.  Given that you never document how to _reproduce_ the problem that you were fixing, I'm proposing a new PR in which I revert this PR #712.  This has the advantages that:
  1.  Things like aarch64, arm, i386, WSL, etc., will start working again.
  2.  A major customer of DMTCP needs to use DMTCP-2.6.0.  But they also require i386 support.  The customer has an urgent situation, and they are currently blocked with respect to DMTCP.

In conclusion, you wrote that "This fixes the problem we've been seeing on modern libc/gcc versions where mtcp_restart has additional readonly-data sections."  But you do not document how to _reproduce_ the problem that you saw.  The problem does not seem to appear when I revert PR #712.  We have an urgent need to again support arm, aarch64, i386, etc.  So, unless you can fix the bug in PR #712, I intend to submit a new PR that reverts this.  My plan is to immediately support the several customers that need other architectures than x86-64/adm64.  We were also forced to drop Debian support for anything except amd64 because of the bug in PR #712.

So, when we can document how to _reproduce_ the problem that you saw for gcc-8/glibc-2.27 (which I am not seeing), we can take a slower pace, and add documentation on reproducing the bug, and then produce a more robust fix for it, and carefully review the new PR.

I appreciate your continuing huge contributions to DMTCP.  It's only because of the urgent situation, that I propose to revert this.  Do you have some thoughts on this?  Thanks.

karya0:
@gc00: I have now added https://github.com/dmtcp/dmtcp/pull/772 that re-enables mtcp_restart debugging. I forgot to remove the `--text-offset` commandline flag from the Makefiles. With this PR, we don't need this flag as the binary now uses mmap instead of memcpy for relocating mtcp_restart regions. This allows the code to emit the correct add-symbol-file address without relying on an external script. I just tried `make -f Makefile.debug gdb` with #772 and I was able to add the symbols correctly.

I also tried compiling with `--enable-m32` and wasn't able to observe the bug. Can you or @pgrosu list the steps to reproduce the 32-bit issue?

karya0:
PS: If we revert this PR, we will stop supporting newer distros. Both @rohgarg and I observed this issue with the extra readonly section with mtcp_restart:

```
00400000-00401000 r--p 00000000 103:06 6887825                           /home/kapil/dmtcp/bin/mtcp_restart
00401000-0040a000 r-xp 00001000 103:06 6887825                           /home/kapil/dmtcp/bin/mtcp_restart
0040a000-0040d000 r--p 0000a000 103:06 6887825                           /home/kapil/dmtcp/bin/mtcp_restart
0040d000-0040f000 rw-p 0000c000 103:06 6887825                           /home/kapil/dmtcp/bin/mtcp_restart
0040f000-00411000 rw-p 00000000 00:00 0                                  [heap]
```
With this layout, we won't be able to restart without #712 unfortunately.

PR #713, opened 2018-11-01
Cleanup autoconf/automake files

karya0:
Mostly a change that updates tooling without touching source code itself.


src/Makefile.am
```diff
@@ -33,100 +33,171 @@ d_bindir = $(top_builddir)/bin
...
-
-__d_bindir__dmtcp_coordinator_LDADD = libdmtcpinternal.a libjalib.a \
-			  libnohijack.a -lpthread -lrt
-__d_bindir__dmtcp_launch_LDADD  = libdmtcpinternal.a libjalib.a \
-			  libnohijack.a -lpthread -lrt -ldl
-__d_bindir__dmtcp_restart_LDADD     = libdmtcpinternal.a libjalib.a \
-			  libnohijack.a -lpthread -lrt -ldl
-__d_bindir__dmtcp_command_LDADD     = libdmtcpinternal.a libjalib.a \
-			  libnohijack.a -lpthread -lrt -ldl
+__d_bindir__dmtcp_restart_SOURCES = dmtcp_restart.cpp
```
JainTwinkle: Is  `util_exec.cpp` required here?

karya0: It's part of libdmtcpinternal.a, which is part of LDADD and hence isn't required here explicitly.

PR #717, opened 2018-11-16
fix trampoline setup code for an edge case

JainTwinkle:
Currently, the code set the page permission to 'rwx' where the
trampoline needs to be set up. This is required as we write a few
bytes to the memory to set the trampoline. However, the code
segfaults if the number of bytes that need to be written reaches the
boundary of the current page and tries to write onto a next read-only page.
This patch would fix the edge case by providing `rwx` permission to
the next page if required.

### DIFF
--- a/include/trampolines.h
+++ b/include/trampolines.h
@@ -77,14 +77,6 @@ dmtcp_setup_trampoline_by_addr(void *addr,
                                void *trampoline_fn,
                                trampoline_info_t *info)
 {
-  /* Trick to get "free" conversion of a long value to the
-     character-array representation of that value. Different sizes of
-     long and endian-ness are handled automatically. */
-  union u {
-    void *val;
-    char bytes[sizeof(void *)];
-  } data;
-
   unsigned long pagesize = sysconf(_SC_PAGESIZE);
   unsigned long pagemask = ~(pagesize - 1);
   void *page_base;
@@ -95,11 +87,11 @@ dmtcp_setup_trampoline_by_addr(void *addr,
   page_base = (void *)((unsigned long)info->addr & pagemask);
   int pagecount = 1;
 
-  /* Increase the pagecount if number of bytes needs to be wriiten
-   * to set a trampoline falls in the next page.
+  /* Increase the pagecount if number of bytes that needs to be written
+   * for the trampoline happens to fall in the next page.
    */
   if (((unsigned long)page_base + pagesize) - (unsigned long)info->addr
-        < sizeof(data.bytes)) {
+        < ASM_JUMP_LEN) {
     pagecount += 1;
   }
   /* Give that whole page RWX permissions. */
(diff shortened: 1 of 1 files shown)


include/trampolines.h
```diff
@@ -85,9 +93,17 @@ dmtcp_setup_trampoline_by_addr(void *addr,
 
   /* Base address of page where func resides. */
   page_base = (void *)((unsigned long)info->addr & pagemask);
-
+  int pagecount = 1;
+
+  /* Increase the pagecount if number of bytes needs to be wriiten
+   * to set a trampoline falls in the next page.
```
gc00: Suggested re-wording:  ... if _the_ number of bytes _that_ needs to be written _for_ the trampoline _happens to fall_ in the next page.

include/trampolines.h
```diff
@@ -85,9 +93,17 @@ dmtcp_setup_trampoline_by_addr(void *addr,
...
+   */
+  if (((unsigned long)page_base + pagesize) - (unsigned long)info->addr
+        < sizeof(data.bytes)) {
```
gc00: Shouldn't this code be:
 `if (((unsigned long)page_base + pagesize) - (unsigned long)info->jump < ASM_JUMP_LEN) {`

(We want to replace ASM_JUMP_LEN bytes and not simply `sizeof(data.bytes)` number of bytes.)

include/trampolines.h
```diff
@@ -77,6 +77,14 @@ dmtcp_setup_trampoline_by_addr(void *addr,
                                void *trampoline_fn,
                                trampoline_info_t *info)
 {
+  /* Trick to get "free" conversion of a long value to the
+     character-array representation of that value. Different sizes of
+     long and endian-ness are handled automatically. */
+  union u {
+    void *val;
+    char bytes[sizeof(void *)];
+  } data;
```
gc00: @JainTwinkle,
    I know that the use of `data.bytes` in `memcpy()` existed even before your code.  But can we clean up that very old code?  It seems to me that:<br>
`memcpy(info->jump + ADDR_OFFSET, &trampoline_fn, sizeof(void *));`<br>
will do exactly the same thing as this older code:
`memcpy(info->jump + ADDR_OFFSET, data.bytes, sizeof(data.bytes));`

If you agree with this modification, then we can finally remove the above lines (`/* Trick to ... union u {...} data;`.
And if you agree with these modifications, then could you create this as a second, separate commit inside this same PR (or a separate PR, if you prefer)?  Thanks.

include/trampolines.h
JainTwinkle: Thanks for pointing out :)

include/trampolines.h
JainTwinkle: Yes, you are right. `ASM_JUMP_LEN` is the right length for this edge case. I will be changing the `if` condition in the next commit.

include/trampolines.h
gc00: Hi @JainTwinkle, Let's discuss the endianness issue face-to-face.  It will be easier for these subtle questions.  I think the revised version will produce the correct endianness, but maybe I'm wrong -- and so it's easier to talk directly.  Anyway, the current PR is good, and that discussion will be about a future PR.

gc00:
Thanks for the changes.  LGTM.  (And concerning a possible future PR on endianness, we can discuss that offline.)

PR #722, opened 2018-12-01
shareddata: Fix a race when multiple clients join simultaneously

rohgarg:
The issue happens when multiple clients try to join a computation at the
same time. One of the clients gets the lock in the kernel to create the
shared data file (using `open(O_EXCL)`). Now, this client (the leader)
is responsible for initializing the shared data header, and the other
clients must wait for the initialization to complete.

The current test for completion of initialization was inadequate; it
relied on testing the size of the shared data file. The race condition
here is that client that gets the lock to create the shared data file
would set the size (using `truncate`) and be preempted before it got a
chance to finish the initialization of the header. Meanwhile, the other
clients would race ahead, check the version string in the shared data
header, and assert out.

The fix here is to let the clients wait until the leader has finished
initializing the shared data.

### DIFF
--- a/src/shareddata.cpp
+++ b/src/shareddata.cpp
@@ -138,7 +138,8 @@ SharedData::initialize(const char *tmpDir = NULL,
     ostringstream o;
     o << tmpDir << "/dmtcpSharedArea."
       << *compId << "." << std::hex << coordInfo->timeStamp;
-    // THIS IS A DUP OF initializeHeader AND OF size, below; Pass this in as an argument to it.
+    // THIS IS A DUP OF initializeHeader AND OF size, below; Pass this in as
+    // an argument to it.
     off_t size = CEIL(SHM_MAX_SIZE, Util::pageSize());
 
     int fd = _real_open(o.str().c_str(), O_RDWR | O_CREAT | O_EXCL, 0600);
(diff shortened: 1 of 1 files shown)


src/shareddata.cpp
```diff
@@ -182,10 +184,17 @@ SharedData::initialize(const char *tmpDir = NULL,
   } else {
     struct stat statbuf;
     while (1) {
+      bool initialized = false;
       Util::lockFile(PROTECTED_SHM_FD);
       JASSERT(fstat(PROTECTED_SHM_FD, &statbuf) != -1) (JASSERT_ERRNO);
+      initialized = sharedDataHeader->initialized;
       Util::unlockFile(PROTECTED_SHM_FD);
-      if (statbuf.st_size > 0) {
+      // If we got here, it implies that some other peer won the race
```
pgrosu: Looks great, but please add a FIXME or TODO for the scenario when the two processes are synchronized and hit the check at the same time, where your recommendation of SysV semaphores/condition variables would be a more robust check.

Thanks,
~p

gc00:
This PR is continued as PR #761. So, I'm closing it here.   (@rohgarg was unavaialble to do the rebase and push in the PR.  I'm doing iit here while continuing to use a signature with him as author.)

PR #726, opened 2019-01-29
sysvipc: Fix a race with two clients creating shm with IPC_PRIVATE

rohgarg:
There was a subtle race condition in our `shmget()` wrapper with two
clients trying to simultaneously create an shm region with `IPC_PRIVATE`
and `IPC_EXCL`.

The code relies on the `VIRTUAL_TO_REAL_SHM_KEY` API, which, in turn,
uses the DMTCP shared data area to share and assign unique keys for SysV
shm regions. The API returns -1 if the key does not exist in the DMTCP
mapping. Now, if one client raced ahead and managed to create the region,
it'd end up populating the DMTCP shared area with a "IPC_PRIVATE ->
realKey" mapping. Later, when the other client would try to create its
own shared memory region, the `VIRTUAL_TO_REAL_SHM_KEY` API would return
the key corresponding the earlier created entry and the `shmget(IPC_EXCL)`
call would fail with an `EEXIST` error.

The fix here is to detect this special case of `IPC_PRIVATE` and use
a processes real pid to create the shared memory region. This also
preserves the semantics of `IPC_PRIVATE`.

### DIFF
--- a/src/plugin/svipc/sysvipcwrappers.cpp
+++ b/src/plugin/svipc/sysvipcwrappers.cpp
@@ -68,7 +68,30 @@ shmget(key_t key, size_t size, int shmflg)
   int virtId = -1;
 
   DMTCP_PLUGIN_DISABLE_CKPT();
-  key_t realKey = VIRTUAL_TO_REAL_SHM_KEY(key);
+  key_t realKey = -1;
+
+  // If multiple clients try to simultaneously create shm regions with
+  // (IPC_PRIVATE | IPC_EXCL), there is a race condition that can cause the
+  // shmget call to fail.
+  //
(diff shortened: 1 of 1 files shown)


gc00:
@rohgarg, Can you add a comment to the code?  Right now, it seems to say that real_key will be set eitehr to the real IPC_PRIVATE key or it will be set to a real pid.

Can you write a few words about why a key used in shmget could actually be a pid (even if it means reviewing something basic about shm and IPC_PRIVATE).  This will make it much easier for others to understand this code.  Thanks.

gc00 (changes requested):
Hi @rohgarg, Could you add a few commetns to help the reader understand why the real_key can be either a real pid or IPC_PRIVATE?  (See my comments in the "conversation" for more on this.)  Thanks.

gc00 (approved):
LGTM.

PR #727, opened 2019-02-11
Initial implementation for pre-suspend barriers.

karya0:
TODO: Add a test plugin with pre-suspend barriers (potentially involving two or more processes; one of the processes should be killed as part of the barrier.


gc00:
Hi Twinkle,
    I think Kapil wrote on the side that he wanted us to test out
this PR ourselves before we approve it.  He didn't have time to
directly test it himself, and there may be minor bugs.
    Could you remove the approval, if Github lets you do that?

Thanks,
- Gene

PR #728, opened 2019-02-15
coordinator-api: Fix caching of coordinator host info

rohgarg:
We use a global pointer to cache the coordinator host string. This can
be problematic if the string object the pointer points to gets destroyed.

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -140,11 +140,19 @@ getCoordHostAndPort(CoordinatorMode mode, string *host, int *port)
     if (*host == "") {
       if (getenv(ENV_VAR_NAME_HOST)) {
         *host = getenv(ENV_VAR_NAME_HOST);
+        _cachedHost = getenv(ENV_VAR_NAME_HOST);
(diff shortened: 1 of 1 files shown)


src/coordinatorapi.cpp
```diff
@@ -140,11 +140,19 @@ getCoordHostAndPort(CoordinatorMode mode, string *host, int *port)
...
+        _cachedHost = getenv("DMTCP_HOST");
       } else {
         *host = DEFAULT_HOST;
+        _cachedHost = DEFAULT_HOST;
       }
+    } else {
+      // The caller's string object needs to be valid across
+      // multiple calls to this function, or else, the _cachedHost
+      // pointer will become a dangling pointer.
+      _cachedHost = host->c_str();
```
karya0: Can we have a char[] and do a strcpy instead?

gc00:
This PR is continued as PR #762. So, I'm closing it here. (@rohgarg was unavaialble to do the rebase and push in the PR. I'm doing iit here while continuing to use a signature with him as author.)

PR #732, opened 2019-04-20
Fix LD_PRELOAD and LD_PRELOAD_M32 env vars

gc00:
Easy review: one-line bug fix

Without this, the env. var. `DMTCP_HIJACK_LIBS` has extra ':' delimiters (mostly harmless), and `DMTCP_HIJACK_LIBS_M32` is missing its ':' delimiters (very harmful for 32-bit applications).

### DIFF
--- a/src/dmtcp_launch.cpp
+++ b/src/dmtcp_launch.cpp
@@ -809,7 +809,7 @@ setLDPreloadLibs(bool is32bitElf)
         preloadLibs += ":";
 #if defined(__x86_64__) || defined(__aarch64__)
         preloadLibs32 += Util::getPath(p->lib, true);
-        preloadLibs += ":";
+        preloadLibs32 += ":";
 #endif // if defined(__x86_64__) || defined(__aarch64__)
       }
     }


pgrosu:
I wanted a few other fixes to be done to this PR before merging.  Can we unmerge?

Thanks,
Paul

karya0:
@pgrosu: You can create a new PR with new changes. You can revert the commit but that's not advisable unless absolutely necessary.

pgrosu:
@karya0 For this time only, but from now on I would like to see a little more than LGTM with PRs here without some verification to confirm it.  The continuous integration would also require updating to capture this scenario.  It's usually best it have things in one PR if they are closely related.

karya0:
I think for a PR this size, LGTM is a reasonable response. It's not a huge change and even if it were, LGTM does say that the reviewer has gone through the code and didn't find anything that should prevent it from getting committed.

Re. closely related changes: this shouldn't matter. If the changes are done by the same person, they can include it in a single PR or separate PRs. Having separate PRs isn't bad as it allows for easier reviewing.

karya0:
Re. CI and verification: I agree that we have a lot of work to do in that aspect to properly support M32 (and other less-frequently used scenarios).

PR #740, opened 2019-07-17
Lowering RLIMIT_NOFILE can cause crash

gc00:
* The commit now checks for a static protected fd base less than 1024,
       and triggers assert during init/exec/restart if the current
       RLIMIT_NOFILE will not leave enough fd's for the protected fd's.
    * The commit adds a wrapper around setrlimit, to refuse to reduce
       RLIMIT_NO below 1024.
    * In the second commit, an autotest was added to check if the user lowered RLIMIT_NOFILE,
      but not below 1024.  This test should pass.

### DIFF
--- a/jalib/jalib.cpp
+++ b/jalib/jalib.cpp
@@ -29,6 +29,8 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <sys/time.h>
+#include <sys/resource.h>
 
 #include "jalib.h"
 #include "jassert.h"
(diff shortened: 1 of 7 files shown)


jalib/jalib.cpp
```diff
@@ -135,9 +137,12 @@ dup(int oldfd)
   REAL_FUNC_PASSTHROUGH(int, dup) (oldfd);
 }
 
-int
-dup2(int oldfd, int newfd)
-{
+int dup2(int oldfd, int newfd) {
+  struct rlimit file_limit;
+  getrlimit(RLIMIT_NOFILE, &file_limit);
+  JASSERT ( (unsigned int)newfd < file_limit.rlim_cur )
```
karya0: This assert shouldn't be here. The caller is allowed to call with an invalid newfd. We should simply return the error code.

For the case, where DMTCP is trying to use a newfd beyond the current limits, please add checks at the call-site.

Also note that this wrapper is specific to jalib. The core and plugins won't end up here when calling dup2.

gc00: Hi @karya0 ,
    I respectfully think the code should stand as is.  I've updated to add comments before the dup2 wrapper to explain my reasoning.
    In addition to those comments, I'll add that jalib should not be calling dup2 with a newfd beyond RLIMIT_NOFILE.  There are currently two dup2 call sites in jalib, and if either one uses a newfd that is too large, then this is a bug in DMTCP (since the call site is internal to DMTCP).  So, we should error out.

pgrosu: I agree with @gc00 here.

PR #743, opened 2019-07-20
Don't initialize DMTCP before libc is initialized (DMTCP-2.6)

gc00:
This is for the 2.6 branch. It fixes a bug (isolated to the 2.x branch). See test/plugin-init.cpp' comments for an analysis of the bug. In essence, a call to std::cout at the INIT event causes dmtcp_initialize()` to be called before libc.so has been initialized. As a result, a call to libc:malloc() fails.  Also, see issue #744 for a fuller discussion, as well as the excerpted comment below in this message.

This bug report was contributed by:

    Johannes Stoelp
    Laurent Buchard
    Pankaj Mehta

from: Synopsys, Inc.

====
Note especially the comment in `src/dmtcpworker.cpp`:

```
+  // FIXME:
+  // PR #742: DMTCP malloc wrapper called prematurely, causing DMTCP
+  //    to initialize prematuruely.  'dmtcpInMalloc' test fixes that.
+  //    But apparently the 'emacs' test causes a different DMTCP
+  //    wrapper to be called prematurely and initialize DMTCP prematurely.
+  //    We should discover all such cases of premature DMTCP initialization,
+  //    and guard against them.  It seems that in github Travis
+  //    in July, 2019, the emacs test fails when
+  //    '(! inDmtcpWorker && dmtcpInMalloc)' is tested for.
+  //    Presumably, there is another function being called in the emacs test
+  //    besides DMTCP's malloc(), and CentOS 7.5 reproduces this failure
+  //    for 'emacs -nw' (emacs-nox not found).
+  //    Similarly, the second restart on vim is failing.
+  //  It seems to be a corner case that can be fixed in the next release.
+  //    For now, both the 'vim' and 'emacs -nw' tests are commented out
+  //    in test/autotest.py
```

### DIFF
--- a/src/dmtcpworker.cpp
+++ b/src/dmtcpworker.cpp
@@ -251,12 +251,31 @@ static void installSegFaultHandler()
   JASSERT (sigaction(SIGSEGV, &act, NULL) == 0) (JASSERT_ERRNO);
 }
 
-//called before user main()
-//workerhijack.cpp initializes a static variable theInstance to DmtcpWorker obj
+static bool inDmtcpWorker = false;
+// A weak symbol will have default value 0 (same as false)
+extern "C" int dmtcpInMalloc __attribute__((weak));
+
+//called before user main() to initialize DMTCP
 extern "C" void dmtcp_initialize()
 {
   static bool initialized = false;
-  if (initialized) {
+  // FIXME:
+  // PR #742: DMTCP malloc wrapper called prematurely, causing DMTCP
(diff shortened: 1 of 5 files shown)


rohgarg:
While I more or less okay with this, please also see my comments in issue #744. I believe there's a need for some refactoring/clean-up of the initialization logic. This PR is only a partial and slightly ugly fix (because of the use of inter-library global state) to patch over a fundamental issue.

gc00:
This has been pushed into the 2.6.0 branch earlier.  I'm closing the PR.

PR #746, opened 2019-07-20
Restore floating point env and RLIMIT_NOFILE/STACK

gc00:
At checkpoint time this saves the floating point env with `fegetenv()` (e.g., the rounding mode) and the `RLIMIT_NOFILE` and `RLIMIT_STACK`.  At restart time, it restores these values.

The second commit creates a test, `test/rlimit-restore`.  The test failed before applying this PR (as expected), and it now succeeds when this PR is used.

The main code for the 3.0/master version was placed in plugininfo.cpp.  In principle, it seemed like it should have gone into src/processinfo.cpp (similarly to the 2.6 version), but it wasn't clear where.  The refresh() method is called both at checkpoint time and pre-exec time.  We want to save the values only at checkpoint time.  At pre-exec time, the values should revert back to their default.  And I was confused about the name `ProcessInfo::refresh()`, since we were saving the current state, and not refreshing (restoring) some state.  Maybe instead of putting the code into `refresh()`, I could add it to `checkpoint()`.  At least the `restart()` method seemed like the obvious place to restore the values.

Should I move the code from `plugininfo.cpp` back to `processinfo.cpp`?  If so, can we rename `refresh()` to something meaningful?  And should I just add my code to `checkpoint()`?

Anyway, the 2.6 version of this is non-controversial and works nicely.  I'll push that in, for the DMTCP 2.6.0 release.

### DIFF
--- a/src/writeckpt.cpp
+++ b/src/writeckpt.cpp
@@ -446,11 +446,40 @@ writememoryarea(int fd, Area *area, int stack_was_seen)
     /* Kernel won't let us munmap this.  But we don't need to restore it. */
     JTRACE("skipping over [stack] segment (not the orig stack)")
       (addr) (area->size);
-  } else if (0 == strcmp(area->name, "[vsyscall]") ||
-             0 == strcmp(area->name, "[vectors]") ||
-             0 == strcmp(area->name, "[vvar]") ||
-             0 == strcmp(area->name, "[vdso]")) {
+  } else if (0 == strcmp(area -> name, "[vsyscall]") ||
+             0 == strcmp(area -> name, "[vectors]") ||
+             0 == strcmp(area -> name, "[vvar]"))
+             // NOTE: We can't trust kernel's "[vdso]" label here.  See below.
+  {
     JTRACE("skipping over memory special section")
+      (area->name) (addr) (area->size);
+  } else if ( area->__addr == ProcessInfo::instance().vdsoStart() ) {
+    //vDSO issue:
+    //    As always, we never want to save the vdso section.  We will use
+    //  the vdso section code provided by the kernel on restart.  Further,
+    //  the user code on restart has already been initialized and so it
+    //  will continue to use the original vdso section determined during
+    //  program launch.  Luckily, during the DMTCP_INIT event, DMTCP recorded
+    //  this vdso address when it called ProcessInfo::instance().init().
+    //    Now, here's the bad news.  During the first restart, the kernel
+    //  may choose to locate the vdso at a new address.  So, in
+    //  src/mtcp/mtcp_restart, DMTCP will mremap the kernel's vdso back
+    //  to the original address known during program launch.  This is as
+    //  it should be.  But when DMTCP does an mremap of vdso, the kernel
+    //  fails to update its own "[vds0]" label.  So, during the second
+    //  checkpoint (after the first restart), we can't trust the "[vdso]"
+    //  label to tell us where the vdso section really is.  And it's even
+    //  worse.  During mtcp_restart, we may have done the mremap, and there
+    //  may even now be some user data that was restored to the address
+    //  where the kernel thinks the "[vdso]" label belongs.  So, we would
+    //  be saving the original vdso section (which is wrong), and we would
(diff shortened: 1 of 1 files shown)


jiajuncao:
I agree with Kapil. Also, may I suggest from now on we should include more detailed change description in the commit message? Although pull requests already have the story, people rely on commit logs in many cases.

gc00:
Thanks for the design modification that I asked for, @karya0 .  I'll follow it.

Concerning the mention of a new plugin, I don't see any new plugin in this PR.  There's a different PR with a plugin used solely in a new test to test the fix in the PR, and I'd be happy to discuss it in that PR.

@jiajuncao suggested that we copy some of the explanation in the PR over to the commits themselves.  If the team wishes to adopt that style, I'm happy to do so.  Are there other opinions on this?

karya0:
> Concerning the mention of a new plugin, I don't see any new plugin in this PR. There's a different PR with a plugin used solely in a new test to test the fix in the PR, and I'd be happy to discuss it in that PR.

This PR adds support for checkpoint/restore of rlimits (right now only for fdlimit, but we can easily imagine other things in future) and so it's essentially a plugin for all intents and purposes. It will be an internal plugin and enabled by default, just like our terminal plugin. That way, we can leverage the existing plugin mechanisms and keep everything local in a single place. In the next iteration, I can imagine bringing in the corresponding wrappers into this file, making it a complete standalone thing.

So for now, if we just create a rlimit.cpp, and follow the model from terminal.cpp, we should be fine. Please let me know if I it doesn't make sense or if I can help with any of that.

gc00:
@karya0, I made the code into an internal plugin now.  I changed the title of the PR to show that it's now a plugin.  I'm including the second commit, that includes the rlimit-restore autotest associated with this new functionality.
It's ready for re-review now.

I note that right now, the 'java1' test seems to be failing under Travis.  No idea why.  The tst works fine locally.  If Travis continues to fail the 'java1' test, I can look into it later, in parallel with you reviewing the new plugin.

src/rlimitfloatenv.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+
+static int roundingMode = -1;
+static fenv_t envp;
+static rlim_t rlim_cur_nofile = 0;
+static rlim_t rlim_cur_stack = 0;
+
+static void
+save_rlimit_float_settings()
+{ roundingMode = fegetround();
```
karya0: Newline after opening brace.

I wonder why wasn't it caught by the pre-commit hook. Perhaps the rules need to be updated to check for opening brace.

src/rlimitfloatenv.cpp
```diff
@@ -0,0 +1,82 @@
...
+  fegetenv(&envp);
+  { struct rlimit rlim = {0, 0};
```
karya0: Not sure if the braces here are intentional. If they are, please insert a newline after the brace.

src/rlimitfloatenv.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+static void
+restore_rlimit_float_settings()
+{ fesetenv(&envp);
+  fesetround(roundingMode);
+  { struct rlimit rlim = {0, 0};
+    getrlimit(RLIMIT_NOFILE, &rlim);
+    rlim.rlim_cur = rlim_cur_nofile;
+    setrlimit(RLIMIT_NOFILE, &rlim);
```
karya0: Should we be checking the return code for failure? Another check would be to see if the current hard limit is lower than rlim_cur_nofile. Similarly for stack.

gc00: I'll check and fix the style issues.  And I'll also see if Travis continues to give a problem with the 'java1' test.

By the way, I just now pushed in one more change.  I'm renaming `ProcessInfo::refresh()` to `ProcessInfo::getState()`.  That seems to be more meaningful, given the several calls to `getXXX()` in the implementation of the method.

gc00:
The stupid github closed the PR, even though I didn't tell it to.  Rather than re-open this, the continuation of this PR is now in the opern PR #751.  The commits to be pushed in are also located in PR #751.

PR #747, opened 2019-07-21
Presuspend code for coordinator

karya0:
Somehow, this commit got excluded from the original PR (#727).


gc00:
@karya0, Ahh!  I wondered why the code seemed so incomplete, and there were so many bugs/missing-features when I was trying to fix the code a few months ago.  :-)  Eventually, I ran out of time, and figured I'd come back to it sometime in the summer when I had more time.

Now that it's the summer, I can take another look at this in August.

karya0:
> @karya0, Ahh! I wondered why the code seemed so incomplete, and there were so many bugs/missing-features when I was trying to fix the code a few months ago. :-) 

Yeah, that was quite dumb on my end. I think I did something like "git ci -m ..." but forgot to do a git add first. And then just did a push which pushed the other commits but these files remained in my working tree.

PR #748, opened 2019-07-21
Add support for POSIX condition variables

gc00:
@karya0 , Some people from Synopsys have generously contributed this enhancement that allows DMTCP to handle condition variables.  The patch that they submitted is for the 2.6.0 branch, and it works great there.  In this PR, I take essentially the same code and port it to the master branch.

This PR for the master branch still has a bug in it specific to the master branch.  I've diagnosed the bug completely here.  But fixing it requires an adjustment to the design of DMTCP-3.0.  Since this was originally your design, I assume that you would prefer to adjust the design and propose the right bug fix (rather than me messing up your code with my own design).  :-)

In this PR, there are two commits.  The first commit added 'follow_mutex()' to the 'pthread_mutex()' wrapper, and it seems to work fine.  To see it in operation, see PR #745, which contains only that first commit.

In the second commit, during initialization (when only the ckpt thread exists), DMTCP will recursively call `pthread_mutex_lock()`, and it recursively acquires the lock:  `ScopedMutex_mapMutexVirtTid lock;` within `follow_mutex()`.  The new code is essential to supporting condition variables, since we need to track locks, their owners, etc.

I've created a `gdbinit` file so that you can easily see the bug within gdb:
```
# BUG:
# pthread_mutex_lock is called recursively on tbl_lock inside:
### dmtcp::VirtualIdTable<void*>::_do_lock_tbl
#
# In follow_mutex(), it constructs a lock:
#   ScopedMutex_mapMutexVirtTid lock;
#     and the constructor calls pthread_mutex_lock on: mutex_mapMutexVirtTid
# See the end of this file for comments on how the bug appears.

b execvp
run
set break pending on
b dmtcp_initialize
cont
disable 2
b clock_gettime
cont
b src/plugin/pid/pid_mutexwrappers.cpp:pthread_mutex_lock
cont
# The new commit adds the following new code to pthread_mutex_lock:
#   pthread_mutex_lock->post_lock->follow_mutex->
#      ScopedMutex_mapMutexVirtTid lock; // constructor inside follow_mutex
# The constructor calls:  __real_pthread_mutex_lock(mutex_mapMutexVirtTid)
b '(anonymous namespace)::ScopedMutex_mapMutexVirtTid::ScopedMutex_mapMutexVirtT
id'
cont
b dmtcp_gettid
cont
# Next, dmtcp_gettid() is going to call pthread_mutex_lock() in order to
#   get the virtual tid from the table.  This will create a recursive
#   call to DMTCP pthread_mutex_lock(), which eventually recursively calls:
#     'ScopedMutex_mapMutexVirtTid lock;'
```

Run this as:
`gdb -x gdbinit --args bin/dmtcp_launch test/dmtcp1`
to see the bug.

And here is the acknowledgment, so that I don't forget it in the final PR:

This bug fix was contributed by:

1) Johannes Stoelp
2) Laurent Buchard
3) Pankaj Mehta

from: Synopsys, Inc.

### DIFF
--- a/src/plugin/pid/pid_mutexwrappers.cpp
+++ b/src/plugin/pid/pid_mutexwrappers.cpp
@@ -10,16 +10,17 @@
 
 #include "jassert.h"
 #include "pidwrappers.h"
-
-#define __real_pthread_mutex_lock      NEXT_FNC(pthread_mutex_lock)
-#define __real_pthread_mutex_trylock   NEXT_FNC(pthread_mutex_trylock)
-#define __real_pthread_mutex_timedlock NEXT_FNC(pthread_mutex_timedlock)
+#include "dmtcp.h" // for NEXT_FNC_DEFAULT macro
 
 #define __real_pthread_mutex_lock       NEXT_FNC(pthread_mutex_lock)
 #define __real_pthread_mutex_trylock    NEXT_FNC(pthread_mutex_trylock)
 #define __real_pthread_mutex_timedlock  NEXT_FNC(pthread_mutex_timedlock)
 #define __real_pthread_mutex_unlock     NEXT_FNC(pthread_mutex_unlock)
 #define __real_pthread_mutex_consistent NEXT_FNC(pthread_mutex_consistent)
+
+// For explanation why we need NEXT_FNC see [1]
+#define __real_pthread_cond_wait        NEXT_FNC(pthread_cond_wait)
+#define __real_pthread_cond_timedwait   NEXT_FNC(pthread_cond_timedwait)
 
 // |=-------------------------------=[ GOAL ]=--------------------------------=|
 //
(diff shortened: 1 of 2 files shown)


karya0:
@gc00: Would it help if we created a new dmtcp-specific mutex type with corresponding dmtcp_mutex_lock, dmtcp_mutex_unlock, etc.? This has been on my radar for a while due to the involvement of barries/mutexes in the low-level design. Essentially, I want to go back to Mike's original futex approach, but with proper abstractions this time around. That way we would be self-sufficient and won't have to get mixed with real/virtual tids, etc.

gc00:
@karya0, Yes!  That seems like a great idea to have DMTCP-specific 'dmtcp_mutex_lock/unlock, etc.  It seems similar to our jalloc strategy where DMTCP uses a spearate malloc arena from the normal libc arena.  Would it be easy for you to create that?

PR #751, opened 2019-07-31
Restore floating point env and RLIMIT_NOFILE/STACK (v2)

gc00:
Continuation of PR #746.  Stupid github has Travis using an incompatible 'java' and 'javac' versions for an obscure reason.  It one does a sanity test to see if an earlier version would pass the Travis test, then github decides to close the pull request because it sees that the most recent push has 0 new commits -- even though no one told it to close the PR.

Let's continue the discussion here, that was started in PR #746.

This version of the PR now adds a third commit to fix Travis's incompatible java vs. javac versions in the presence of Python (see below).

### DIFF
--- a/src/rlimitfloatenv.cpp
+++ b/src/rlimitfloatenv.cpp
@@ -1,0 +1,94 @@
+#include <fenv.h>
+#include <sys/time.h>
+#include <sys/resource.h>
+#include "../jalib/jassert.h"
+#include "shareddata.h"
+#include "util.h"
+#include "config.h"
+#include "dmtcp.h"
+
+/*************************************************************************
+ *
+ *  Save and restore rlimit and float exception settings.
+ *
+ *************************************************************************/
+
+namespace dmtcp
+{
+
+static int roundingMode = -1;
+static fenv_t envp;
+static rlim_t rlim_cur_nofile = 0;
+static rlim_t rlim_cur_stack = 0;
+
+static void
+save_rlimit_float_settings()
+{
+  roundingMode = fegetround();
+  fegetenv(&envp);
+
+  struct rlimit rlim = {0, 0};
+  getrlimit(RLIMIT_NOFILE, &rlim);
+  rlim_cur_nofile = rlim.rlim_cur;
+
+  getrlimit(RLIMIT_STACK, &rlim);
+  rlim_cur_stack = rlim.rlim_cur;
+}
+
+static void
+restore_rlimit_float_settings()
+{
+  fesetenv(&envp);
+  fesetround(roundingMode);
+
+  struct rlimit rlim = {0, 0};
+  getrlimit(RLIMIT_NOFILE, &rlim);
+  JWARNING(rlim_cur_nofile <= rlim.rlim_max)
+    (rlim_cur_nofile) (rlim.rlim_max)
+    .Text("Previous soft limit of RLIMIT_NOFILE lowered to new hard limit");
+  rlim.rlim_cur = MIN(rlim_cur_nofile, rlim.rlim_max);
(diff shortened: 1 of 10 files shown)


gc00:
@karya0, I've now converged.  Travis was using one path for javac in test/Makefile (using Javac-11), and a different path inside Python running autotest.py on the 'java1' test, where it uses 'java' (using Java-8).  This failed because the class file format used by javac-v11 is not compatible with running it in java-v8.

The commit "Fix autotest.py: find correct Travis path for Java" now fixes that.  (And I made teh style changes requested earlier.)  This PR should now be ready to be pushed in.

src/rlimitfloatenv.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+
+static int roundingMode = -1;
+static fenv_t envp;
+static rlim_t rlim_cur_nofile = 0;
+static rlim_t rlim_cur_stack = 0;
+
+static void
+save_rlimit_float_settings()
+{ roundingMode = fegetround();
```
karya0: Please insert a newline after the brace.

src/rlimitfloatenv.cpp
```diff
@@ -0,0 +1,92 @@
...
+  getrlimit(RLIMIT_NOFILE, &rlim);
+  rlim_cur_nofile = rlim.rlim_cur;
+
+  getrlimit(RLIMIT_STACK, &rlim);
+  rlim_cur_stack = rlim.rlim_cur;
+}
+
+static void
+restore_rlimit_float_settings()
+{ fesetenv(&envp);
```
karya0: Newline.

test/autotest.py
```diff
@@ -15,6 +15,19 @@
...
+
+# FIX for bad path for Java:  Travis prepended
+#     "/usr/bin:/opt/pyenv/libexec:/opt/pyenv/plugins/python-build/bin:/"
+# to os.environ['PATH'] on July 31, 2019.  It does this, even though
+# "/usr/bin" occurs later in the path.  /usr/bin/java exists as Java-8.
+# So, it never finds java (version Java-11) in
+#   "/usr/local/lib/jvm/openjdk11/bin", which occurs later in path.
+# Unfortunately, 'Makefile' finds javac as verion Javac-11,
+# and so the 'java1' test fails, using java (Java-8).
+if os.environ.has_key('PATH') and os.environ['PATH'].startswith('/usr/bin:') \
```
karya0: Instead of having the PATH updates here, let's have configure detect and export a couple of make variables like JAVAC_PATH and JAVA_PATH that we can use directly in the test suite. That way we won't have any side-effects here.

gc00: @karya0, It won't work to configure for JAVA_PATH and JAVAC_PATH.  We have to test for this inside Python when hitting the Travis bug.

Basically, we're doing a hack-ish workaround because Travis si broken here.  For reasons that I don't understand completely, Travis decides to use pyenv for some PR's, but not others.  In this case, Travis wants to use pyenv for Python.  So, Travis prepends to the PATH and places /usr/bin earliy int he path, which is incorrect.  So, in Python _only_, we see a PATH that causes us to use /usr/bin/java (which is wrong, it's java-v8) instead of the correct path to a java v11.  This occurs _only_ inside Python due to some weird thing done _only_ by Travis.

src/rlimitfloatenv.cpp
gc00: Newlines done.

gc00:
@karya0, In the earlier version of the PR, you had accepted the PR.  I had to do something minor to get Travis to pass it.  Could you re-accpet this PR now?  Thanks.

karya0:
@gc00: Not sure if I follow. I already approved and merged this PR. Did I miss something?

PR #757, opened 2019-08-07
Duplicate of PR #618 for 3.0.

gc00:
This is a copy of PR #656 by @jiajuncao.  I'm pushing in this commit because it retains @jiajuncao as author, instead of hiding the origin via a merge.


pgrosu:
LGTM

karya0:
@rohgarg: Aren't we moving away from static std::string declarations? I think we'll need to revert this PR.

CC: @gc00

PR #761, opened 2019-08-09
shareddata: Fix a race when multiple clients join simultaneously

gc00:
This is a continuation of PR #722.  See that PR for the details.  Since @rohgarg was unavailable to rebase, and then push in the PR, I'm doing it here, while maintaining a commit with his signature.


pgrosu:
LGTM

PR #762, opened 2019-08-09
coordinator-api: Fix caching of coordinator host info

gc00:
This is a continuation of PR #728.  See that PR for the details. Since @rohgarg was unavailable to rebase, and then push in the PR, I'm doing it here, while maintaining a commit with his signature.)

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -51,8 +51,7 @@ namespace CoordinatorAPI {
 const int coordinatorSocket = PROTECTED_COORD_FD;
 int nsSock = -1;
 
-static bool _firstTime = true;
-static string _cachedHost;
+// Shared between getCoordHostAndPort() and setCoordPort()
 static int _cachedPort = 0;
 
 void init();
@@ -129,6 +128,13 @@ restart()
 void
 getCoordHostAndPort(CoordinatorMode mode, string *host, int *port)
 {
+  static bool _firstTime = true;
+  // FIXME:  Could make _cachedHost a 'char[]'.  But then, would
+  //         "*host = _cachedHost;"  replace the string inside *host as needed?
+  //         In particular, if the _cachedHost string object gets destroyed,
+  //         e.g., by a destructor during exit, then this could be a problem.
+  static string _cachedHost;
+
   if (SharedData::initialized()) {
     *host = SharedData::coordHost();
     *port = SharedData::coordPort();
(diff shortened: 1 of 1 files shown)


src/coordinatorapi.cpp
```diff
@@ -129,6 +128,13 @@ restart()
 void
 getCoordHostAndPort(CoordinatorMode mode, string *host, int *port)
 {
+  static bool _firstTime = true;
+  // FIXME:  Could make _cachedHost a 'char[]'.  But then, would
+  //         "*host = _cachedHost;"  replace the string inside *host as needed?
+  //         In particular, if the _cachedHost string object gets destroyed,
+  //         e.g., by a destructor during exit, then this could be a problem.
+  static string _cachedHost;
```
karya0: Let's not use non-pod static variables. Please convert it to a char array -- `*host = _cachedHost` will continue to work (will invoke copy constructor equivalent).

src/coordinatorapi.cpp
```diff
@@ -140,11 +146,19 @@ getCoordHostAndPort(CoordinatorMode mode, string *host, int *port)
     if (*host == "") {
       if (getenv(ENV_VAR_NAME_HOST)) {
         *host = getenv(ENV_VAR_NAME_HOST);
+        _cachedHost = getenv(ENV_VAR_NAME_HOST);
```
karya0: We'll need strncpy here.

src/coordinatorapi.cpp
```diff
@@ -140,11 +146,19 @@ getCoordHostAndPort(CoordinatorMode mode, string *host, int *port)
...
       } else if (getenv("DMTCP_HOST")) { // deprecated
         *host = getenv("DMTCP_HOST");
+        _cachedHost = getenv("DMTCP_HOST");
```
karya0: ditto.

src/coordinatorapi.cpp
```diff
@@ -140,11 +146,19 @@ getCoordHostAndPort(CoordinatorMode mode, string *host, int *port)
...
       } else {
         *host = DEFAULT_HOST;
+        _cachedHost = DEFAULT_HOST;
```
karya0: here too

src/coordinatorapi.cpp
```diff
@@ -140,11 +146,19 @@ getCoordHostAndPort(CoordinatorMode mode, string *host, int *port)
...
       }
+    } else {
+      // The caller's string object needs to be valid across
+      // multiple calls to this function, or else, the _cachedHost
+      // pointer will become a dangling pointer.
+      _cachedHost = host->c_str();
```
karya0: And here.

gc00: @karya0, It seems to me that what you're requesting should be the subject of a second PR or github issue.  This PR tries to fix a bug in which the hostname can be lost.  But it creates additional issues of concern (see below).  So, maybe it's better to separate out your concern as a github issue or a second PR to be addressed later by someone with more time available.

For the second proposed PR, we don't know `strlen(host->c_str())`.  So, in order to switch to `char[]` (actually to `char *`), we would need to do:
```
char *_cachedHost;
_cachedHost = JALLOC_HELPER_MALLOC(strlen(host->c_str())+1);
strcpy(_cachedHost, host->c_str());
```
This has its own dangers.  For example, there can be a small memory leak at this point in the code.  Of greater concern, `_cachedHost` is now pointing directly into the environment, and environment variables could possibly change during some later, orthogonal design change.  Of course, we could use `JALLOC_HELPER_MALLOC()` to copy environment strings, but then we create that small memory leak even in the standard case.

So when we're all done, it seems to me that we're just exchanging one hack (creating string objects in a constructor) for another hack (creating potential memory leaks using `char*` and `JALLOC_HELPER_MALLOC()`).

karya0: Hi @gc00, There was some issue that probably @rohgarg and @jiajuncao looked at and decided to remove the static non-pod members. I don't remember the exact details but it had something to do with segfault during teardown.

Regarding the length of host string, we can do:
```char _cachedHost[HOST_NAME_MAX] = { 0 };```

karya0: Didn't realize you inherited the static non-pod member from earlier. Approving this PR now. Will create another one to take care of the static aspect.

PR #767, opened 2019-08-15
Update changelog

alexmyczko:
drop unuploaded changelog entry, close bug, fix version

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,20 +1,12 @@
-dmtcp (2.6.0-1) testing; urgency=high
+dmtcp (2.6-1) testing; urgency=high
 
-  * New upstream release 2.6.0.
+  * New upstream release. (Closes: #791940)
   * Boosted the debian/compat for debhelper to 9.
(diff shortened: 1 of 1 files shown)


alexmyczko:
I also tried to modernize debian/rules, here's my complete package (with tests disabled, more arches in, tried sparc64/kfreebsd-amd64 (they fail))

http://phd-sid.ethz.ch/debian/dmtcp/26/

yarikoptic:
@alexmyczko I have not looked into your changes yet but FYI: modernization of `debian/*` should be done with backports in mind since dmtcp is also destined for a wide range of backports to be provided from neurodebian, see http://neuro.debian.net/pkgs/dmtcp.html .  So debhelper compat should be no greater than 10 ATM to get most of the recent debian/ubuntu releases covered

alexmyczko:
it should be 12 for sid and 10 for all backports then. can give it a try tomorrow

yarikoptic:
Sid version of debhelper still supports 10 afaik

PR #769, opened 2019-08-21
Added DmtcpMutex and DmtcpRWLock functions.

karya0:
These functions replace pthread_{mutex,rwlock} functions within DMTCP, thus decoupling us from the libc implementation. This enables us to track user pthread_mutex/rwlock requests without having to worry about calls originating from DMTCP code.

The functions are a drop-in replacement for pthread_* functions and should be side-effect free.

### DIFF
--- a/configure
+++ b/configure
@@ -7764,10 +7764,12 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 { $as_echo "$as_me:${as_lineno-$LINENO}: result: $gccAtomicBuiltins" >&5
 $as_echo "$gccAtomicBuiltins" >&6; }
 
-if test "$gccAtomicBuiltins" = "yes"; then
-
-$as_echo "#define HAS_ATOMIC_BUILTINS 1" >>confdefs.h
-
+if test "$gccAtomicBuiltins" = "no"; then
+  { { $as_echo "$as_me:${as_lineno-$LINENO}: error: in \`$ac_pwd':" >&5
+$as_echo "$as_me: error: in \`$ac_pwd':" >&2;}
+as_fn_error $? "Support for atomic built-ins is required to build DMTCP.
+                  Consider using GCC 4.1 or newer.
+See \`config.log' for more details" "$LINENO" 5; }
 fi
 
 { $as_echo "$as_me:${as_lineno-$LINENO}: checking for __WAIT_STATUS type" >&5
--- a/configure.ac
+++ b/configure.ac
@@ -670,9 +670,9 @@ AC_LINK_IFELSE(
 AC_LANG_POP([C++])
 AC_MSG_RESULT([$gccAtomicBuiltins])
 
-if test "$gccAtomicBuiltins" = "yes"; then
-  AC_DEFINE([HAS_ATOMIC_BUILTINS],[1],
-            [Can use __sync_bool_compare_and_swap(), etc.])
+if test "$gccAtomicBuiltins" = "no"; then
+  AC_MSG_FAILURE([Support for atomic built-ins is required to build DMTCP.
+                  Consider using GCC 4.1 or newer.])
 fi
 dnl Else will use mutex implementation; see jalib/jalloc.cpp
 
--- a/include/config.h.in
+++ b/include/config.h.in
@@ -18,9 +18,6 @@
 /* Child process does checkpointing */
 #undef FORKED_CHECKPOINTING
 
-/* Can use __sync_bool_compare_and_swap(), etc. */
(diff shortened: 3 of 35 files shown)


karya0:
> 1. Thanks very much for this important PR!  I'm now requesting an extra ...

Added a commit to re-enable pthread3.

> 2. Related to the issue of `--enable-mutex-wrappers`, I have a question.
>    It seems that with this PR, we no longer need `--enable-mutex-wrappers` or `HAS_MUTEX_WRAPPERS`.  If this is the case, then you can do `git revert 153d3625b6b1b637d4220de4b31999a226add0db` .  (But be careful, since `HAS_MUTEX_WRAPPERS` used to apply to all pthread tests, and more recently, it applied only to `pthread3`.)

We still use this flag to enable/disable pthread_mutex related wrappers (which are disabled by default).

> 3. Please do _not_ include the commit:
>    "Removed support for compilers without atomic-builtins."
>    [ Commit: [de47a0c](https://github.com/dmtcp/dmtcp/commit/de47a0ca615539e59dd82f0a16c4405fa00a004c) ]
>    Right now, DMTCP supports compilers even if they do not support the
>    gcc extension:
>    __sync_bool_compare_and_swap
>    This is important in keeping DMTCP close to widely-used standards.
>    The gcc "sync" extensions (gcc legacy intrinsics) are _not_ part of ANSI C.
> 
> Even the gcc-9.2 manual labels these as "Legacy __sync Built-in
> Functions for Atomic Memory Access" intended to be compatible with
> the ancient Itanium processor:
> https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/_005f_005fsync-Builtins.html#g_t_005f_005fsync-Builtins
> 
> DMTCP should _not_ refuse to build if these legacy functions are not
> available on the current compiler.

There are three alternatives here to get lock-free atomics (unless I am missing some other alternative) that are listed in the order of my preference:
1. We move to C11 and use the [atomics provided by the language](https://en.cppreference.com/w/c/atomic) itself; 
2. Use [gcc built-ins](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) equivalent to C11 atomics.
2. We implement our own atomic operations by directly using `cmpxchg` (and other platform-specific assembly instruction).

I would most-certainly prefer (1) or (2) given the extreme maintainability burden of (3). We can use (2) as an intermediate step until we move the code base to C11/C++11.

Thoughts?

> 2. QUESTION:  I would like to back-port this enhancement to DMTCP-2.6.1.
>    Do you see any difficulties, or do you have advice for me?  (I'll do the
>    work.)

I think this would be a fairly-straightforward port modulo some code-movement. Let me know if you need help with that.

gc00:
I agree with all your points, except one.  For item 3, let's omit that commit, and we can start a different ISSUE or PR or private conversation.  There's no need to delay this PR just because of only a partially related item 3.

===

Here, I continue the discussion solely about item 3:

To summarize for item 3, I had writen:
> 3.  Please do not include the commit:
>     "Removed support for compilers without atomic-builtins."
>    [ Commit: de47a0c ]
>    Right now, DMTCP supports compilers even if they do not support the gcc extension:
>    __sync_bool_compare_and_swap
   ....

You replied:
> There are three alternatives here to get lock-free atomics (unless I am missing some other alternative) that are listed in the order of my preference:

> 1. We move to C11 and use the atomics provided by the language itself;
> 2. Use gcc built-ins equivalent to C11 atomics.
> 3. We implement our own atomic operations by directly using cmpxchg (and other platform-specific assembly instruction).
> I would most-certainly prefer (1) or (2) given the extreme maintainability burden of (3). We can use (2) as an intermediate step until we move the code base to C11/C++11.

> Thoughts?

Hopefully, we can defer this to another PR/ISSUE/private conversation.  Could you remove this one commit for later discussion, and I'll approve the PR?

Concerning item 3, I would add a fourth mechanism:
4. Keep the current DMTCP strategy if atomics are not supported:  provide an _inefficient_, but simple and easy to maintain, workaround:  redefine the atomics we need by using mutex locks.

My preference would be to move to a combination of proposals (1), (2) as an intermediate step, and proposal (4) as an intermediate step.  Before making any decisions, I'd like to look up:
(a) what do other compilers like llvm support (gcc built-ins or C11 atomics or both)?
(b) with what version did gcc, llvm, etc., start support (1: C11 atomics)?
(c) are all of the well-known customers using either (1: C11 atomics) or (2: gcc built-ins)?  If not, we still need proposal (4).  I could make inquiries about this point.

PR #770, opened 2019-08-21
Removed some unused wrappers from DMTCP.

karya0:
This is just a cleanup of some code leftover from the FReD era. These wrappers (and the corresponding _real_ functions aren't needed anymore).


karya0:
This PR will be ready to be merged in right after #769 is merged into master.

PR #775, opened 2019-09-17
3.0 fix restart commit

gc00:
The commit message is self-explanatory.

### DIFF
--- a/src/mtcp/Makefile.debug
+++ b/src/mtcp/Makefile.debug
@@ -10,9 +10,6 @@ ifeq ($(M32),1)
   CFLAGS_M32 = -m32 -march=i386 -Wa,--32
 endif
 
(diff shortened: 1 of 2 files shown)


karya0:
@gc00: I'm curious if there is a reason to not use `-std=gnu99` flag for Makefile.debug. The rest of the code currently has `-std=gnu99` enabled by default on most (all?) platforms.

gc00:
@karya0 , actually, the configure.ac for the rest of DMTCP just says `AC_PROG_CC_STDC` in `configure.ac`.  On some platforms, that gets translated to --std=gnu99, but on others it can get translated differently (and the translation of `AC_PROG_CC_STDC` will chagne again in the future, based on future versions of C).  Thanks for the "approve".

PR #776, opened 2019-09-17
Windows WSL: fix 'make check'

gc00:
Windows build 1903 doesn't yet support MAP_GROWSDOWN.  Also, Windows-1903 doesn't support fcntl on pipes, and so 'make check' fails when it tries to use pipes for the test.

This is an incomplete fix that I had written about having ready a few weeks ago.  But I had wanted to do debugging including in src/mtcp_restart, before committing my PR.  Debugging in src/mtcp was made possible again only six days ago, with PR #772 .

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -151,8 +151,8 @@ static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo);
 
 
 #define MB                 1024 * 1024
-#define RESTORE_STACK_SIZE 5 * MB
-#define RESTORE_MEM_SIZE   5 * MB
+#define RESTORE_STACK_SIZE 16 * MB
+#define RESTORE_MEM_SIZE   16 * MB
 #define RESTORE_TOTAL_SIZE (RESTORE_STACK_SIZE + RESTORE_MEM_SIZE)
 
 // const char service_interp[] __attribute__((section(".interp"))) =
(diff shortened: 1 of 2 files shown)


gc00:
@karya0 , I've adjusted my PR to now make incremental changes on top of yourr PR #774 .  Could you approve this now?  Thanks.

karya0:
Hi @gc00, the mtcp_restart.c changes are not needed anymore. The `MAP_GROWSDOWN` shouldn't have been there in the first place (there is no place to grow per se).

Let's also keep the assert there. We can add a comment to explain its purpose. Basically, we have reserved some stack space during ckpt and if the mtcp_restart stack is bigger than the reserved space, this assert would trigger. If we ignore the error and continue, we might corrupt other mtcp_restart memory region or might segfault later on trying to access a part of extra stack that's not mapped properly.

In #774, I increased the stack size to 16 MB (from 5 MB) to accommodate the 8 MB mtcp_restart stack  on WSL. A more robust solution would remove the 16 MB limit and do something dynamically in mtcp_restart but I think that's a bit of a stretch given the current scope and size of our mtcp_restart. I think we can increase it further but I would want to know what distro/OS uses more than 16 MB to understand it better.

gc00:
@karya0 , `MAP_GROWSDOWN` was added by you in Nov., 2018.  If you're confident that we don't need it, I'll remove it, and update the PR.

I can also reinstate the assert for 16 MB.  I'm curious where the "8 MB mtcp_restart stack on WSL" comes from, and why WSL needs a larger restart stack.  Do you have any insights?  Thanks.

karya0:
> @karya0 , `MAP_GROWSDOWN` was added by you in Nov., 2018. If you're confident that we don't need it, I'll remove it, and update the PR.

Yeah, we don't need it. The flag was an attempt to handle a smaller mtcp_restart stack but then I ended up adding the ASSERT anyways (but didn't take out the MAP_GROWSDOWN flag at the time).

> I can also reinstate the assert for 16 MB. I'm curious where the "8 MB mtcp_restart stack on WSL" comes from, and why WSL needs a larger restart stack. Do you have any insights? Thanks.

If I have to make a guess, it's precisely because they don't support MAP_GROWSDOWN flag. Without that flag, it makes sense to just create a stack that is as big at the rlimit.

gc00:
@karya0 , I've made the changes that we agreed on.  Could you approve this now?  Thanks.

PR #780, opened 2019-09-19
Make plugin source independent of dmtcp/version.h

gc00:
Right now, a user (like me) will copy a plugin from test/plugin into their own directory.  They will then coyp `include/dmtcp.h` into that plugin directory and adjust `Makefile~, so ~DMTCP_INCLUDE=.`.  They will then do `make` to make binaries from the source.

This doesn't work because their is a hidden file, `include/dmtcp/version.h`, that is included through `include/dmtcp.h`.

So, the plugin API should not change with newer 3.x DMTCP versions (or at least not change too much).  But the user is not allowed to copy their plugin tarball to a new machine, with their built-in `dmtcp.h`, and compile and run using a public DMTCP executable in their path.  They also have a dependency on `/usr/include/dmtcp/version.h` (or `/usr/local/include/dmtcp/version.h` or some other location that they need to track down), so that they will inherit the version number of the production machine, and not the version number of the development version.  _Why do we care to enforce silly things like this?_  (The rest of DMTCP doesn't care about these versions.)

With this PR, the user can default `DMTCP_PACKAGE_VERSION` to `"3.0.0"` and not worry about it.

### DIFF
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -31,7 +31,9 @@
 # undef __USE_GNU
 #endif // ifdef __USE_GNU_NOT_SET
 
-#include "dmtcp/version.h"
+#ifndef DMTCP_PACKAGE_VERSION
+# include "dmtcp/version.h"
+#endif
 
 #ifdef __cplusplus
 # define EXTERNC extern "C"
--- a/test/plugin/applic-delayed-ckpt/Makefile
+++ b/test/plugin/applic-delayed-ckpt/Makefile
@@ -19,6 +19,11 @@ override CFLAGS += -I${DMTCP_INCLUDE}
 override CXXFLAGS += -I${DMTCP_INCLUDE}
 LINK = ${CC}
 
+# if version.h not found:
+ifeq (,$(wildcard ${DMTCP_INCLUDE}/dmtcp/version.h))
+  override CFLAGS += -DDMTCP_PACKAGE_VERSION='"3.0.0"'
+endif
+
 DEMO_PORT=7781
 
 default: ${LIBNAME}.so applic
--- a/test/plugin/applic-initiated-ckpt/Makefile
+++ b/test/plugin/applic-initiated-ckpt/Makefile
@@ -19,6 +19,11 @@ override CFLAGS += -I${DMTCP_INCLUDE}
 override CXXFLAGS += -I${DMTCP_INCLUDE}
 LINK = ${CC}
 
+# if version.h not found:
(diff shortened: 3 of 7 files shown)


karya0:
Hi @gc00: I understand the motivation behind not having to worry about the version string, but I am not sure about the right approach. The `dmtcp.h` change in PR is small enough and so I don't really mind merging it in (the rest of the changes are mostly internal, so don't matter as much) .

Having said that, I want a make a couple of related points here:
1. `dmtcp.h` has grown quite big and I want to split dmtcp.h further into two or more files.
    a. include/dmtcp/plugin.h -- that contains the plugin interface only.
    b. include/dmtcp/XXX.h -- for just using dmtcpaware-like interface.
    c. include/dmtcp/util.h -- C++ utilties (for logging, etc.) to make it easier for plugin developers to use this framework if they'd like.
    d. promote other include/*.h file for use by external dmtcp plugins. This would become handy for our binary distribution packages. We can install `/usr/include/dmtcp/util.h` but can't install `/usr/include/util.h`, so it makes more sense to put these files under the dmtcp directory.

2. Version information is useful to have for backwards compatibility. Historically we haven't supported any backwards compatibility per se, but as we are stabilizing the plugin API, it will become more important in the future.
    a. There is also the DMTCP version string that might come in handy when dealing with checkpoint images from different dmtcp versions. I can imagine long-running computations (several months) that might need an updated mtcp_restart to support some specific change in the platform. It's probably overengineering at the moment, but I don't want to entirely discount the possibility either.

Having said that, I do understand the additional burden it places on testing plugins right now, and so wonder if it would be easier if all our header files were inside a single `include/dmtcp/` directory? That way you can just copy the entire directory without having to worry about different include files (or, set the include path accordingly).

Thoughts?

gc00:
@karya0 , Thanks for your comments.  I think we're in one of thsoe "The perfect is the enemy of the good." moments.

For now, I'd like the user to just copy the single `dmtcp.h` file to their local development directory, and then to build a small plugin.  The motivation is that they can then copy it to any DMTCP on any host, and as long as the DMTCP versions are not wildly different, the tests should just work.  This is important for testing a plugin in multiple environments on different hosts.  So, I'm not worried about hardwiring in a default DMTCP version for now.

Eventually (when we target the "perfect" :-) ), we can also think about this use case, and perhaps include some mechanism for `DMTCP_VERSION_DEFAULT` and warn the user to use at their own risk --- or maybe we'll know an even better solution in the future.

In the meantime, could you approve this PR, so we support the use case above for the present?  Thanks.

PR #785, opened 2019-10-10
Add guard pages to restore-buffer memory region

JainTwinkle:
This pull request adds a guard page at each end of the restore-buffer memory region to make sure that the region is not merged with the same permission pages in proc maps. DMTCP can then correctly identifies and skip this region to write into a checkpoint image. (see mtcp_writememoryareas function in src/writeckpt.cpp)

### DIFF
--- a/src/writeckpt.cpp
+++ b/src/writeckpt.cpp
@@ -146,7 +146,11 @@ mtcp_writememoryareas(int fd)
   }
 
   /* Finally comes the memory contents */
+  JTRACE("addr and len of restoreBuf (to hold mtcp_restart code)")
+    ((void *)ProcessInfo::instance().restoreBufAddr())
+    (ProcessInfo::instance().restoreBufLen());
   procSelfMaps = new ProcSelfMaps();
+  // We must not cause an mmap() here, or the mem regions will not be correct.
   while (procSelfMaps->getNextArea(&area)) {
     // TODO(kapil): Verify that we are not doing any operation that might
     // result in a change of memory layout. For example, a call to JALLOC_NEW
@@ -454,7 +458,7 @@ writememoryarea(int fd, Area *area, int stack_was_seen)
     JTRACE("skipping over memory special section")
       (area->name) (addr) (area->size);
   } else if ( area->__addr == ProcessInfo::instance().vdsoStart() ) {
-    //vDSO issue:
+    //  vDSO issue:
     //    As always, we never want to save the vdso section.  We will use
     //  the vdso section code provided by the kernel on restart.  Further,
     //  the user code on restart has already been initialized and so it
@@ -475,7 +479,7 @@ writememoryarea(int fd, Area *area, int stack_was_seen)
     //  be saving the original vdso section (which is wrong), and we would
     //  be failing to save the user's memory that was restored into the
     //  location labelled by the kernel's "[vdso]" label.  This last
-    //  case is even worse, since we have now failed to restore some user data. 
+    //  case is even worse, since we have now failed to restore some user data.
     //    This was observed to happen in RHEL 6.6.  The solution is to
     //  trust DMTCP for the vdso location (as in the if condition above),
     //  and not to trust the kernel's "[vdso]" label.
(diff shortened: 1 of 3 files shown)


rohgarg:
Why do we need this? Don't we pass the restore address as part of the checkpoint image header? If we do, then `mtcp_restart` should just move itself to that address at restart time.

May be I'm missing some context. What is the scenario where we can hit a bug without these guard pages?

src/writeckpt.cpp
```diff
@@ -144,6 +144,20 @@ mtcp_writememoryareas(int fd)
...
+   *       However, this is a more cleaner solution.
+   */
+  JASSERT(munmap((void *)ProcessInfo::instance().restoreBufAddr(),
+            ProcessInfo::instance().restoreBufLen()) == 0)
+    ((void *)ProcessInfo::instance().restoreBufAddr())
+    (ProcessInfo::instance().restoreBufLen())
+    (JASSERT_ERRNO);
 
   /* Finally comes the memory contents */
   procSelfMaps = new ProcSelfMaps();
```
karya0: Thanks for the patch, @JainTwinkle! This is an important fix.

However, I think we need to do a bit more work here. The `new` operator might end up putting `ProcSelfMaps` object in the area that we just unmapped. This could lead to weird issues.

I suggest the following:
1. Move ProcSelfInfo constructor code (https://github.com/dmtcp/dmtcp/blob/master/src/procselfmaps.cpp#L57-L91) to a separate `init()` function.
2. Call the constructor before unmapping the reserved region.
3. Call `init()` after unmapping.

Does that sound reasonable to you?

gc00:
The `java1` test is failing in travis.  In a discussion between @JainTwinkle and me, I agreed to look at this issue more closely.  Once it works with `java`, we can push in the PR.

gc00:
@JainTwinkle , I looked more closely at why Travis was failing on the java1 test.

On the one hand, `./configure' has found 'java' and 'javac' in /usr/local/bin or in /usr/bin.
I should do a small test with Travis to have ./configure print out the answer to `javac --version` and `java-version` to see which java is being used.

Separately, I note that at the beginning of Travis, it reports that gcc has been compiled to enable `gcj`.  So, it's possible that `java` is invoking 'gcj'.  If so, this is not a common choice.  Finally, we can add:
`which java && java --version && which javac && javac --version` to `make display-build-environment` to directly see which java we're getting.

Next, note that autotest.py is using:  `java -Xmx5M java1`.  If those flags are not valid for the actual java that we're using, then autotest will fail on java1 simply because our runtime flags don't correspond to which java we're using -- and not due to a bug on our part.

Finally, maybe we should be testing on OpenJDK and OracleJDK.  (But OracleJDK might be considered 'non-free' and therefore not available.)  In any case, we can test on these by setting up our `.travis.jml` file to designate one or both Javas.  For more details, see here:
    https://www.deps.co/guides/travis-ci-latest-java/
and (maybe also) here:
    https://docs.travis-ci.com/user/languages/java/

Currently, our `.travis.yml` file at top-level in DMTCP doesn't specify anything at all with respect to Java.

When one of us has some spare time, let's go forward along the plan that I outline above.  Thanks.

gc00:
And this URL, https://stackoverflow.com/questions/29753560/how-to-reproduce-a-travis-ci-build-environment-for-debugging , tells how to create a Travis build environment for interactive debugging.  When one of us gets a chance, this might be the easiest way to fix this bug that appears only in Travis.

src/writeckpt.cpp
gc00: @rohgarg, In the end @JainTwinkle and I decided that her previous solution using guard pages was the simpler solution.  So, we're going back to the "guard pages" around the restoreBuf.

gc00:
@JainTwinkle , Ping!  This PR has been approved a while ago.  It's based on my modification of your original solution.  If you like this code, could you hit the "Rebase and merge" button on github.com?  (And if you don't like this, feel free to change it further.

Thanks.

PR #789, opened 2019-10-17
added a test pthread6 that expose a bug related to TID and a fix

JainTwinkle:
If a thread T1 sets its FS base register to another thread T2 then
gettid() in T1 returns the tid of T2 as dmtcp save tid of each thread
in a thread local variable called _dmtcp_thread_tid.
Note that any TLS variable of thread is accessed via FS base register.

This commit adds a test case i.e. pthread6 to expose the bug and resolves
the issue by intercepting the arch_prctl syscall and resets the tid back to
correct virtual tid.

### DIFF
--- a/src/plugin/pid/pid_miscwrappers.cpp
+++ b/src/plugin/pid/pid_miscwrappers.cpp
@@ -344,6 +344,23 @@ syscall(long sys_num, ...)
   va_start(ap, sys_num);
 
   switch (sys_num) {
+  case SYS_arch_prctl:
+  {
+    SYSCALL_GET_ARGS_2(int, code, void *, addr);
+    ret = _real_syscall(sys_num, code, addr);
+    if (ret == 0 && code == ARCH_SET_FS) {
+      /*
+        DMTCP saves each thread's tid in a thread local variable called
+        _dmtcp_thread_tid. It may be possible that one thread sets its FS base-
+        register to another thread. If so, dmtcp_gettid() will point to another
+        thread's _dmtcp_thread_tid. Note that glibc don't have any wrapper for
+        gettid().
+      */
+      // reset the tid after calling __real_syscall
+      dmtcpResetTid(REAL_TO_VIRTUAL_PID(_real_syscall(SYS_gettid)));
+    }
+    break;
+  }
   case SYS_gettid:
   {
     ret = dmtcp_gettid();
--- a/src/plugin/pid/pidwrappers.h
+++ b/src/plugin/pid/pidwrappers.h
@@ -84,6 +84,8 @@ struct user_desc {
 #include <sys/epoll.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <asm/prctl.h>
+#include <sys/prctl.h>
 #include <unistd.h>
 
 #include "dmtcp.h"
--- a/test/autotest.py
+++ b/test/autotest.py
@@ -841,6 +841,7 @@ S=DEFAULT_S
 
 runTest("pthread4",      1, ["./test/pthread4"])
 runTest("pthread5",      1, ["./test/pthread5"])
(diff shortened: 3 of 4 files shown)


rohgarg:
Thanks for the fix @JainTwinkle. I have a general design question before I start looking at the patch.

Why can't we change `dmtcp_getttid()` to make the gettid syscall (`_real_syscall(SYS_gettid)`), do the real-to-virtual translation, and return the virtual tid? Or why not just change the handling of `syscall(SYS_gettid)` within `pid_miscwrappers.cpp` in a similar fashion?

This way, we don't have to catch the `arch_prclt(ARCH_SET_FS)` syscall. Also, this will preserve the semantics of `gettid` from a user's perspective. `gettid` is anyway not defined by libc, so the only way for user code to use it is through `syscall(SYS_gettid)`. So, when they are doing this syscall, they are indeed expecting a syscall to take place (and any associated time cost). If any thing, DMTCP breaks this semantics by not doing a syscall and returning from a thread-local variable (`_dmtcp_thread_tid`). I think that our current design is violating the principle of transparency.

src/plugin/pid/pid_miscwrappers.cpp
```diff
@@ -344,6 +344,23 @@ syscall(long sys_num, ...)
...
+    ret = _real_syscall(sys_num, code, addr);
+    if (ret == 0 && code == ARCH_SET_FS) {
+      /*
+        DMTCP saves each thread's tid in a thread local variable called
+        _dmtcp_thread_tid. It may be possible that one thread sets its FS base-
+        register to another thread. If so, dmtcp_gettid() will point to another
+        thread's _dmtcp_thread_tid. Note that glibc don't have any wrapper for
+        gettid().
+      */
+      // reset the tid after calling __real_syscall
```
rohgarg: Why two separate comment styles? May be merge this with the comment above?

src/plugin/pid/pid_miscwrappers.cpp
```diff
@@ -344,6 +344,23 @@ syscall(long sys_num, ...)
   va_start(ap, sys_num);
 
   switch (sys_num) {
+  case SYS_arch_prctl:
+  {
+    SYSCALL_GET_ARGS_2(int, code, void *, addr);
+    ret = _real_syscall(sys_num, code, addr);
+    if (ret == 0 && code == ARCH_SET_FS) {
+      /*
+        DMTCP saves each thread's tid in a thread local variable called
```
rohgarg: thread-local variable

src/plugin/pid/pid_miscwrappers.cpp
```diff
@@ -344,6 +344,23 @@ syscall(long sys_num, ...)
...
+        _dmtcp_thread_tid. It may be possible that one thread sets its FS base-
+        register to another thread. If so, dmtcp_gettid() will point to another
```
rohgarg: to point to another thread's TLS.

rohgarg: If so, _dmtcp_thread_tid ...

src/plugin/pid/pid_miscwrappers.cpp
```diff
@@ -344,6 +344,23 @@ syscall(long sys_num, ...)
...
+        thread's _dmtcp_thread_tid. Note that glibc don't have any wrapper for
```
rohgarg: Add a line:

As a result, later, when the user calls gettid (using syscall(SYS_gettid), we'll end up returning incorrect result based on the value of _dmtcp_thread_tid.

rohgarg: Note that glibc doesn't have ...

rohgarg (changes requested):

PR #791, opened 2019-11-17
Fix epoll1.c for epoll2 test; Was silently failing

gc00:
The epoll2 test had never worked -- even from when it was created.  It was silently failing to bind, and `perror("bind")` reported "Bad permission", but the process didn't exit.  So, the checkpoint and restart of that trivial process was succeeding, even though our test was failing.

The new Python3-aware version of autotest.py was failing on the old epoll2, but for some reason, the
Python2 running of autotest.py did not turn this  up.

I'm not sure why the two Pythons should give different results.  Apprently, a checkpoint is taken before the epoll2 processes can exit, and checkponit files are created.  But I note that after checkpoint, there is a defunct coordinator left over, and on restart, it tries to use the defunct cocoridnator and hanges, which autotest.py interprets as PASS.  I've seen enough so far to be sure there is a bug in our dmtcp_restart.cpp in the case of 'epoll2'.

### DIFF
--- a/test/epoll1.c
+++ b/test/epoll1.c
@@ -13,11 +13,11 @@
 #include <stdio.h>
 
 #define MAX_EVENTS 10
-/* Code to set up listening socket, 'listen_sock',
+/* Code to set up listening socket, with 'listen_sockfd',
    (socket(), bind(), listen()) omitted */
 
 struct sockaddr_in sockaddr;
-int listen_sock;
+int listen_sockfd;
 int use_epoll_create1 = 0;
 
 void server()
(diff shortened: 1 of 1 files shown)


test/epoll1.c
```diff
@@ -118,15 +118,26 @@ main(int argc, char *argv[])
...
+    rc = bind(listen_sockfd, (struct sockaddr *)&sockaddr, sizeof(sockaddr));
   } while (rc == -1 && errno == EADDRINUSE);
   if (rc == -1) {
     perror("bind");
+    exit(1);
   }
-  listen(listen_sock, 5);
+  listen(listen_sockfd, 5);
+
+  // This is informmational only; it could be deleted.
```
rohgarg: informmational -> informational

gc00: The code was failiing due to a bug in the first line of 'main()'.  The test epoll2 in autotest.py says to run `./test/epoll1 --use-epoll-create`.  The ffirst line of main() says `int port = (argc == 2 ? atoi(argv[1]) : 6500);`  `argv[1]` is a `char *`.  But atoi converts that to an `int` without any type checking.  So, `atoi()` is wrong when it is give `"--use-epoll-create"` as its argument.

Lower down, the code starts with port 6500, and then successively looks at 6501, 6502, ..., until it finds a ffree port.  That part of the code seems correct.

Originally, whoever wrote epoll1.c was probably intending that it would take at most one argument:  a port number.  Later, they forgot that, and decided that it would take at most one argument, the string `"--use-epoll-create"`.

The rest of my changes were just small cleanups (e.g., sockfd instead of sock, to correspond to the man page).  I'll update `informmational`.  Thanks for that fix.

rohgarg: Thanks for the clarification, Gene.

LGTM, after the typo fix.

PR #796, opened 2019-12-05
jalib/jalloc.c: 'size_t _blockSize;' wasn't initialized to 0

gc00:
In `jalib/jalloc.c`, `size_t _blockSize;` was not initialized to 0.

This one-line bug fix initializes it to `0`.  This should be easy to review.

### DIFF
--- a/jalib/jalloc.cpp
+++ b/jalib/jalloc.cpp
@@ -220,7 +220,7 @@ class JFixedAllocStack
 
   private:
     FreeItem *volatile _root;
-    size_t _blockSize;
+    size_t _blockSize = 0;
     char padding[128];
     int volatile _numExpands;
 };


gc00:
Thanks @rohgarg.  I accidentally used 2.6 instead of master for the PR.  I think I've fixed it now, and it's now based on master.  There is now only 1 commit, as there should be.  It's ready for review.

jalib/jalloc.cpp
```diff
@@ -220,7 +220,7 @@ class JFixedAllocStack
 
   private:
     FreeItem *volatile _root;
-    size_t _blockSize;
+    size_t _blockSize = 0;
```
rohgarg: Was this a compiler warning/error issue? Or some other bug?

(Will this work with older compilers? Non-static data member initialization is probably a C++11 thing. Since we depend on C++11, we'd probably fail to compile with older compilers anyway. So I guess this is fine.)

gc00: In g++-9.2, the compiler issues a warning here about an uninitialized data member.

In fact, this causes a bug.  If you look higher up in this same file, you'll see code sindie the constructor of JFixedAllocStack that says:
```
    if (_blockSize == 0) {
      _blockSize = 4*MAX_CHUNKSIZE;
```
After the constructor is called, there is still a call within `jalib::JAllocDispatcher::initialize()` that inicializes each level once again and sets `_blockSize`.  That seems to occur at the first call to `jalib::JAllocDispatcher::allocate()`.

So, it may be that the main effect of this PR is simply to make the g++-9.2 warning go away.  I'm not sure if it has an effect on the actual code, given that `initialize()` will be called from within the first call to `allocate()`.

I came back to look at this code because of a bug report from a user.  But that user has not yet reported if this hapens to fix the bug.  It might be that this has no effect.

gc00: @rohgarg, I've verified that using the C+11 feature on login at Khoury (CentOS 7.6) doesn't break anything.  They're using gcc-4.8, and it produces a warning that it is accepting the use of the C+11 feature.  So, I think the private initializer is okay.

karya0: Dismissing @rohgarg review to merge it. It's blocking me :-).

PR #805, opened 2019-12-11
configure, and 'autotest hellompich-n[12]' now tests for MPICH3

gc00:
DMTCP used to test for an obsolete version of MPICH, using mpidboot, etc.  So, it would fail when testing on MPICH3.  The updated test now assumes MPICH3, with the modern, cleaner syntax.


configure.ac
```diff
@@ -611,33 +611,50 @@ else
...
+                            [Compile mpich3 tests in `make check`])],
+            [mpich_path_prefix=$withval],
+            [mpich_path_prefix=no])
+
+mpich_path="no"
+if test "$mpich_path_prefix" != "no"; then
+  mpich_path="$mpich_path_prefix:$mpich_path_prefix/bin"
+else
+  # Let's see if we can find mpicc in the default path
+  mpiccpath=`which mpicc`
```
karya0: Hi @gc00, should we just set `mpiccpath` to `$PATH` (or the autoconf equivalent) instead? I vaguely remember debian or some other place frowning upon use of `which` to find executables.

If we don't find mpich in the user-path or system-path, we can simply say `not found`. Is this also your goal to find if the user gave us wrong mpich path?

PR #806, opened 2019-12-11
connectionrewirer: Fix bug in creation of restore socket for IPv4 sockets

rohgarg:


src/plugin/ipc/socket/connectionrewirer.cpp
```diff
@@ -170,13 +170,17 @@ ConnectionRewirer::openRestoreSocket(bool hasIPv4Sock,
 
   // Open IP4 Restore Socket
   if (hasIPv4Sock) {
-    jalib::JServerSocket restoreSocket(jalib::JSockAddr::ANY, 0);
+    // Bind and listen on all local interfaces
```
gc00: @rohgarg , could you expand on this comment?

I see that `const jalib::JSockAddr jalib::JSockAddr::ANY(NULL);`.  So, sockAddr appears to be the default copy constructor that takes `NULL` (of type `jalib::JSockAddr`) and binds it to `sockAddr`.  After that, we construct `restoreSocket` using 'sockAddr' instead of directly using 'alib::JSockAddr::ANY'.  Why do we need to use this copy called 'sockAddr' instead of the original value?

Thanks.

rohgarg: @gc00: Here, we are constructing a `jalib::JSockAddr` object that can be used to bind to all local interfaces (loopback, and others).

Later, we use this object to construct a`jalib::JServerSocket` object, `restoreSocket` (line 175). As part of the object construction, we also bind the socket to the specified interfaces and port (See [here](https://github.com/dmtcp/dmtcp/blob/master/jalib/jsocket.h#L186).)

In order to initialize `_ip4RestoreAddr.sin_addr`, we'd like to access the socket address of the `restoreSocket` object (line 181). Unfortunately, the `jalib::JServerSocket` class does not provide any interface that can allow us to access the socket address being used by the socket.

For this reason, I had introduced this additional local `jalib::JSockAddr` object (`sockAddr`), which I could use in two places: (1) to create the `jalib::JServerSocket` object (`restoreSocket`), and (2) to initialize the fields of `_ip4RestoreAddr` struct.

gc00: I think your next to last paragraph can be converted into an appropriate comment.  Maybe something like:
```
// In order to initialize _ip4RestoreAddr.sin_addr, we'd like to access the socket address
// of the restoreSocket object (line 181). Unfortunately, the jalib::JServerSocket class
// does not provide any interface that can allow us to access the socket address being
// used by the socket.  So, sockAddr is introducted to create the JServerSock and also
// initialize _ip4RestoreAddr .
```

PR #809, opened 2019-12-13
autotest.py syscall-tester: ckpt w/ -Kc: no resume

gc00:
This fixes a long-standing bug that affected `./autotest.py --slow syscall-tester`.  The autotest file uses `b'Kc'` to do an exit after checkpoint that kills the process after the first checkpoint.  As explained in the autotest.py file, this is needed for a correct test in case of syscall-tester calling 'unlink'.  But autotest.py had a line that expected the process to resume after checkpoint.  If `b'Kc'` is used, the process will _not_ resume after checkpoint.  So now, if `b'Kc'` was used, it's not an error if the process doesn't resume.

### DIFF
--- a/test/autotest.py
+++ b/test/autotest.py
@@ -745,7 +745,7 @@ resource.setrlimit(resource.RLIMIT_STACK, [newCurrLimit, oldLimit[1]])
 runTest("dmtcp5",        2, ["./test/dmtcp5"])
 resource.setrlimit(resource.RLIMIT_STACK, oldLimit)
 
-# Test for a bunch of system calls. We want to use the 'kc' mode for
+# Test for a bunch of system calls. We want to use the 'Kc' mode for
 # (sets exitAfterCkptOnce in src/dmtcp_coordinator.cpp) for
 # checkpointing so that the process is killed right after checkpoint. Otherwise
 # the syscall-tester could fail in the following case:


test/autotest.py
```diff
@@ -741,7 +741,7 @@ def saveResultsNMI():
 runTest("dmtcp5",        2, ["./test/dmtcp5"])
 resource.setrlimit(resource.RLIMIT_STACK, oldLimit)
 
-# Test for a bunch of system calls. We want to use the 'xc' mode for
+# Test for a bunch of system calls. We want to use the 'bc' mode for
 # checkpointing so that the process is killed right after checkpoint. Otherwise
```
rohgarg: > so that the process is killed right after checkpoint"

Could you please update the comment? This statement is no longer valid.

test/autotest.py
```diff
@@ -753,7 +753,7 @@ def saveResultsNMI():
 # program will try to unlink the file once again, but the unlink operation will
 # fail, causing the test to fail.
 old_ckpt_cmd = CKPT_CMD
-CKPT_CMD = 'xc'
+CKPT_CMD = b'bc'
```
rohgarg: Isn't 'bc' for "blocking until done"? And 'xc' is for "kill after checkpoint"? Has the behavior changed recently?

(I just checked `dmtcp_command.cpp` and the 'xc' command is commented out. So it seems like we don't support that anymore.

Also, the documentation for 'bc' in `dmtcp_command.cpp` (line 52) says:

> -bc, --bcheckpoint:    Checkpoint all nodes, blocking until done

If this is true, then then test would fail after this change. According to the comment on line 751 above, the program would unlink the temp file after resuming from checkpoint, and then the unlink operation would fail on restart, causing the test to fail.)

test/autotest.py
rohgarg: Please see my comment below. I thought 'bc' is for "blocking until done".

rohgarg (dismissed):

test/autotest.py
gc00: When I have time, I'll look mroe closely.  Apparently in the code, 'xc' and 'bc' were always synonyms.  I think that it causes exit after first checkpoint.  But I'd need to check that to be sure.

Also, what exactly does 'blocking until done' mean?  This is not a meaningful phrase for me.  Should we update teh English explanation of the command?

test/autotest.py
gc00: I looked at when 'xc' and 'bc' were introduced.  It was around 2013.  I think the intention was that it was originally 'xc', and we wanted to change the name to 'bc', but I'm not sure.  In any case, from the beginning, `src/dmtcp_command.cpp` shows that these two commands were always synonyms, and never had different effects.

So, I think we're donig a late cleanup here, for somethign that happened in 2013.

test/autotest.py
rohgarg: >  Should we update teh English explanation of the command?

Yes.

I think 'xc' is to kill the program after resuming from a checkpoint, and 'bc' is for forcing 'dmtcp_command' to block or wait synchronously until the checkpointing is completed. With just 'c', `dmtcp_command` sends the command to the coordiantor and returns immediately, without waiting for the checkpoint to complete. I have used both the commands in the past for different use cases. I don't know when 'xc' was deprecated.

gc00: @rohgarg, You're correct in all of the above.  I'm now using 'xc' again as I should.  Also, if you look at commit ee798b0 , you'll see why we had commented out the documentation on the 'xc' command.  That was a mistake and it is now documented as it should be.

test/autotest.py
gc00: @rohgarg, You're correct in all of the above.  See my newer description in this PR now.

gc00:
@rohgarg, Ping!  (I think I've addressed the concerns on this PR.)

gc00:
@rohgarg, Ping! (I think I've addressed the concerns on this PR.)

test/autotest.py
gc00: And since this PR has been waiting for so long, we've already changed 'xc' to 'Kc' in the above code.  Wherever 'xc' is written, the code now uses 'Kc', after doing a rebase and resolving conflicts.

elIdrissiYounes (approved):
LGTM!

elIdrissiYounes (approved):
LGTM!

PR #810, opened 2019-12-14
dmtcp_coordinator: Change 'xc' (ckpt & exit) to 'kc' (ckpt & kill peers)

gc00:
Logically, the dmtcp_coordinator is asking to "kill the peers after checkpoint".  'xc' ('--exit-after-ckpt' flag) was getting confused with the dmtcp_coordinator flag:  '--exit-on-last'.  So, now we have:
`--kill-after-ckpt` and `--exit-on-last`.  So, we also change 'xc' to 'kc' to correspond.

I've tested this locally, and it works for me.  This is ready for review.

====

As for the history, the 'xc' command was never documented to users.
In the usage command for src/dmtcp_command.cpp, the 'xc' command was commented out in 2013 at commit ee798b0 .   (Maybe it was always commented out.)  The commit comment says "
```
The command to set the flag is 'x' and it must be followed by a 'c'
    command.
    
    Not documenting it right now to allow a name/letter change later.
```
I guess it's about time to document it now. :-) The command is not going to be changing.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -93,6 +93,7 @@ static const char *theHelpMessage =
   "  l : List connected nodes\n"
   "  s : Print status message\n"
   "  c : Checkpoint all nodes\n"
+  "  Kc : Checkpoint and then kill all nodes\n"
   "  i : Print current checkpoint interval\n"
   "      (To change checkpoint interval, use dmtcp_command)\n"
   "  k : Kill all nodes\n"
@@ -115,7 +116,7 @@ static const char *theUsage =
   "      Directory to store temporary files (default: env var TMDPIR or /tmp)\n"
   "  --exit-on-last\n"
   "      Exit automatically when last client disconnects\n"
-  "  --exit-after-ckpt\n"
+  "  --kill-after-ckpt\n"
   "      Kill peer processes of computation after first checkpoint is created\n"
   "  --daemon\n"
   "      Run silently in the background after detaching from the parent "
@@ -144,8 +145,8 @@ static string thePortFile;
 
 static bool exitOnLast = false;
 static bool blockUntilDone = false;
-static bool exitAfterCkpt = false;
-static bool exitAfterCkptOnce = false;
+static bool killAfterCkpt = false;
+static bool killAfterCkptOnce = false;
 static int blockUntilDoneRemote = -1;
 
 static DmtcpCoordinator prog;
@@ -281,9 +282,9 @@ DmtcpCoordinator::handleUserCommand(char cmd, DmtcpMessage *reply /*= NULL*/)
     JTRACE("blocking checkpoint beginning...");
     blockUntilDone = true;
     break;
-  case 'x': case 'X':  // prefix exit command, prior to checkpoint command
-    JTRACE("Will exit after creating the checkpoint...");
-    exitAfterCkptOnce = true;
+  case 'K': case 'x':  // prefix kill command, after ckpt cmd ('x' deprecated)
+    JTRACE("Will kill peers after creating the checkpoint...");
+    killAfterCkptOnce = true;
     break;
   case 'c': case 'C':
     JTRACE("checkpointing...");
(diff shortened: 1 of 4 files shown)


src/dmtcp_coordinator.cpp
```diff
@@ -281,9 +282,9 @@ DmtcpCoordinator::handleUserCommand(char cmd, DmtcpMessage *reply /*= NULL*/)
     JTRACE("blocking checkpoint beginning...");
     blockUntilDone = true;
     break;
-  case 'x': case 'X':  // prefix exit command, prior to checkpoint command
-    JTRACE("Will exit after creating the checkpoint...");
-    exitAfterCkptOnce = true;
+  case 'K': case 'x':  // prefix kill command, after ckpt cmd ('x' deprecated)
```
karya0: Hi @gc00, I think there is a bug in this PR. If we type `kc` in the coordinator window, it would immediately kill the computation without checkpointing (because `k` is interpreted first). `xc` was a special case since `x` wasn't used anywhere else and so pairing it with `c` was ok.

However, with this new change, if we press `k` in the coordinator window, it wouldn't kill the computation immediately. 

Minor: you forgot to remove the other `x` :-).

gc00:
@karya0, there is a difference between 'k' (lower case) and 'Kc' (upper case).  I think maybe you weren't reading the different cases.  Also, I was keeping 'x' in the code for backward compatibility, but I can remove that if you like.  Separately, dmtcp_command can separately offer 'kc' and 'k', but maybe for consistency, dmtcp_command should offer 'Kc' and 'k'.

The help on dmtcp_coordinator now has the help msg:

  "COMMANDS:\n"                                                                                                          
 "  l : List connected nodes\n"                                                                                          
"  s : Print status message\n"                                                                                         
 "  c : Checkpoint all nodes\n"                                                                                          
"  Kc : Checkpoint and then kill all nodes\n"                                                                          
 "  i : Print current checkpoint interval\n"                                                                            
 "      (To change checkpoint interval, use dmtcp_command)\n"                                                              "  k : Kill all nodes\n"                                                                                               
 "  q : Kill all nodes and quit\n"                                                                                       
"  ? : Show this message\n"                                                                                            
 "\n";

karya0:
> @karya0, there is a difference between 'k' (lower case) and 'Kc' (upper case). I think maybe you weren't reading the different cases. Also, I was keeping 'x' in the code for backward compatibility, but I can remove that if you like. Separately, dmtcp_command can separately offer 'kc' and 'k', but maybe for consistency, dmtcp_command should offer 'Kc' and 'k'.

Aah, I didn't realize the case change was intentional -- I do like the idea. So why not add `C` (upper case) for checkpoint and kill? It can be in addition to `Kc` or we drop `Kc` altogether. Also, we can drop `x`. This wasn't documented, so we don't have to worry about backwards compatibility.

gc00:
@karya0 , I refuse to accept your approval just yet.  :-)

I like your idea of making 'C' (capital) a synonym for 'Kc'.  Then we can go back to a single letter for each option.  But currently, the capital version of each option in dmtcp_command is a synonym for the lower case version of each option.  That's why I offer 'Kc' instead of just 'K' right now.  The 'K' option is currently a synonym for lower-case 'k'.

But I don't mind switching to single-letter options in all cases.  I think that's cleaner.  In that case, we should abandon all capital letters that are synonyms for the lower-case letter.  But if we do that, should we make 'K' or 'C' or both the command for checkpoint and then kill?  And does 'b' or 'B' become the option for 'block until the checkpoint is complete'?

I'd like to get your view on what would be cleanest before I push this in.  Should we make the minimal change (delete the undocumented 'xc' option) or make a larger change?  Thanks.

karya0:
> @karya0 , I refuse to accept your approval just yet. :-)
> 
> I like your idea of making 'C' (capital) a synonym for 'Kc'. Then we can go back to a single letter for each option. But currently, the capital version of each option in dmtcp_command is a synonym for the lower case version of each option. That's why I offer 'Kc' instead of just 'K' right now. The 'K' option is currently a synonym for lower-case 'k'.
> 
> But I don't mind switching to single-letter options in all cases. I think that's cleaner. In that case, we should abandon all capital letters that are synonyms for the lower-case letter. But if we do that, should we make 'K' or 'C' or both the command for checkpoint and then kill? And does 'b' or 'B' become the option for 'block until the checkpoint is complete'?

Let's abandon all capital letters that are synonyms for the lower-case letter. Most other Unix tools are that way so let's stay close to standard practice.
 
> I'd like to get your view on what would be cleanest before I push this in. Should we make the minimal change (delete the undocumented 'xc' option) or make a larger change? Thanks.

Two-letter commands are a pain anyways, so let's remove "xc".

gc00:
I added an extra commit following your suggestions:
* Get rid of upper case synonyms for lower case
* Remove 'xc'

PR #811, opened 2019-12-19
Remove compiler warnings from processinfo.cpp

gc00:
The PR title says it all.  I introduced LIB_FNC_OFFSET_FAILED as a possible return value for get_dlsym_lib_fnc_offset().  This allows us to replace `ptrdiff_t` by `uint64_t` everywhere.  So, we no longer have warnings about mismatch between 'uint64_t' and 'ptrdiff_t'.


gc00:
Hi @jiajuncao , When I chose uint64_t as the basic data type (instead of ptrdiff_t), I did it in analogy with 'mmap'.  The 'mmap' call returns 'void *', and MAP_FAILED is a cast of -1 to 'void *'.

Your point is an interesting one.  But in this case, the function of interest is 'get_dlsym_lib_fnc_offset()'.  So I can't use 'void *' as an offset.  An alternative to 'uint64_t' might be 'off_t'.  Would that be a better return type?

I didn't want to use a return type of 'ptrdiff_t' because that is inherently a _signed_ value.  And 'get_dlsym_lib_fnc_offset()' tries to return an _unsigned_ value.  But 'off_t' might be a good choice.  Should I use that?

jiajuncao:
@gc00,

> Your point is an interesting one. But in this case, the function of interest is 'get_dlsym_lib_fnc_offset()'. So I can't use 'void *' as an offset. An alternative to 'uint64_t' might be 'off_t'. Would that be a better return type?

I think `off_t` applies more to file size. And it's signed as well, if I remember correctly.

> 
> I didn't want to use a return type of 'ptrdiff_t' because that is inherently a _signed_ value. And 'get_dlsym_lib_fnc_offset()' tries to return an _unsigned_ value. But 'off_t' might be a good choice. Should I use that?

Can you explain a bit more here? I don't understand why returning unsigned is important.

src/dmtcp_dlsym.cpp
```diff
@@ -612,13 +612,13 @@ dmtcp_dlsym_lib(const char *libname, const char *symbol)
...
- * or -1 if the function does not exist in the library
+ * or LIB_FNC_OFFSET_FAILED if the function does not exist in the library
  */
-EXTERNC ptrdiff_t
+EXTERNC uint64_t
 dmtcp_dlsym_lib_fnc_offset(const char *libname, const char *symbol)
 {
   dt_tag tags;
-  ptrdiff_t ret = -1;
+  uint64_t ret = LIB_FNC_OFFSET_FAILED;
```
gc00: Lower down, just after line 631, I should add `JASSERT(ret >= 0)(result)(tags.base_addr)`.

I should then add a comment that `ret` is either positive (non-negative), or else -1 (LIB_FNC_OFFSET_FAILED).  This is in analogy with 'mmap' returning either an unsigned 'void *' or -1 (MAP_FAILED).

gc00:
@jiajuncao,
Thanks for your comments.  It's helping me to clarify the code so that it can be maintained better.
> Can you explain a bit more here? I don't understand why returning unsigned is important.

[ Recall that the compiler warning occurs because of a comparison of a signed and unsigned quantity.  So, the question is whether to cast everything to 'ptrdiff_t' (signed) or 'uint64_t' (unsigned).]
1.  First, you're right and I was wrong about 'off_t'.  'off_t' is signed.  So, I still want to use an unsigned quantity like `uint64_t`.
2.  The key to your question about 'unsigned' is to recognize that the type of the return value of 'dmtcp_dlsym_lib_fnc_offset()' is always a positive number or '-1' (what I'm calling 'LIB_FNC_OFFSET_FAILED').  That's why I want to use a type 'uint64_t'.
3.  If you'd like to dig deeper, you can read inside the code of 'dmtcp_dlsym_lib_fnc_offset)()', the code:
```
  void *result = dlsym_default_internal_flag_handler(...);
  if (result) {
    ret = (char*)result - tags.base_addr;
 }
```
To see why 'ret' is positive, you next need to read the code of 'dlsym_default_internal_library_handler()' in dmtcp_dlsym.cpp.  That function has a return type of 'void *', and it has a line:
```
return tags.base_addr + tags.symtab[default_symbol_index].st_value
```
Finally, 'st_value' must be positive, based on the ELF spec for symtab (symbol table).

4.  In the "Files changed", I've now added an in-line comment proposing that I add a `JASSERT(ret >= 0);`, and then some further comments of clarification.  This might explain it to others who read this code.

If I add the 'JASSERT()' and add the proposed comment, will that make this PR clear?  Thanks.

===
PREVIOUS CONVERSATION:
> @gc00,
> 
> > Your point is an interesting one. But in this case, the function of interest is 'dmtcp_dlsym_lib_fnc_offset()'. So I can't use 'void *' as an offset. An alternative to 'uint64_t' might be 'off_t'. Would that be a better return type?
> 
> I think `off_t` applies more to file size. And it's signed as well, if I remember correctly.
> [Gene:  I just checked and you're right.  It is signed.]
> 
> > I didn't want to use a return type of 'ptrdiff_t' because that is inherently a _signed_ value. And 'dmtcp_dlsym_lib_fnc_offset()' tries to return an _unsigned_ value. But 'off_t' might be a good choice. Should I use that?
> 
> Can you explain a bit more here? I don't understand why returning unsigned is important.

src/dmtcp_dlsym.cpp
jiajuncao: >     1. In the "Files changed", I've now added an in-line comment proposing that I add a `JASSERT(ret >= 0);`, and then some further comments of clarification.  This might explain it to others who read this code.

Now I'm more confused :-) Doesn't casting `ret` to `uint64_t` make it always non-negative?

jiajuncao:
>     1. In the "Files changed", I've now added an in-line comment proposing that I add a `JASSERT(ret >= 0);`, and then some further comments of clarification.  This might explain it to others who read this code.

Now I'm more confused :-) Doesn't casting ret to uint64_t make it always non-negative?

karya0:
> > ```
> > 1. In the "Files changed", I've now added an in-line comment proposing that I add a `JASSERT(ret >= 0);`, and then some further comments of clarification.  This might explain it to others who read this code.
> > ```
> 
> Now I'm more confused :-) Doesn't casting `ret` to `uint64_t` make it always non-negative?

Hi @jiajuncao, I think the idea here is to distinguish a successful return (>0) from a failure (-1). It's the caller's responsibility to properly handle a `-1` return value, otherwise it might end up with improper address calculation.

karya0:
Side note: `ptrdiff_t` and `off_t` are inherently bad for us since they have different sizes for 32-bit and 64-bit platforms. So I do like having fixed size integers (`int64_t`, etc.) even though we sometimes need to handle them properly on 32-bit arch.

gc00:
@jiajuncao , think of this in analogy with 'mmap()'.  mmap returns a 'void *' (similar in spirit to our 'uint64_t'), and both mmap and our fuunction return '-1'  for error.  They do this by casting '-1' to 'uint64_t'

@karya0 or @jiajuncao, could one of you approve this.  It's annoying to see that constant compiler warning.  :-)  Thanks.

PR #814, opened 2019-12-27
Fix: Coord didn't reset alarm() if nothing running

gc00:
I found a better fix for the original problem and for a second problem.  This new version has nothing to do with the earlier version under this PR.  The git commit also has some explanation.
  (And the second commit is just to improve the error msg.  It was too small for a separate PR.)

[FROM git commit:]
 - This fixes a long pause (e.g., GDB) when at main() in the target application. 
   The ckpt interval timer goes off at this time 
 - In addition, it fixes a problem of:  dmtcp_launch and break at execvp (e.g., in GDB) 
   In this case also, the ckpt interval timer goes off, and the workerState is not yet running\
   or the last worker exited and the worker state is UNKNOWN                                       

[FURTHER EXPLANATION:]
This fixes two bugs.  Both can occur when using GDB for debugging and there are long pauses while using a ckpt interval timer.
1.  Even though `alarm()` was called by the coordinator for periodic checkpointing, the alarm might have gone off at a time when there was no active executable running under DMTCP control.  In this case, no checkpoint was executed.  But we would call resetCkptTimer() only at the last step before finishing the checkpoint.  (That's important, since we don't want the interval timer going off in the _middle_ of an ongoing checkpoint.)  So, in this scenario, 'theCheckpointInterval' is correct and 'timerExpired' is false, as they should be.  But we never called 'alarm()' again for the next checkpoint interval.

2.  A second bug was occurring when we start the dmtcp_coordinator() and then start the application (e.g., test/plugin/presuspend).  If we run the application under GDB and stop at main within the application and then pause while the checkpoint interval timer goes off, then the worker state can be set to RUNNING, and then set to PRESUSPEND early.  Then, some JASSERT's fail that were expecting the state RUNNING.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -47,6 +47,7 @@
  * Checkpoint: RUNNING -> SUSPENDED -> CHECKPOINTING                        *
  *                     -> (Checkpoint barriers) -> CHECKPOINTED             *
  *                     -> (Resume barriers) -> RUNNING                      *
+ *             [State returns to UNKNOWN if no active workers.]
  *                                                                          *
  * Restart:    RESTARTING -> (Restart barriers) -> RUNNING                  *
  * If debugging, set gdb breakpoint on:                                     *
@@ -867,6 +868,7 @@ DmtcpCoordinator::onConnect()
     client->virtualPid(hello_remote.from.pid());
     _virtualPidToClientMap[client->virtualPid()] = client;
   } else if (hello_remote.type == DMT_NEW_WORKER) {
+    // Comping from dmtcp_launch or fork(), ssh(), etc.
     JASSERT(hello_remote.state == WorkerState::RUNNING ||
             hello_remote.state == WorkerState::UNKNOWN);
     JASSERT(hello_remote.virtualPid == -1);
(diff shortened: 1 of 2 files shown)


src/dmtcp_coordinator.cpp
```diff
@@ -660,6 +660,20 @@ DmtcpCoordinator::onData(CoordClient *client)
...
+    //       In that case, no checkpoint occurred and
+    //       'DmtcpCoordinator::releaseBarrier()' failed to be called
+    //       and so resetCkptTimer() was not called again.
+    //  (ii) Maybe 'dmtcp_coordintor -i 5' set a checkpoint interval.
+    //       And maybe one job, 'dmtcp_launch --join-coorindator a.out'
+    //       completed.  After that, there were no jobs, and so, as before,
+    //       resetCkptTimer() was not called at the next alarm, since there
+    //       were no jobs.
+    // So, we reset the ckpt timer now, when a new job is init'ed or exec'ed.
+    resetCkptTimer();
```
karya0: Resetting alarm here could prevent a checkpoint from happening if the computation creates a new process with higher frequency than the alarm frequency. In that case, we would keep resetting alarm on process creation thus never checkpointing.

Perhaps, a better fix would be to "delay" checkpoint until the computation reaches a "checkpointable" state, i.e., all processes have finished init/exec.

gc00:
@karya0, I also was looking for a better fix to the bug.  My new comment in the description of this PR describes what I think is a better fix.  I'm eager to hear your thoughts in the review.

gc00:
@karya0 , Ping.  Isn't this fix good enough for now?  It's simple, and it solves the basic problems, so that I can return to using GDB.  If you want to improve it later, that's fine.  But why handicap me in my work by making it difficult for me to use GDB elsewhere?

karya0:
> @karya0 , Ping. Isn't this fix good enough for now? It's simple, and it solves the basic problems, so that I can return to using GDB. If you want to improve it later, that's fine. But why handicap me in my work by making it difficult for me to use GDB elsewhere?

Hi @gc00, I will drop the change request for now. But please see my previous comment:

> Resetting alarm here could prevent a checkpoint from happening if the computation creates a new process with higher frequency than the alarm frequency. In that case, we would keep resetting alarm on process creation thus never checkpointing.

This bug must be fixed before we do a release. Can you create an issue targeted to 3.0 to track it?

karya0:
Oh nevermind. Just saw the updated commit. Approving it now.

PR #815, opened 2019-12-27
After system call & reading errno, reset errno to 0

gc00:
```
  **********************************************
  * README:  This PR is extremely important!!! *
  **********************************************
```

This is primarily an enhancement.  Originally, I thought that without this enhancement, it would expose a certain bug.  I was wrong about that.  What I was seeing late at night was not actually a bug -- just a little confusion over the code.

Next, I feel that we have been incredibly sloppy in our use of `errno`.  Most of this is from many years ago, when we are not as careful in our coding.  Most of these others are not actual bugs, but they violate the convention that our utility function should act like a syscall:  either it returns '-1', and sets errno, or it returns something else, and either sets 'errno' to 0, or doesn't change the original value of 'errno'.
  (And I include a minor second commit to correct a comment in src/dmtcp_coordinator.cpp.  It's
    not worth including that as a separate PR.)

We have similar, potential bugs in our code for `Util::readAll()`, `Util::writeAll()`, and in lots of other places in the code that is many years old.  While these are not actual bugs, they make it easy to later create new bugs, because we violate the convention that either `rc == -1 && errno != 0` or `rc >= 0 && errno == 0`.  So, we end up with 'errno' set to a non-zero value in lots of places where there is no actual error.

Please someone!  Check my logic to see that I got most or all of these bugs and that my bug fixes are correct.  For example, where we see 'JASSERT(rc == -1 && errno != 0)', there is no bug since we will exit with an assert error.  But when we see a 'while' loop, we often forgot to reset 'errno' to 0 for the next iteration where we will repeat of the system call.

### DIFF
--- a/src/coordinatorapi.cpp
+++ b/src/coordinatorapi.cpp
@@ -1045,6 +1045,7 @@ waitForCheckpointCommand()
       break;
     }
     JASSERT(errno == EINTR) (JASSERT_ERRNO); /* EINTR: a signal was caught */
+    errno = 0;
     if (ckptInterval > 0) {
       struct timeval end;
       JASSERT(gettimeofday(&end, NULL) == 0) (JASSERT_ERRNO);
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -44,7 +44,7 @@
  *   maximumState for states of all workers, accessed through getStatus()   *
  *   or through minimumState()                                              *
  * The states for a worker (client) are:                                    *
- * Checkpoint: RUNNING -> SUSPENDED -> CHECKPOINTING                        *
+ * Checkpoint: RUNNING -> PRESUSPEND -> SUSPENDED -> CHECKPOINTING          *
  *                     -> (Checkpoint barriers) -> CHECKPOINTED             *
  *                     -> (Resume barriers) -> RUNNING                      *
  *                                                                          *
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -1385,14 +1385,17 @@ DmtcpCoordinator::eventLoop(bool daemon)
   while (true) {
     // Wait until either there is some activity on client sockets, or the timer
     // has expired.
+    // If epoll_wait errors out due to signal, and timer not expired, retry.
     int nfds;
     do {
+      errno = 0;
       nfds = epoll_wait(epollFd, events, MAX_EVENTS, -1);
     } while (nfds < 0 && errno == EINTR && !timerExpired);
 
 
-    // The ckpt timer has expired; it's time to checkpoint.
+    // The ckpt timer expired and interrupted epoll_wait; time to checkpoint.
     if (nfds == -1 && errno == EINTR && timerExpired) {
+      errno = 0;
       timerExpired = false;
       startCheckpoint();
       continue;
@@ -1400,7 +1403,7 @@ DmtcpCoordinator::eventLoop(bool daemon)
 
     // alarm() is not always the only source of interrupts.
     // For example, any signal, including signal 0 or SIGWINCH can cause this.
-    JASSERT(nfds != -1 || errno == EINTR) (JASSERT_ERRNO);
+    JASSERT(nfds != -1 && errno == 0) (JASSERT_ERRNO);
 
     for (int n = 0; n < nfds; ++n) {
       void *ptr = events[n].data.ptr;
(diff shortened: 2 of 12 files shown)


karya0 (changes requested):
Hi Gene, as I wrote in a separate email, we do not need to reset errno to '0' before library calls unless we are calling something that returns -1 as success.

gc00:
@karya0 , The issue is how 'errno' was intended to be designed.  Either:
(a) a function returns -1 (error) and an errno; or
(b) a function returns 0; and doesn't change the original errno.

This is what 'read()' does.  So, this is what 'Util::readAll()' should do.

If it were true that the 'rc' and the 'errno' don't have to correspond, then we could simply write weird wrapper functions around syscalls such that they handle something that DMTCP needs and return 'rc = SUCCESS', but leave 'errno' to an error condition.  Then our wrapper function for the syscall is not following the 'man' page for that syscall.

Our various utilities, like 'Util::readAll()' should be following the same convention as 'read()' itself.  If our 'readAll()' follows the same convention as 'read()', then there's less chance of someone using it wrong.

karya0:
> @karya0 , The issue is how 'errno' was intended to be designed. Either:
> (a) a function returns -1 (error) and an errno; or
> (b) a function returns 0; and doesn't change the original errno.
> 
> This is what 'read()' does. So, this is what 'Util::readAll()' should do.
> 
> If it were true that the 'rc' and the 'errno' don't have to correspond, then we could simply write weird wrapper functions around syscalls such that they handle something that DMTCP needs and return 'rc = SUCCESS', but leave 'errno' to an error condition. Then our wrapper function for the syscall is not following the 'man' page for that syscall.

I agree 100% with you on this. For our wrappers, we must respect the semantics of errno in case of a successful return. Right now we don't do that in several wrappers (we do take care of the failed cases though) and it would be good to fix it.

My earlier comment was about respecting errno for dmtcp-internal calls -- those that have nothing to do with wrappers such as calls originating from coordinator, dmtcp_restart, etc.

gc00:
> My earlier comment was about respecting errno for dmtcp-internal calls -- those that have nothing to do with wrappers such as calls originating from coordinator, dmtcp_restart, etc.

First, I would argue that when two functions are called from DMTCP, any unrelated code in the middle should not see the 'errno' suddenly go from '0' to an unrelated error.

Second, ny DMTCP wrapper functions or utilities that could be executed by a thread of the end user must _not_ change 'errno' from '0' to an unrelated error.  The 'errno' is thread-specific, and we don't want to mess up code of the end user.

Can we enumerate a principle here, and start to whitelist some of the changes and blacklist others, until we converge?  Thanks.

karya0:
@gc00: First off, this is an important topic and I appreciate your continued engagement :-).

> Second, ny DMTCP wrapper functions or utilities that could be executed by a thread of the end user must _not_ change 'errno' from '0' to an unrelated error. The 'errno' is thread-specific, and we don't want to mess up code of the end user.
> 
> Can we enumerate a principle here, and start to whitelist some of the changes and blacklist others, until we converge? Thanks.

I agree with this principle. Whenever user code ends up in a dmtcp wrapper, upon returning from the wrapper, DMTCP shouldn't change `errno`. The underlying function can (and probably will) change `errno` in case of a failure, and we should just return that value as is. However, if the underlying call were to succeed, it won't change `errno`, neither should we. 

> > My earlier comment was about respecting errno for dmtcp-internal calls -- those that have nothing to do with wrappers such as calls originating from coordinator, dmtcp_restart, etc.
> 
> First, I would argue that when two functions are called from DMTCP, any unrelated code in the middle should not see the 'errno' suddenly go from '0' to an unrelated error.

I think I am missing something here. Could you please provide an example scenario (that doesn't involve any user call like above)?

gc00:
@karya0,  I could think about this more, and then decide what I think.

There is a nice discussion of this business of whether a successful syscall (or library call?) is allowed to set errno to non-zero anywhere.  The discussion is here:
  https://softwareengineering.stackexchange.com/questions/209729/rationale-behind-c-library-functions-never-setting-errno-to-zero

They argue there that ANSI has revised the recommended use of errno to say that a successful call can still set errno to non-zero anyway.  That's a point in your favor.

They also give the small counter-example of 'man strtod'.  In that man page, they insist on setting errno to 0 before the call to strtod.  (But people forget to do things like that.  So, I felt it was a little safer to not gratuitously sett errno to non-zero over _many_ functions in DMTCP, and hope that the user realizes that the default value of errno could actually be non-zero (on success).)

I'm open to further comments or recommendations.

PR #816, opened 2019-12-29
IPC Cleanup to prepare for followup PRs.

karya0:
This is a minor cleanup that mostly involves code movement without any functional changes.

### DIFF
--- a/src/plugin/ipc/event/eventconnlist.cpp
+++ b/src/plugin/ipc/event/eventconnlist.cpp
@@ -46,6 +46,23 @@ dmtcp_EventConnList_EventHook(DmtcpEvent_t event, DmtcpEventData_t *data)
   }
 }
 
+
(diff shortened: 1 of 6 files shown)


karya0:
@gc00: The single function was just a shortcut at the time we created the IPC plugin. This was because these plugins shared a bunch of code and refactoring it at the time seemed extra work.

Since IPC plugin exists only as a wrapper around the underlying plugins (file, socket, etc.), this PR is trying to get to the point where all plugin-specific code is in their corresponding directories/files. If we had separate .so files, this is how it would have looked anyways.

Having said that, I am working on a few PRs that disentangle file and socket plugin further so that we can make file, pty, etc., more maintainable. The current design/implementation is socket-centric-- drain/refill/rewire, etc., and that makes it harder to update file/pty/event plugins.

PR #817, opened 2019-12-29
Added WrapperLock class for efficient lock mgmt.

karya0:
This PR adds a new WrapperLock class that makes lock acquisition in complex wrappers more manageable. This doesn't involve any functional changes yet.

### DIFF
--- a/src/popen.cpp
+++ b/src/popen.cpp
@@ -79,7 +79,7 @@ FILE * popen(const char *command, const char *mode)
   }
 
   {
-    WrapperLock wrapperLock;
+    WrapperLock disableCheckpoint;
     if (pipe(pipe_fds) < 0) {
       return NULL;
     }
@@ -136,7 +136,7 @@ FILE * popen(const char *command, const char *mode)
   }
 
   {
-    WrapperLock wrapperLock;
+    WrapperLock disableCheckpoint;
 
     fp = fdopen(parent_fd, new_mode);
     if (!do_cloexec) {
--- a/src/popen.cpp
+++ b/src/popen.cpp
@@ -79,7 +79,7 @@ FILE * popen(const char *command, const char *mode)
   }
 
   {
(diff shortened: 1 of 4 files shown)


karya0:
Hi @gc00, I didn't remove ENABLE/DISABLE_CKPT. It was only removed from popen wrapper since it's a simple one and showcases how to use this class. I have some other incoming changes that depend upon WrapperLock class so I wanted to have this PR done separately :).

src/popen.cpp
```diff
@@ -79,7 +79,7 @@ FILE * popen(const char *command, const char *mode)
   }
 
   {
-    WRAPPER_EXECUTION_DISABLE_CKPT();
+    WrapperLock wrapperLock;
```
gc00: Thanks for the extra comment.  As I understand it, DISABLE/ENABLE_CKPT() is replaced by a local variable, 'wrapperLock', with a constructor and an implicit destructor when the local variable goes out of scope.

Can we think of a more meaningful variable name, or else a standard comment to always attach to this usage?  In particular, we've lost the word 'CKPT'.  So, a casual reader will understand only that we're acquiring a lock, and will not understand that we're disabling checkpoints.  Here is one possible pattern, but I'm open to other patterns that will be meaningful to those not used to C++.

```
// C++: local variable implicitly calls lock constructor, and later implicitly calls destructor when the variable is out of scope.
WrapperLock lockToDisableCheckpoints;
```

I'm absolutely sure that many people who aren't used to the niceties of C++ (e.g., new Ph.D. students) will otherwise be totally confused by this implicit use of a constructor/destructor.

gc00:
See my inline comment about the replacement for 'WRAPPER_EXECUTION_DISABLE_CKPT();'.

src/popen.cpp
karya0: Renamed to `disableCheckpoint` since "lock" is implicit in the type.

PR #820, opened 2019-12-31
Removed unnecessary barriers

karya0:
This PR removes redundant/unnecessary barriers that existed in-between plugins. Any plugin that requires a barrier should explicitly request one via `dmtcp_global_barrier` api.


src/dmtcp_coordinator.cpp
```diff
@@ -473,7 +473,7 @@ DmtcpCoordinator::processBarrier(const string &barrier)
   if (currentBarrier.empty()) {
     currentBarrier = barrier;
   } else {
-    JASSERT(barrier == currentBarrier);
+    JASSERT(barrier == currentBarrier) (barrier) (currentBarrier);
```
gc00: Just below this, you have:
```
++workersAtCurrentBarrier;
releaseBarrier(barrier);
```

I'm requesting a name change:  'releaseBarrier' -> 'releaseBarrierIfAllWorkersReady'
In fact, this name change stems from an earlier commit:
> commit 09e15ca53cfaea42cf36feb19a9e0112cb6220ae
Author: Kapil Arya <kapil.arya.17@gmail.com>
Date:   Mon Sep 9 14:19:58 2019 -0700
    Newer barrier API implementation in Coordinator.

I should have noticed at that time.  Sorry.
The name change is requested because the code used to say '--workersAtCurrentBarrier;' and one could easily imagine that 'releaseBarrier' was waiting for a countdown to zero before releasing.  That was meaningful.  But with commit 09e15ca53cfaea42cf36feb19a9e0112cb6220ae , it seemed like we would increment the worker count forever and maybe do some bookkeeping about releasing internal to 'releaseBarriers'.   That is inaccurate.  The name change to 'releaseBarrierIfAllWorkersReady' makes clear that there is no internal bookkeeping.  If all workers are ready, we rel3ease.  If some workers are not release, then this function is _side-effect free_.  The more meaningful name tells us that the function will be side-effect free or it will release, and that makes reasoning about the logic _much_ easier.

karya0:
> 1. I've added an inline comment requesting that 'releaseBarrier' be renamed to 'releaseBarrierIfAllWorkersReady' everywhere in dmtcp_coordinator.cpp.  Naturally, I'm open to other names that you may prefer.  But let's please modify the original name 'releaseBarrier' to make it clear that: (a) it doesn't always release; and (b) if it doesn't release, then it is side-effect free.

That's a good idea. I've been meaning to revamp some of that barrier code so if I don't get to it in the next few days, I'll create a separate commit to rename. Since this is not directly related to changes in this PR, I'll merge it as is.

> 2. Second, I'm including, below, in this review a list of all barriers encountered in checkpointing 'dmtcp1'  Compared to 2.6, we now have about twice as many stages due to separate pipelines for Event/File and Socket/SVIPC.  I think it's fine to live with that for now, but I'd like to document it.  @karya0 , would you like me to create a github issue for this?

It's always a good idea to have more documentation, so yeah, please create an issue. 

Further, I don't think the number of barriers is large enough to cause significant performance issues. The previous design interleaved plugin barriers, and that's why had fewer of them.

Having said that, I do want to create a `dmtcp_local_barrier()` API similar to `dmtcp_global_barrier` but without coordinator involvement. I estimate that would significantly reduce the number of global barriers to less than six, with the rest being demoted to local barriers.

PR #822, opened 2020-01-02
Converted some global barriers into local.

karya0:
There is a minor commit removing some unused functions from the event plugin. Removal of those functions resulted in removal of barriers around those functions.

### DIFF
--- a/src/dmtcpplugin.cpp
+++ b/src/dmtcpplugin.cpp
@@ -519,11 +519,18 @@ dmtcp_no_coordinator(void)
 EXTERNC void
 dmtcp_global_barrier(const char *barrier)
 {
-  JTRACE("Waiting for barrier") (barrier);
+  JTRACE("Waiting for global barrier") (barrier);
   if (!CoordinatorAPI::waitForBarrier(barrier)) {
     JTRACE("Failed to read message from coordinator; process exiting?");
     JASSERT(DmtcpWorker::isExitInProgress());
     DmtcpWorker::ckptThreadPerformExit();
   }
   JTRACE("Barrier Released") (barrier);
+}
+
+EXTERNC void
+dmtcp_local_barrier(const char *barrier)
+{
+  JTRACE("Waiting for local barrier") (barrier);
+  SharedData::waitForBarrier(barrier);
 }
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -222,6 +222,7 @@ int dmtcp_enable_ckpt(void) __attribute__((weak));
 void dmtcp_initialize_plugin(void) __attribute((weak));
 
 void dmtcp_global_barrier(const char *barrier) __attribute((weak));
+void dmtcp_local_barrier(const char *barrier) __attribute((weak));
 
 // See: test/plugin/example-db dir for an example:
 int dmtcp_send_key_val_pair_to_coordinator(const char *id,
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -221,6 +221,15 @@ int dmtcp_enable_ckpt(void) __attribute__((weak));
 
 void dmtcp_initialize_plugin(void) __attribute((weak));
 
+/*
+ * Global barriers are required when a plugin needs inter-node synchronization,
+ * such as using the coordinator name-service database. Currently, only the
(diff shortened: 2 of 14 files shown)


gc00:
@karya0 , I'll do a careful review later.  But right now, I have a question.

If I understand, PRECHECKPOINT will become local (i.e., only communication among the same host is needed).  But couldn't it happen that a user would want to use our nameservice (publish/subscribe) during a PRECHECKPOINT?  (I haven't thought it through myself, but it seemed easy to just ask the question first.)  Thanks.

karya0:
> @karya0, I really like the change. And with enough work, I could independently verify that the events that you choose to convert from global to local are valid.
> 
> But it would help me and future readers of the code a lot, if you could add a short phrase or sentence in a comment above each _local_ barrier, explaining why it's safe to make that barrier local, and not global.
> 
> I suspect that this will even help you six months in the future, if you have to come back to this code, and check that there are no mistakes in what was changed to local.
> 
> I'd like to read the phrases or sentences before approving. We all have blind spots, and it's best if an independent person reads the comments to check for clarity. Thanks.

@gc00: Good point. I did another review and it looks like we only the socket plugin needs a global barrier. All other plugins are dealing with node-local resources (e.g., pids, fds, etc.), and just need node-local barriers. Socket plugin (and IB/RM plugins too) goes beyond a single node and needs to involve coordinator for a global name-service database. A simple thumb rule is to always insert a global-barrier between registering and querying the coordinator name-service database.

I have now added a small comment for the socket plugin. The rest plugins are fine, I think.

src/dmtcpplugin.cpp
```diff
@@ -519,11 +519,18 @@ dmtcp_no_coordinator(void)
 EXTERNC void
```
gc00: Can we add a _verbose_ comment in front of dmtcp_global_barrier giving the advice that you included in the PR discussion?  Specifically, I'm thinking of something like:
> we only the socket plugin needs a global barrier. All other plugins are dealing with node-local resources (e.g., pids, fds, etc.), and just need node-local barriers. Socket plugin (and IB/RM plugins too) goes beyond a single node and needs to involve coordinator for a global name-service database. A simple thumb rule is to always insert a global-barrier between registering and querying the coordinator name-service database.

Could you word-smith according to make it sufficiently precise, but also to clearly give the intuition to users of DMTCP who didn't find this information elsewhere?  Thanks.

src/dmtcpplugin.cpp
```diff
@@ -519,11 +519,18 @@ dmtcp_no_coordinator(void)
...
-  JTRACE("Waiting for barrier") (barrier);
+  JTRACE("Waiting for global barrier") (barrier);
   if (!CoordinatorAPI::waitForBarrier(barrier)) {
     JTRACE("Failed to read message from coordinator; process exiting?");
     JASSERT(DmtcpWorker::isExitInProgress());
     DmtcpWorker::ckptThreadPerformExit();
   }
   JTRACE("Barrier Released") (barrier);
 }
+
```
gc00: And as for 'dmtcp_global_barrier' above, can we do the same for 'dmtcp_local_barrier'?  Can we add a _verbose_ comment in front of dmtcp_global_barrier giving the advice that you included in the PR discussion?  Specifically, I'm thinking of something like:
> we only the socket plugin needs a global barrier. All other plugins are dealing with node-local resources (e.g., pids, fds, etc.), and just need node-local barriers. Socket plugin (and IB/RM plugins too) goes beyond a single node and needs to involve coordinator for a global name-service database. A simple thumb rule is to always insert a global-barrier between registering and querying the coordinator name-service database.

Could you word-smith according to make it sufficiently precise, but also to clearly give the intuition to users of DMTCP who didn't find this information elsewhere?  Thanks.

karya0:
@gc00: Can you please approve the PR if there are no more comments to address? :)

karya0:
@gc00: ping :)

gc00:
@karya0 , Sorry for my delay in getting back to you.  It's been a crazy month.

I had suggested a somewhat verbose comment in include/dmtcp.h (where users will see it) about  dmtcp_global_barriers.  I have a comment for `include/dmtcp.h` about this.  Could you write something there also?  It should be quick, and I promise to approve the PR then.

karya0:

PR #823, opened 2020-01-02
3.0 fix restart mtcp header

gc00:
Two commits (This is now final.  Ready for review.):

1. MtcpHeader: size_t->uint64_t: for 32-/64-bit mode
    This is important for mixed 32-64-bit computations.  mtcp/mtcp_header.h is shared between dmtcp_restart.cpp (64-bit process) and mtcp_restart_m32 (32-bit process).  So, the fields in the struct must be 64-bit aligned.  In 32-bit Linux, 'size_t' and 'ssize_t' are 4 bytes.  There are changed to 8 bytes here to be 64-bit aligned.  The bug would be seen on restart when doing:
`./configure --enable-multilib && make -j check-dmtcp1-m32`

2. restore_size->restore_len: like restoreBufLen/mmap
    This doesn't change any functionality.  It's just a variable name change from 'restore_size' to 'restore_len'.  This is in keeping with the existing variable/fnc name 'restoreBufLen()'.  In turn, that is in keeping with 'man mmap', which refers to the second argument as 'length'.

NOTE: I'm seeing a compiler warning about comparison of signed and unsigned int expressions.  But it seems to happen _only when compiling in 32-bit mode_, and when it involves Util::readAll() or Util::writeAll().  I have now fixed that in the next PR (PR #824).  So, this PR can still be reviewed independently of the next PR.

### DIFF
--- a/src/mtcp/mtcp_header.h
+++ b/src/mtcp/mtcp_header.h
@@ -1,6 +1,7 @@
 #ifndef MTCP_HEADER_H
 #define MTCP_HEADER_H
 
+#include <stdint.h>  // For 'uint64_t'
 #include "ldt.h"
 
 #ifdef __i386__
@@ -45,7 +46,8 @@ typedef union _MtcpHeader {
     void *saved_brk;
     void *end_of_stack;
     void *restore_addr;
-    size_t restore_size;
+    // uint64_t processinfo.h:_restoreBufLen; 64-bit aligned for 32-64-bit
+    uint64_t restore_size; /* must be 64-bit aligned */
     void *vdsoStart;
     void *vdsoEnd;
     void *vvarStart;
--- a/src/mtcp/mtcp_header.h
+++ b/src/mtcp/mtcp_header.h
@@ -47,7 +47,7 @@ typedef union _MtcpHeader {
     void *end_of_stack;
     void *restore_addr;
     // uint64_t processinfo.h:_restoreBufLen; 64-bit aligned for 32-64-bit
-    uint64_t restore_size; /* must be 64-bit aligned */
+    uint64_t restore_len; /* must be 64-bit aligned */
     void *vdsoStart;
     void *vdsoEnd;
     void *vvarStart;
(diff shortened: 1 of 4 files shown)


karya0:
Hi @gc00, I am a little confused here. The MTCP header is not used by dmtcp_restart. It only comes into play once we exec into `mtcp_restart` or `mtcp_restart_32`. Am I missing something here?

gc00:
@karya0, You may be right about mtcp_header.h.  Indeed, it is not used by dmtcp_restart.cpp.  It's just that as I was testing './configure --enable-m32', I noticed a bug, and it vaguely seemed that after I made the change to mtcp_header.h, the bug seemed to go away.  If this is of concern, let me know, and I'll go back and set up that environment one more time, and if the problem is there, I can more precisely describe the causal chain.

It also seemed to me to be generally a good idea for mtcp_header.h to have a 64-bit-safe header.  It allows general tools like util/readdmtcp.sh to read the header without worrying about 32- vs. 64-bit quantities.  But let me know if you want me to confirm a specific bug.

karya0:
> @karya0, You may be right about mtcp_header.h. Indeed, it is not used by dmtcp_restart.cpp. It's just that as I was testing './configure --enable-m32', I noticed a bug, and it vaguely seemed that after I made the change to mtcp_header.h, the bug seemed to go away. If this is of concern, let me know, and I'll go back and set up that environment one more time, and if the problem is there, I can more precisely describe the causal chain.
> 
> It also seemed to me to be generally a good idea for mtcp_header.h to have a 64-bit-safe header. It allows general tools like util/readdmtcp.sh to read the header without worrying about 32- vs. 64-bit quantities. But let me know if you want me to confirm a specific bug.

To making mtcp_header 32-/64-bit safe, we'll have to convert other fields (e.g., `void *`) to `uint64` as well. That I think is a bit of an overkill since we'll also need to adapt the rest of the code to properly typecast between pointer types and unsigned int types.

karya0:
Closing as stale. Please reopen if still relevant.

PR #825, opened 2020-01-17
protected-fds: Add ability to set protected fd base at runtime

gc00:
_[ This PR is interacting badly with test/rlimit-*.c.  Most likely the issue is that in 3.0/master some
  variables are now initialized globally, whereas in DMTCP-2.6, they were members of a class
  that were initialized after main() was called by the init method.  This was the case with the first
  of the two commits here.  I'll debug test/rlimit-* later.  But those tests are not currently passing. ]_

PR #468 created this commit in DMTCP-2.5.0.  At the time, @karya0 and @gc00 believed there was a better approach, to be implemented for 3.0.  But it was never implemented, and we hit the same bug again now in 3.0/master.  So, this ports the 2.x branch commit to 3.0/master for common code. 

This creates an internal environment variable "DMTCP_PROTECTED_FD_BASE" in DMTCP.  It chooses the maximum fd's for our protected fd's.  It is reset on launch, exec, restart, etc.  This is important on restart, since we can move to a differently configured host.  (QUESTION:  Does it also detect the maximum fd after restart, and set things like coordinatorapi.cpp:coordinatorSocket correctly if we're on a new host with a different maximum fd?)

If we decide later to implement the better approach, the improvements can be made next to both branches.  In particular, if some application also wants to use our protected fds, via dup2(), then they will overwrite it.  In the next PR, we will have dup2() check if the destination is a protected fd, and JASSERT in that case.

For now, let's make sure that "the perfect is not the enemy of the good".  This commit has been valid (no bugs reported) for more than two years in the 2.6 branch.   It's needed right now, to avoid a bug seen by a "customer" using 3.0/master.

### DIFF
--- a/include/protectedfds.h
+++ b/include/protectedfds.h
@@ -22,30 +22,71 @@
 #ifndef DMTCPPROTECTEDFDS_H
 #define DMTCPPROTECTEDFDS_H
 
+#include <stdint.h>
+#include <stdlib.h>
+
+#define DEFAULT_PROTECTED_FD_BASE   820
+
+uint32_t protectedFdBase();
+
+inline uint32_t protectedFdBase()
+{
+  static uint32_t base = DEFAULT_PROTECTED_FD_BASE;
+  const char *buf = getenv("DMTCP_PROTECTED_FD_BASE");
+
+  if (buf) {
+    base = atol(buf);
+  }
+  return base;
+}
+
 enum ProtectedFds {
-  PROTECTED_FD_START = 820,
-  PROTECTED_COORD_FD,
-  PROTECTED_VIRT_COORD_FD,
-  PROTECTED_RESTORE_IP4_SOCK_FD,
-  PROTECTED_RESTORE_IP6_SOCK_FD,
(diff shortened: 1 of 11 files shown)


include/protectedfds.h
```diff
@@ -22,31 +22,72 @@
...
+  SOCKET_FDREWIRER_FD,
+  EVENT_FDREWIRER_FD,
+  READLOG_FD,
+  ENVIRON_FD,
+  NS_FD,
+  DEBUG_SOCKET_FD,
+  FD_END
 };
 
+#define PROTECTED_FD_START               protectedFdBase() + FD_START
```
karya0: Please add a parenthesis around the substitution.

karya0 (approved):

gc00:
This PR is waiting for PR #853 to be pushed in.  That should then fix the bug that is exposed in this PR.  After that, it will pass the Travis test.

include/protectedfds.h
gc00: Done.

PR #828, opened 2020-01-19
Disable ckpt while processing dmtcp_dlsym.

karya0:
Other plugins might call dmtcp_dlsym without holding ckpt-lock and can introduce a deadlock if the ckpt-thread suspends a user-thread while it is holding an internal dlsym lock. Later on, ckpt-thread will deadlock if it also tries to call dmtcp_dlsym.

### DIFF
--- a/src/dmtcp_dlsym.cpp
+++ b/src/dmtcp_dlsym.cpp
@@ -545,6 +545,13 @@ print_debug_messages(dt_tag tags,
 EXTERNC void *
 dmtcp_dlsym(void *handle, const char *symbol)
 {
+  // Acquire checkpoint lock. We need to release it before returning it from the
+  // function. Ideally, we would use the WrapperLock class to avoid calling
+  // dmtcp_enable_ckpt multiple times within the same function but that would
+  // require removing this file's dependency from dmtcp_restart.cpp and
+  // adjusting Makefile.am.
+  dmtcp_disable_ckpt();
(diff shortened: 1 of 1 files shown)


gc00:
I'm continuing this from part of the discussion in PR #829 

@gc00 wrote:
1.   The commit 'Disable ckpt while process dmtcp_dlsym' does dmtcp_disable_ckpt(); inside print_debug_messages(). But it does dmtcp_enable_ckpt(); ouside and after print_debug_messages().  That seems weird. We're violating a software principle of nesting. Is there a way to do this differently? If no, can you put up a big comment there, since you'ree violating a standard convention of software design.

@karya0 wrote:
We need to release the lock before the function returns and since it happens twice, we have to put an unlock call at two places. I agree with you in that ideally, we'll just have one place to unlock, but that would require restructuring the function uncleanly. I plan to replace the lock/unlock pair with the WrapperLock (https://github.com/dmtcp/dmtcp/blob/master/src/threadsync.h#L95) object but it would require more work.

From @gc00:
That sounds reasonable.  Could you please add a comment saying something like this?  It's okay to break the rules of style.  But it's important to add a comment to the reader about why you're breaking the rules of style.  Otherwise, we lose context, and code becomes old an crufty. Thanks.
Please add a comment, and I promise to accept the PR then.

karya0:

PR #829, opened 2020-01-19
Simplified file wrappers.

karya0:
This consolidates file-wrappers in DMTCP core and create new events for open, close, dup, etc. This way, any plugin doesn't need to implement dozens of wrappers for capturing open/close.

Merged wrappers weren't possible in 2.x branch due to improper plugin layering.
    
A similar improvement is needed for fork/exec/popen/system wrappers. We can then use our equivalent of atfork events more extensively.


karya0:
> I have some smaller comments:

Hi @gc00, I have inlined my responses.

> 1. The commit 'Disable ckpt while process dmtcp_dlsym' does `dmtcp_disable_ckpt();` inside `print_debug_messages()`.  But it does  `dmtcp_enable_ckpt();` ouside and after `print_debug_messages()`.  That seems weird.  We're violating a software principle of nesting.  Is there a way to do this differently?  If no, can you put up a big comment there, since you'ree violating a standard convention of software design.

We need to release the lock before the function returns and since it happens twice, we have to put an unlock call at two places. I agree with you in that ideally, we'll just have one place to unlock, but that would require restructuring the function uncleanly. I plan to replace the lock/unlock pair with the WrapperLock (https://github.com/dmtcp/dmtcp/blob/master/src/threadsync.h#L95) object but it would require more work.

> 2. In 'Simplified file-wrappers.', you write: "A similar improvement is needed for fork+exec wrappers."  Can you please change that to "... fork+exec+system+popen wrappers.".   We came across this issue recently with a major user.  They call 'popen' a lot for some short-lived processes (i.e., simple shell commands).  This creates huge runtime overhead when running in DMTCP.  We're creating a new plugin to fix this:  In the LD_PRELOAD environment variable, set the first char after '=' to '\0' before popen/system and set it back after the call.  If you add events for fork/exec, please also include system/popen.  In this use case, it's too late to change LD_PRELOAD within 'exec' if we discover 'exec' is calling a short shell command.  The 'fork' has already created high overhead.  We need to do this at the time of system/popen.

The vfork PR (https://github.com/dmtcp/dmtcp/pull/831) contains the fork/exec events. (It's not ready for review right now, but will be soon, once we merge the other PRs and Cadence can verify that it works). They don't disrupt the existing wrapper for pope/systems or other metacalls that call fork/exec underneath and thus we don't have to create a popen/system dependency right now. We should do that as a separate PR if needed.

> 3. It seems like this PR includes PR #828.  Is that correct?  If so, I'll only review the current PR.

Yes, butlease review and approve #828 as well. I can then rebase and merge this PR. That we can have clean history.

gc00:
> 
> 
> > I have some smaller comments:

(And I've inliined my responses.)
> 
> Hi @gc00, I have inlined my responses.
> 
> > 1. The commit 'Disable ckpt while process dmtcp_dlsym' does `dmtcp_disable_ckpt();` inside `print_debug_messages()`.  But it does  `dmtcp_enable_ckpt();` ouside and after `print_debug_messages()`.  That seems weird.  We're violating a software principle of nesting.  Is there a way to do this differently?  If no, can you put up a big comment there, since you'ree violating a standard convention of software design.
> 
> We need to release the lock before the function returns and since it happens twice, we have to put an unlock call at two places. I agree with you in that ideally, we'll just have one place to unlock, but that would require restructuring the function uncleanly. I plan to replace the lock/unlock pair with the WrapperLock (https://github.com/dmtcp/dmtcp/blob/master/src/threadsync.h#L95) object but it would require more work.
> 
It sounds fine.  But please include a comment.  (And I've shifted this discussion to PR #828.  When it's settled (just a comment --- easy!), then I'll approve it.)  :-)

> > 2. In 'Simplified file-wrappers.', you write: "A similar improvement is needed for fork+exec wrappers."  Can you please change that to "... fork+exec+system+popen wrappers.".   We came across this issue recently with a major user.  They call 'popen' a lot for some short-lived processes (i.e., simple shell commands).  This creates huge runtime overhead when running in DMTCP.  We're creating a new plugin to fix this:  In the LD_PRELOAD environment variable, set the first char after '=' to '\0' before popen/system and set it back after the call.  If you add events for fork/exec, please also include system/popen.  In this use case, it's too late to change LD_PRELOAD within 'exec' if we discover 'exec' is calling a short shell command.  The 'fork' has already created high overhead.  We need to do this at the time of system/popen.
> 
> The vfork PR (#831) contains the fork/exec events. (It's not ready for review right now, but will be soon, once we merge the other PRs and Cadence can verify that it works). They don't disrupt the existing wrapper for pope/systems or other metacalls that call fork/exec underneath and thus we don't have to create a popen/system dependency right now. We should do that as a separate PR if needed.

I guarantee that we will need to handle popen/system.  I'm seeing that now with a user.  I just want either a shell event that does nothing, or at least a comment to remind us that these are important.  As I explained, for short-lived processes using popen/system, it's too late to catch it only when seeing an exec, and to make only the exec a short-lived process.  We need to make the reader of the code aware of this issue.  If you'd like me to write the comment and have you copy it in, let me know, and I'll do that.
> 
> > 3. It seems like this PR includes PR #828.  Is that correct?  If so, I'll only review the current PR.
> 
> Yes, but please review and approve #828 as well. I can then rebase and merge this PR. That we can have clean history.

Got it.  I've started the review on PR #828 there.  There also, I'm just asking for a comment in the code, and then I promise to approve it.  :-)

karya0:
> > The vfork PR (#831) contains the fork/exec events. (It's not ready for review right now, but will be soon, once we merge the other PRs and Cadence can verify that it works). They don't disrupt the existing wrapper for pope/systems or other metacalls that call fork/exec underneath and thus we don't have to create a popen/system dependency right now. We should do that as a separate PR if needed.
> 
> I guarantee that we will need to handle popen/system. I'm seeing that now with a user. I just want either a shell event that does nothing, or at least a comment to remind us that these are important. As I explained, for short-lived processes using popen/system, it's too late to catch it only when seeing an exec, and to make only the exec a short-lived process. We need to make the reader of the code aware of this issue. If you'd like me to write the comment and have you copy it in, let me know, and I'll do that.

I am not sure if I understand. I am not saying that we won't need improved popen/system wrappers and/or their entire handling. I am just arguing that the popen/system wrappers can be handled as a separate _after_ the vfork PR. In any case, this current PR has nothing to do with fork/exec/system/popen except the description which is just meant as future work :-).

karya0:

gc00:
Actually, all I wanted is that the commit message should note:  fork/exec/system/popen is for future work.  Right now in "Simplified file wrappers", the commit message says only that "fork+exec wrappers" is future work.
_I'll approve it now anyway.  But could you just modify that commit message to become:_
"fork+exec+system+popen wrappers".
Right now, our handling of 'popen' is not good at all.  I don't want these things falling through the cracks.

**PREVIOUS COMMIT MSG FOR "Simplified file wrappers" (PLEASE UPDATE):**

> A similar improvement is needed for fork+exec wrappers. We can then
> use our equivalent of atfork events more extensively.

PR #833, opened 2020-03-02
Fix 'dmtcp_command -l' and show port w/ 'ps', 'top'

gc00:
Two commits:
  1.  `dmtcp_command -l`     now works.  This bug has been in DMTCP forever.
  2.  `top` and `ps` should now show the port number for dmtcp_coordinator.


elIdrissiYounes (approved):
LGTM!

PR #831, opened 2020-02-13
Vfork Support

karya0:
Vfork support is implemented by suspending all user threads until the vfork call has returned in the parent. This would happen after the child has either exited or exec'd into a new program.

Each plugin should also use the following vfork-related events:
* DMTCP_EVENT_VFORK_PREPARE: Save a copy of any local data that might get overwritten by subsequent syscalls from the child.
* DMTCP_EVENT_VFORK_PARENT/FAILED: Restore local data saved during the PREPARE phase.
* DMTCP_EVENT_VFORK_CHILD: To perform any child-specific operations (e.g., setup locks, etc.).

PR #835, opened 2020-03-05
Misc changes to prepare for the vfork PR

karya0:
This PR contains several unrelated commits that would be used by the vfork PR #831. I decided to bundle them together into this single PR because most of them are not large enough to justify creating several PRs.


karya0:
Hi @gc00,

TLDR: In principle, I agree with your philosophy of keeping things simpler so as to maintain a relatively low barrier for newcomers. However, I think you have misread/misinterpreted most of the context in this particular PR. I probably should have been more verbose in the commit logs but anyways.

I'm addressing some of your concerns inline. They are meant to be constructive, please interpret them as such :-).

> @karya0 , I really dislike most of these "cleanups". The problem here is that the cleanups are proposing that DMTCP will stop emphasizing the large intersection of the C++, C, and Java languages. Instead, these changes would require a deeper knowledge of C++. So, for example, think back to students coming in as PhD students from, for example, your own undergraduate school. Would they really be ready to start contributing to DMTCP in their first year?

I think we both agree on this already. We don't want any C++ feature that would be pain to read and/or maintain. I, in particular, don't care much about compatibility with Java. Instead, I want to keep it as close to C as possible mostly for performance reasons (with some help from STL for ease of coding).
 
> For example, using the 'auto' keyword means that the type of a variable can no longer be figured out by reading the local code. Things like STL used to be reasonably intuitive. But now, you wish to have students study override, auto, how the scope works in C++, and other such things. The net effect is that we'll have to abandon the support of DMTCP by PhD students, until they have reached a more advanced stage, and have taken a course on C++ and and principles of software development.

Re. auto: I mostly agree with your philosophy but this is where I think you have misinterpreted the changes. The common wisdom around using auto is that it shouldn't be used as a lazy mechanism to avoid writing full types. Instead, it should be used where the type is "obvious" and wouldn't need digging through headers to find the result. One of the biggest usecases is STL iterators during range loops. For example, instead of:
```
for (map<int64_t, map<int64_t, jalib::JBuffer> >::iterator i = myMap.begin(); i != myMap.end(); i++) 
```
one can use:
```
for (auto i = myMap.begin(); i != myMap.end(); i++)
```
The latter is easier to read and intuitive. An improvement on top of this is something like:
```
for (auto keyValue : myMap)
```
I can understand if this is somewhat difficult for new students who are not familiar with C++ but we can discuss the exact semantics.

Re: override: This is something I feel strongly in favor of. Not having override has costed me several days of debugging effort. The most recent one took over 20 hours to figure out that we accidentally changed the signature of a base class function slightly without updating "all" the signatures in the derived class. As a results, instead of calling the virtual function from derived class, it silently called the base class function. Such bugs are super hard to chase and the override keyword catches them at compile time without the need for a careful code review. 
 
> I really think this is a bad idea for us to go elitist, and make the code unreadable for a large number of contributors. What do we gain from this? Well, we can go to a .h file and change a variable type, and the compiler will stop notifying us of where variables of that type are used, and the effects on other parts of the code. For those who have a good overview of all of DMTCP, this is not a problem. For those who don't have a good overview of all of DMTCP, the compielr will silently change the implied type in many places in DMTPC, and a newer PhD student will shrug, and assume that if that changes the behavior in some far-off corner case, then the guru who wrote this must know what they're doing, and the student has no need to worry about correctness anymore.

I totally agree that we shouldn't use "auto" in .h files or other places where it could have undesired side-effects, but I would strongly argue for using it in loops where really long type names are painful to read and write.

gc00:
@karya0 , I agree that I want to be constructive.  But just as it has cost you many hours to debug without the use of 'override', it has cost me many hours to decipher what a previous 'auto ' was
added by a student.  (I think you know how students would react.  `auto` is a cool feature that's
being used all over DTMCP.  So, they will want to use `auto` to eliminate as many of their
own type declarations as possible, because it's so cool that the compiler can figure out
the information that the student is hiding form the reader.

Or to take another example, it cost me some hours to figure out that we had replaced 
`Util::dupFds(tempfd, _fds);` in DMTCP-2.6.0 by `restoreDupFds(tempfd);` in DMTCP-3.0.  I eventually discovered that `restoreDupFds()` referred to a hidden instance variable, `_fds`,
and that the intention of `restoreDupFds(tempfd);` was to _close_ `tempfd`, but only after it
had been dup'ed to a hidden variable, `_fds`, that required me to read the surrounding
code to understand what it was for.  If I'm having trouble with these examples of information hiding,
I can only imagine what newcomers think of it.  In systems, when dealing with low-level
objects, hiding information from the reader is a very bad idea.  This is not a case of creating
a high-level abstraction.  It is a case of hiding very many low-level details that are required
for understanding how DMTCP works.

There are too many examples of this information hiding.  Another example is:
    `return (size_t)id > (size_t)getpid()` gets changed to `return (size_t)id > (size_t)_base`
  with no local comment about what `_base` refers to in `virtualidtable.h`.  So, a newcomer needs to
understand the use of `base` in the class `VirtualIdTable`, even though `base` has no comment
about what it's for.  Only after the casual user has deciphered this, can that user then understand
what the `isIdCreatedByCurrentProcess(IdType id)` method is doing in some unrelated part
of the code, elsewhere.

I am now spending very many hours working on parts of DMTCP, where I'm worried that
new students can't understand it.  I don't have the time to continue to do that.
I'm already seeing the effect of this with the MANA split process addition by @rohgarg .
A lot of people want to use it.  But it seems almost impossible to merge it into
DMTCP-3.0, because DMTPC-3.0 is becoming esoteric faster than @rohgarg or anyone else can merge their code.

If we diverge much more from something that new
students can understand, then I would be forced to fork DMTCP into something that
new students can understand.

To take your example of 'auto', you're arguing that a newcomer should not be allowed to see:
`for (map<int64_t, map<int64_t, jalib::JBuffer> >::iterator i = myMap.begin(); i != myMap.end(); i++) '

If we can hide the type of `myMap` from the newcomer, then you're arguing that when they want to create a map involving `JBuffer`, they will just use `for (auto i = myMap.begin(); i != myMap.end(); i++)`, and even though they are still hazy on how C++ works, somehow they will have a clear and precise understanding of the internal data structures for JBuffer that they have just created.
Of course, GDB will refuse to show them this internal data structure.  And there are techniques to
force the C++ compiler to display the hidden type by creating an error of incomplete type,
but C++ will display an incredibly low-level representation requiring more than a screen of text.
And it's then totally impractical for a user to replace `auto` by this screenful of text.

I might find a way to agree with you about 'override', but only if we add a comment about what
is being overridden.

But for me, the idea of hiding the type of a variable by using `auto` is a non-starter.  I absolutely
don't believe that a newcomer will be able to explain to you or me what is the hidden datatype that
the `auto` keyword is referring to.  And I refuse to double or triple my own time in DMTCP
because newcomers cannot read the `auto` code.  That's why I would consider forking
DMTCP before I try to teach each new student how to go from `auto` to the low-level
data structures and system concepts that DMTCP is intended to work with.

karya0:
@gc00: Ok how about I remove auto but keep virtual override in?

PR #837, opened 2020-03-07
Added dmtcp_get_libc_addr(libc_fnc)

gc00:
Example:
```
    static typeof(getpid)* libc_getpid_ptr = dmtcp_get_libc_addr("getpid");
    int pid = (*libc_getpid_ptr)();
```

This calls the actual 'getpid()' in libc to get the real address.  It's available to plugins through `include/dmtcp.h`.  It's also useful for DMTCP internals when we want to call the actual libc function, but we are forced to use things like `_real_pid()` and hope that it lands inside libc.

For more interesting examples of its use, see the comment about it in `include/dmtcp.h`.
I am also adding the example code below as a typical use case.  This code doing 1000 forks
used to take about a second for a fork.  Now, it takes about a millisecond.  Several customers
need this fix for the performance bug.

```
// Be sure to compile this with  -fPIC, or the call to dmtcp_get_libc_addr() will always report
//    DMTCP_IS_NOT_PRESENT (value 3).
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <time.h>
#include "dmtcp.h"

int main() {
  // typeof(fork) *libc_fork_ptr = dmtcp_get_libc_addr("fork");
  int i;
  const struct timespec delay = {0, 1000000};
  for (i = 0; i < 1000; i++) {
    // int childpid = (*libc_fork_ptr)();
    int childpid = fork();
    if (childpid) { //if parent
      waitpid(childpid, NULL, 0);
      nanosleep(&delay, NULL);
    } else {
      // printf("child pid: %d\n", (int)getpid());
      return 0;
    }
  }
  return 0;
}

// This function is placed inside the plugin as a wrapper for fork()
/*
 * pid_t fork() {
 *   static typeof(fork) *libc_fork_ptr = NULL;
 *   if (libc_fork_ptr == NULL) {libc_fork_ptr = dmtcp_get_libc_addr("fork");}
 *   // Skip DMTCP interposition on fork; child doesn't coonect to coord.
 *   (*libc_fork_ptr)();
 * }
 */
```


include/dmtcp.h
```diff
@@ -262,6 +262,34 @@ int dmtcp_enable_ckpt(void) __attribute__((weak));
 #define dmtcp_enable_ckpt() \
   (dmtcp_enable_ckpt ? dmtcp_enable_ckpt() : DMTCP_NOT_PRESENT)
 
+/**
+ * Example:  dmtcp_get_libc_addr("fork") returns the address of "fork"
+ *           in libc in memory.
```
JainTwinkle: in libc in memory? Were you intending to write that it returns the address of "fork" in libc loaded in memory?

JainTwinkle:
@gc00  DMTCP has multiple wrappers of a function in different DMTCP libraries. `_real_func` calls redirect the call to the next plugin in line. Is this utility intended to replace the `_real_func` (in future)?  I believe it is not intended to replace real calls but to have another utility to skip redirections within DMTCP. Please correct me if I am wrong.
Please see one small comment of mine in `include/dmtcp.h`. Otherwise LGTM :)

PR #838, opened 2020-03-14
Fix stdin/ctty w/ cat FILE|dmtcp_restart ckpt*dmtp

gc00:
It used to be that if the discovered controlling tty (ctty) and stdin were the same, then it would
close the original stdin, and open a new one for the same tty.  Similarly, if the stdin was reading from a pipe or a socket, then we should not close the stdin and re-open it to /dev/tty.  This fix allows the following to work:

```
# Create a ckpt image of a process waiting to read from stdin:
bin/dmtcp_launch -i5 cat -
# Now, this will print /etc/passwd as if by:  'cat /etc/passwd | cat -'
cat /etc/passwd | bin/dmtcp_restart ckpt_cat_*.dmtcp
```

### DIFF
--- a/src/plugin/ipc/file/ptyconnection.cpp
+++ b/src/plugin/ipc/file/ptyconnection.cpp
@@ -386,6 +386,17 @@ PtyConnection::postRestart()
         controllingTty = jalib::Filesystem::GetControllingTerm(getppid());
       }
       stdinDeviceName = (jalib::Filesystem::GetDeviceName(STDIN_FILENO));
+      if (controllingTty.c_str() == stdinDeviceName ||
+           (Util::strStartsWith(controllingTty.c_str(), "/dev/pts/") &&
+            Util::strStartsWith(stdinDeviceName.c_str(), "pipe:")) ||
+           (Util::strStartsWith(controllingTty.c_str(), "/dev/pts/") &&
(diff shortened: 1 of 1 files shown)


src/plugin/ipc/file/ptyconnection.cpp
```diff
@@ -477,6 +488,10 @@ PtyConnection::postRestart()
     }
   }
 
+  // Apparently, the PtyConnection object has a private vector of _fds
```
karya0: FYI: it's not just PtyConnection, but Connection class itself that contains the _fds vector.

src/plugin/ipc/file/ptyconnection.cpp
```diff
@@ -386,6 +386,17 @@ PtyConnection::postRestart()
         controllingTty = jalib::Filesystem::GetControllingTerm(getppid());
       }
       stdinDeviceName = (jalib::Filesystem::GetDeviceName(STDIN_FILENO));
+      if (controllingTty.c_str() == stdinDeviceName ||
```
karya0, line 389: Did you mean it to be `&&` instead of `||`? Unfortunately, I missed it during the original review :(.

Apparently, this commit breaks tcsh, zsh, etc., because the controlling terminal / stdin/out/err are not restored properly.

PR #840, opened 2020-03-30
Spades plugin

JainTwinkle:
This pull request adds spades plugin and a new launch option --spades to enable it. Spades is being widely used at NERSC. Spades require extra files to be saved than just the opened files at checkpoint time in order to restart it correctly. We call those files as precious files. The precious files can be at multiple locations. This plugin is dedicated to spades because in this use case, we know where to find the precious files. The plugin can be made more general in the future.


JainTwinkle:
This is not ready for review yet. I will ping once it is ready for review.

JainTwinkle:
This PR is specific to spades application. However, in a long run, we'd want to improve DMTCP by adding a general plugin that can save precious files. Precious files are those external files that are not opened at the time of checkpoint by the process but require but need to be saved for a correct restart. This scenario can happen when an application or a process deletes file(s) after taking a checkpoint and need to be recreated for correct restoration.

My notes on a general PR for saving precious files are as follows:
Precious files can exist:
1) per-process
  - each process will save its precious files
2) general for all the process
  - elect a leader to save precious files on behalf of all the processes

Precious files types:
1) can be a directory
2) can be a number of individual files with path

Where to save?
- we will need to save similarly as we save open files.

How to provide precious files as input?
- each file as a command-line option
- a command-line option or an environment variable that points to a file which contains 1) absolute file/directory name 2)type - file or directory

- We will also need to check if the precious files have enough permission for us to save and restore files or not.
- Should we save precious files at the pre-ckpt barrier or resume barrier? does it matter?

- Add support for unique copies of precious files when unique-ckpt is enabled.

PR #851, opened 2020-06-30
jalib: Fix ABA bug in Jalloc implementation

rohgarg:


rohgarg:
@gc00, @karya0: The first commit uses inline assembly to do a 128-bit CAS; the second commit uses GCC built-ins to do a 128-bit CAS.

While the second commit is cleaner, it requires an additional GCC flag (`-mcx16`) to generate the correct CAS instruction (`cmpxchg16b`) and compile. So, I had to do other changes in configure/Makefiles etc.

We can choose whichever one we like: inline assembly (and no additional GCC flags), or no inline assembly (and additional GCC flags).

Also, I haven't been able to really test this code on ARM/AArch64/i386 systems. I have only tested this on x86-64.

rohgarg:
Ping!

gc00 (commented):
Can we change `is_x64_host` to `is_x86_64_host` ?  I'll approve it as soon as that change is there.  Thanks!

I agree that it's fine to push it in, and if there's a problem with the inline assembly for aarch64, etc., we'll discover it quickly enough next time we have access to those machines.

gc00:
@rohgarg, ping.  :-)

As I wrote in my earlier comment:
> Can we change is_x64_host to is_x86_64_host ? I'll approve it as soon as that change is there. Thanks!

elIdrissiYounes (approved):
LGTM!

PR #853, opened 2020-08-04
Fixes related to DMTCP initialization.

karya0:


src/util_init.cpp
```diff
@@ -148,13 +148,6 @@ Util::initializeLogFile(const char *tmpDir,
     a << "\nPrev JAssertLog path: " << prevLogPath;
   }
 
-  a << "\nArgv: ";
-  vector<string>args = jalib::Filesystem::GetProgramArgs();
-  size_t i;
```
gc00, line 153: I think there's a compile-time bug here.  You'll need `size_t i;` in the `for` loop just below this:  `for (i = 0; environ[i] != NULL; i++)  {`.  But this commit deletes the declarationof `i`.

PR #860, opened 2020-08-11
FIXME: Fix unmaintainable jassert code

gc00:
The code described here is unmaintainable for the reasons below.
It uses a C++ trick (function on the left hand side of assignment)
in order to hide a variable that one needs to see in GDB.  It forces
the GDB person to execute user code (which might call malloc())
in order to discover the value of a hidden variable.

* Bad style that can waste hours for those unfamiliar with the code
 * Each of three functions in jassert.cpp has an identical body.
   The meaning of the three functions must be inferred from how
   the function is called elsewhere, since the function definition
   provides no clue as the function meaning.

*** If someone wants to translate this into _readable_ C++ code, that's great.
*** Otherwise, I'll change it into direct, readable C code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ language: cpp
 compiler:
   - gcc
 
-script: ./configure && make -j4 && make tests && ./test/autotest.py --retry-once
+# script: ./configure && make -j4 && make tests && ./test/autotest.py --retry-once
+# Force re-build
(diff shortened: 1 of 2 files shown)


karya0:
This is a good point. JASSERT shouldn't be "discovering" tmpdir/logdir/logfilepath, etc., settings. Instead, these things should be passed through `jalib_init` interface (https://github.com/dmtcp/dmtcp/blob/master/jalib/jalib.h#L89).

PR #864, opened 2020-08-15
Delay *.ckpt.temp to *.ckpt rename until after WriteCkpt barrier.

karya0:
This helps prevent partial checkpoints when a large process takes long time to write a ckpt image.

### DIFF
--- a/src/dmtcpworker.cpp
+++ b/src/dmtcpworker.cpp
@@ -464,16 +464,26 @@ void
 DmtcpWorker::postCheckpoint()
 {
   WorkerState::setCurrentState(WorkerState::CHECKPOINTED);
+
+  // TODO: Merge this barrier with the previous `sendCkptFilename` msg.
+  JTRACE("Waiting for Write-Ckpt barrier");
+  CoordinatorAPI::waitForBarrier("DMT:WriteCkpt");
+
+
+  /* Now that temp checkpoint file is complete, rename it over old permanent
+   * checkpoint file.  Uses rename() syscall, which doesn't change i-nodes.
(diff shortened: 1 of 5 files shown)


gc00:
This guarantees that all processes rename .ckpt.temp to .ckpt at the same time.  So, there is only a short inconsistency while the request to rename is in progress by the several processes, possibly on different nodes.

QUESTION:  Does it also guarantee the right behavior for checkpointing of open files?  If using  `dmtcp_launch --checkpoint-open-files` (or the default behavior of checkpointing open files whose offset is at the end of the file), then will it overwrite any previous checkpoints of the open files, before renaming the .ckpt.temp to .ckpt?  The issue is thata .ckpt.temp file might be gigabytes large, leaving us in an inconsistent state for many minutes.

src/dmtcpworker.cpp
```diff
@@ -464,17 +464,27 @@ void
 DmtcpWorker::postCheckpoint()
 {
   WorkerState::setCurrentState(WorkerState::CHECKPOINTED);
+
+  // TODO: Merge this barrier with the previous `sendCkptFilename` msg.
+  JTRACE("Waiting for Write-Ckpt barrier");
```
gc00, line 469: Please add a comment here.  Perhaps something like:
```
// It's important for all processes to rename .dmtcp.temp to .temp at the same time.
// If one process is gigabytes-large, then without the barrier, it would rename long after
// other renaming has finished, leaving the checkpoints in an inconsistent state for many minutes.
```

gc00:
close and re-open to force a new Travis build.

PR #865, opened 2020-08-15
Fixed ckpt-interval bug.

karya0:
The interval needed to be reset after checkpoint completes.

### DIFF
--- a/src/ckptserializer.cpp
+++ b/src/ckptserializer.cpp
@@ -416,13 +416,10 @@ CkptSerializer::createCkptDir()
 
 // See comments above for open_ckpt_to_read()
 void
-CkptSerializer::writeCkptImage(void *mtcpHdr, size_t mtcpHdrLen)
+CkptSerializer::writeCkptImage(void *mtcpHdr,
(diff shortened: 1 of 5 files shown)


gc00 (commented):
**PR #865 #866 and #867 are all failing on openmp-2 in Travis in the same way.  There is probably a common cause.**

We should dig deeper.  I tested PR #867 on dekaksi (Ubuntu 18.04), and openmp-2 works fine there.

gc00:
@karya0 ,
    I believe that if you rebase and merge, then this PR will now pass.  Apparently, it failed on openmp-2, even though that test has nothing to do with this PR.  I believe that Travis randomly takes extra time occasionally.

If you find that it still doesn't pass, then it's a matter of allowing more time for the openmp-2 test.  The emacs test was on the edge, and I added more time.  Note that there are two knobs you can use:
```
S=10*DEFAULT_S
POST_LAUNCH_SLEEP=3
```
In the case of emacs, I found I had to increase `POST_LAUNCH_SLEEP`.  I suspect that this is also what you need for openmp-2.

PR #866, opened 2020-08-16
util/gdb-dmtcp-utils: Further enhancements

gc00:
Some more enhancements to `util/gdb-dmtcp-utils`.  For example
```
whereis-address $pc
```
now works.  Also, the comments in `util/gdb-dmtcp-utils` referred to its old filename (`util/gdb-add-symbol-files-all`).  This fixes it.

### DIFF
--- a/util/gdb-dmtcp-utils
+++ b/util/gdb-dmtcp-utils
@@ -23,13 +23,13 @@ import textwrap
 
 # This adds a GDB command:  add-symbol-files-all    (no arguments)
 # To use it, either do:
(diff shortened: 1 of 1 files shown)


gc00:
**PR #865 #866 and #867 are all failing on openmp-2 in Travis in the same way.  There is probably a common cause.**

gc00:
The Travis check is now passing, with no essential change in the code.  Probably it was a transient bug in Travis itself.

elIdrissiYounes (approved):
LGTM!

PR #867, opened 2020-08-24
Fix checkpoint-restore of file-backed regions

rohgarg:
The issue is related to the fact that an application is allowed to
mmap a large region of memory even though the entire region may not be
backed by a file. For example, for a file that is just 1 MB on disk,
the application can make an mmap call asking for 2 GB, but only the
first 1 MB of the memory region will be actually made available. There
are two different behaviors of the Linux kernel when one tries to read
anything beyond the first 1 MB:

- If one tries to read any byte in user-land (load instruction) beyond the first
  1 MB, it generates a SIGBUS.

- If one makes a system call such as write() that uses the entire memory
  region (2 GB), the system call returns success, meaning that it wrote
  2 GB, but it actually only writes out 1 MB!

Now, this confuses the restart code: since we recorded that the size
of the region is 2 GB, the code expects that the memory region would be
2 GB, but the kernel only wrote 1 MB. The code then tries to read 2 GB from
the image. There are two possible outcomes:

 - Either the read fails (because of EOF), or

 - The read succeeds (if the image had sufficient data). However, for the
   subsequent memory region, we'll end up seeing garbage data.

Basically, the checkpoint image has all the checkpoint data but it's
not at the offsets that we are expecting.

### DIFF
--- a/include/procmapsarea.h
+++ b/include/procmapsarea.h
@@ -98,6 +98,7 @@ typedef union ProcMapsArea {
       uint64_t __inodenum;
     };
 
+    off_t    mmapFileSize;
     uint64_t properties;
 
     char name[FILENAMESIZE];
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -570,16 +570,14 @@ mtcp_simulateread(int fd, MtcpHeader *mtcpHdr)
     }
     if ((area.properties & DMTCP_ZERO_PAGE) == 0 &&
         (area.properties & DMTCP_SKIP_WRITING_TEXT_SEGMENTS) == 0) {
-      void *addr = mtcp_sys_mmap(0, area.size, PROT_WRITE | PROT_READ,
-                                 MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
-      if (addr == MAP_FAILED) {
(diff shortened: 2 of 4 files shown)


rohgarg:
This is a duplicate of #688 for master.

gc00:
Copying comment from PR#688 to here for quick reference by anybody reviewing this.

LG. It turns out this behavior is in POSIX:
http://pubs.opengroup.org/onlinepubs/009695399/functions/mmap.html
> The mmap() function can be used to map a region of memory that is larger than the current size of the object. Memory access within the mapping but beyond the current end of the underlying objects may result in SIGBUS signals being sent to the process. The reason for this is that the size of the object can be manipulated by other processes and can change at any moment. The implementation should tell the application that a memory reference is outside the object where this can be detected; otherwise, written data may be lost and read data may not reflect actual data in the object.

gc00 (commented):
This is clearly an important PR.  The openmp-2 test is failing on the second checkpoint (after restart).  Can we analyze first why that is failing.  (Presumably, it's related to Ubuntu.)

I tried fetching your branch on dekaksi (Ubuntu 18.04), and the openmp-2 test passes fine there.  I guess we should dig deeper to find the source of this bug.  There are currently some other PRs that are also mysteriously failing in Travis.  Could it be related?

karya0 (approved):

PR #868, opened 2020-08-29
Ensure all restarting processes have connected before releasing a barrier.

karya0:
We removed the numPeers check when we moved to the newer barrier design. This PR reintroduces the concept.


gc00 (approved):
LGTM.  (Sorry for the delay in reviewing this.  This PR is important for Spades and other applications.)

PR #869, opened 2020-09-07
Modify #include declarations to be POSIX compliant

gc00:
Two commits:
  1. include decl's: sys/errno.h -> errno.h  
  2.  Add include files required by POSIX
       (and move stray file `save-symbol-files-to-gdb-script.py` back to util directory where it belongs)

This should be easy to review.  Just cleaning up include declarations


JainTwinkle (approved):
I am observing some already-used or unused header files in DMTCP. I will be creating a separate PR to clean up header files in the next 2-3 days. For now, I am approving this PR so that the new PR can take these newly added header files into account.

PR #872, opened 2020-09-30
Add include files required by POSIX/and gcc warnings (part 2)

gc00:
sys/errno.h -> errno.h
sys/fcntl.h -> fcntl.h

### DIFF
--- a/src/dmtcp_restart.cpp
+++ b/src/dmtcp_restart.cpp
@@ -20,6 +20,7 @@
  ****************************************************************************/
 
 #include <elf.h>
+#include <fcntl.h>
 #include <limits.h>
 #include <stdio.h>
 #include <sys/mman.h>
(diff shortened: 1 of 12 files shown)


rohgarg (approved):
LGTM. Can you please add some context for this change? Is it about some specific GCC version/distro?

gc00:
@rohgarg , I added the changes for POSIX and gcc when I was using gcc-9.0.  I was using gcc-9.0 as provided by Alpine Linux.  Alpine Linux uses musl libc instead of glibc.  mustl libc is a lightweight POSIX-compliant libc.  Alpine libc is often the preferred Linux inside Docker for distributing lightweight services.  I don't yet have DMTCP working with musl libc, but I have a private branch in which I hope to finish off this port to include musl libc.

PR #876, opened 2020-10-10
configure.ac: MPICH_MPICC: Bug fix: c6d50c0ae04

gc00:
This is a bug fix to configure.ac.  The bug was introduced in commit c6d50c0ae04.  It's my fault -- I used AC_PATH_PROG incorrectly.  This is a fix for PR #805.  It should be trivial to review.


elIdrissiYounes (approved):
LGTM!

PR #883, opened 2021-02-03
Bug fix for misidentifying region as stack

mcandress:
In mtcp_restart.c function read_one_memory_areas(): limit using string comparison to identify a stack region to regions whose name is not filename starting with "/".

This fixes a problem with dmtcp_restart failing to mmap a file whose path name includes the substring "stack" because  the MAP_GROWSDOWN flag gets used.


karya0 (approved):
Looks good. Thanks for the fix!

PR #885, opened 2021-02-16
Fix for master branch for glibc 2.33 using stat() family

gc00:
DMTCP fails to compile with glibc-2.33.  In particular, the future release of  Fedora 2.34 (and other distros) will start to use gcc-10.0 and glibc-2.33.

The problem is that in glibc version 2.33, they stop defining `_STAT_VER`.  glibc originally used `_STAT_VER` as the value in an extra argument in `__xstat(vers, path, buf)` .  Presumably, gcc/g++ then macro-expanded `stat(path,buf)` to `__xstat(_STAT_VER,path,buf)`.  But glibc-2.33 now removes `_STAT_VER` and directly defines `stat(path,buf)`, etc.

This PR uses `#ifdef _STAT_VER` to have DMTCP refer either to `__xstat()` or `stat()`, according to whether `_STAT_VER` is defined.

**NOTE:**  mknod() in test/syscall_tester.c also uses _STAT_VER prior to glibc-2.33.  In Fedora rawhide, test/syscall_tester.c fails even without using DMTCP, because mknod can succeed with glibc-2.33, where it would have failed in the past.  I'm not fixing this now.  Let's wait for the release of the distros to see if this problem persists.

Finally, please note the comment in this commit:
```
glibc-2.33: src/plugin/pid/: __xstat->stat, etc.

  **************************************************************************
  * PLEASE READ THE FIXME IN:  src/plugin/pid/pid_syscallsreal.c           *
  * We should remove definitions of _real_*stat*(), since they're not used.*
  **************************************************************************
```

The previous code was the result of commit 3d43ac6e9e0bd8bf596aac9a5d827fbe66d7848a by @karya0.  I think that code is dead code and can be removed.  But I'd rather have @karya0 check me on this, and then eitehr submit his own PR or ask me to do it.

### DIFF
--- a/plugin/pathvirt/pathvirt.cpp
+++ b/plugin/pathvirt/pathvirt.cpp
@@ -23,6 +23,27 @@
 #include <cstring>
 #include <cstdlib>
 
+// glibc version 2.33 and later stopped defining _STAT_VER (for the 'vers'
+// argument of the xstat family of functions), and stopped
+// defining the xstat family.  It now defines the stat family directly.
+// instead of defining stat as a macro that expands to __xstat, etc.
+// We are macro expanding the xstat family to the stat family,
+// whenever _STAT_VER not defined..
+#ifndef _STAT_VER
+# undef __xstat
+# undef __xstat64
+# undef __lxstat
+# undef __lxstat64
+# define __xstat(vers,path,buf)         stat(path,buf)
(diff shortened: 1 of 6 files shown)


elIdrissiYounes (approved):
LGTM!

PR #887, opened 2021-02-20
Add '--parallel' option to autotest.py

gc00:
This now provides:  `./autotest.py --parallel`

The tests are run with up to `PARALLEL_JOBS` jobs (currently 5).  It works by creating `PARALLEL_JOBS` child processes in backgrouind executing `./autotest.py NEXT_TEST`.  If the test fails or has not completed, then the child process is killed and the original process re-executes it in the foreground.

In this PR, `make check` does not yet use the `--parallel` option.  It's best to commit this PR first, and get some confirmation that it is robust.  But that should also work.

### DIFF
--- a/test/autotest.py
+++ b/test/autotest.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+# NOTE: .travis.yml at DMTCP_ROOT calls autotest.py for github/travis
+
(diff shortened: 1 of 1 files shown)


elIdrissiYounes (approved):
LGTM!

PR #889, opened 2021-02-22
Stop using cast for basename (Fix for 97d45952)

gc00:
When libgen.h is included, then glibc uses POSIX `basename()` instead of GNU `basename()`.  The GNU version takes a `const char *` argument.  But the POSIX version takes a `char *` as its argument, since it may choose to destructively alter its string argument.  So, we make a `char *`  local copy of the string to pass to GNU `basename()`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ addons:
     packages:
     - tcsh
     - zsh
-    
+ 
 language: cpp
 compiler:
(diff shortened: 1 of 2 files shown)


elIdrissiYounes (approved):
LGTM!

gc00:
Force Travis to re-build w/ close/re-open.

PR #890, opened 2021-02-22
Add --timeout param to coord; use it in autotest

gc00:
`dmtcp_coordinator.cpp` now takes a `--timeout <seconds>` flag.  If the timeout occurs, then the coordinator exits.

This is useful in `autotest.py`.  This was now changed to call `./bin/dmtcp_coordinator --timeout 10800` (a timeout of 3 hours).  This should get rid of the frequent runaway coordinator processes that occur when `make check` is interrupted.


elIdrissiYounes (approved):
LGTM!

PR #891, opened 2021-02-22
autotest.py: Check for low disk space in /tmp

gc00:
The `file2` test creates several megabytes in `/tmp`.  The test fails when there is not enough disk space in `/tmp`.  However, `autotest.py` simply shows that `file2` is failing, with no clues that the problem is disk space.  This can be a problem when many debugging sessions cause large DMTCP-related files and directories to accumulate in `/tmp`.

This PR now tests for this condition, and prints out a warning just before the tests execute.


elIdrissiYounes (approved):
LGTM!

PR #892, opened 2021-02-22
autotest.py:  Fix params for script, bash, etc.

gc00:
In the script test and the shell tests (python, perl, bash, dash, tcsh, zsh), `POST_LAUNCH_SLEEP` is set to 2 seconds before the test.  This is needed because the shells need to launch complex commands with one or more child processes.  We need to wait until they are launched before checkpointing, or else the test fails because too few processes were checkpointed.


gc00:
Forcing a new Travis build by close/open.

elIdrissiYounes (approved):
LGTM!

PR #905, opened 2021-04-15
Small bug fixes

mcandress:
A few very small fixes to the following bugs:

1.  Sporadic crashes in restarted applications with very small process ids (e.g. 127)  due to invalid tid_offset being used.
2.  Propagating rohgarg's workaround to issue 508
3.  Crash on dmctp_coordinator startup due to using 0x1 as string address.


gc00 (approved):
LGTM.  Sorry we took so long to review these really nice bug fixes.

PR #915, opened 2021-08-03
mtcp_restart.c: fix vdso/vvar overlap w/ orig addr

gc00:
* If vvar or vdso overlaps with their pre-checkpoint addresses, and
   are not the same addresses, we now try to move vvar to original
   address, if that does not create an address conflict.  After that,
   it must be safe to move vds0 to its original address.

 * The FIXME comment describes how to fix this fully, by first moving
   to a temporary address, but this is good enough for now.  If an error
   still occurs, the error message refers to this FIXME comment.

   Let's be conservative and first get experience with this, before
   adding the extra fix.  The extra fix would require mremap'ing
   to an empty memory location, and first verifying that it is
   empty, so that the kernel's mremap doesn't unmap something.

This restart bug is observed reliably on SUSE-15 (Cori) for Linux 4.13 when testing dmtcp5.


src/mtcp/mtcp_restart.c
```diff
@@ -793,18 +794,31 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
   }
 
   // Check for overlap between newer and older vDSO/vvar sections.
-  if (doAreasOverlap(vdsoStart, vdsoEnd - vdsoStart,
+  if (
+#if 0
```
JainTwinkle, line 798: @gc00 Do we need this code (\#if 0 ... \#endif) now that we have a fix for overlapping regions?

src/mtcp/mtcp_restart.c
```diff
@@ -1279,6 +1285,28 @@ hasOverlappingMapping(VA addr, size_t size)
...
+                               dest);
+  if (rc == dest) {
+    return 0; // Success
+  } else if (rc == MAP_FAILED) {
+    MTCP_PRINTF("***Error: failed to mremap; errno: %d.\n",
+                mtcp_sys_errno);
+    return -1; // Error
+  } else {
+    // We failed to move the memory to 'dest'.  Undo it.
+    mremap_move(src, rc, size);
```
JainTwinkle, line 1305: @gc00 could you explain what would be the workflow after we undo the mremap_move?
Wouldn't the vvar/vdso will still have overlapping areas?
If we need to abort after undo-ing then why do we need to undo mremap_move in the first place?

PR #918, opened 2021-09-12
RWLock: Don't release mutex after acquiring wrlock.

### LINKED ISSUE #917: Bug in dmtcp_disable_plugin_ckpt()
The following bug was observed in MANA using the latest DMTCP-3.0.  The essence is that dmtcp_disable_plugin_ckpt() calls DmtcpRWTryRdLock(), which tries to calls DmtcpMutexUnlock().  That last "unlock" fails to happen because the checkpoint signal comes in at that moment.

This would be fine, but the checkpoint thread is then blocked trying to call DmtcpMutexLock() on the same mutex lock as above.

The full stack for the checkpoint and user threads follow:
[dmtcp_stack_deadlock.txt](https://github.com/dmtcp/dmtcp/files/7148420/dmtcp_stack_deadlock.txt)

karya0:
Fixes #917.

### DIFF
--- a/src/rwlock.cpp
+++ b/src/rwlock.cpp
@@ -96,6 +96,7 @@ int DmtcpRWLockWrLock(DmtcpRWLock *rwlock)
   while (1) {
     // Detect deadlock.
     if (rwlock->writer == dmtcp_gettid()) {
+      JASSERT(DmtcpMutexUnlock(&rwlock->xLock) == 0);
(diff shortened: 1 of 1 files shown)


karya0:
@gc00 @rohgarg @Twinklejain : Friendly reminder to review.

gc00 (approved):
LGTM

PR #919, opened 2021-10-27
code cleanup: removed duplicate headers

JainTwinkle:
I applied a code cleanup tool called `cppclean` to the DMTCP repository. It gave many general suggestions for code cleanup. Out of many suggestions, removing duplicate system and library headers was straightforward. This should be an easy PR to review.


gc00 (approved):
LGTM

PR #924, opened 2022-01-08
Added new Key-Value store API.

karya0:


src/lookup_service.cpp
```diff
@@ -101,6 +103,64 @@ LookupService::registerData(const DmtcpMessage &msg, const void *data)
...
+  case DMTCP_KVDB_NOT:
+    kvmap[msg.kvdb.key] = ~kvmap[msg.kvdb.key];
+    break;
+
+  case DMTCP_KVDB_AND:
+    if (kvmap.find(msg.kvdb.key) == kvmap.end()) {
+      kvmap[msg.kvdb.key] = msg.kvdb.value;
+    }
+
+    kvmap[msg.kvdb.key] &= kvmap[msg.kvdb.key];
```
JainTwinkle, line 156: Is the key's value supposed to AND to itself or `msg.kvdb.value`?
Also, could you add a comment stating why you assume the value to be present in other operations (OR and XOR) but not in AND? In line 153, you set the value if the key is not found, do you intend to use AND as a set operation?

src/lookup_service.cpp
```diff
@@ -101,6 +103,64 @@ LookupService::registerData(const DmtcpMessage &msg, const void *data)
...
+{
+  KeyValueMap64 &kvmap = _maps64[msg.kvdbId];
+
+  if (kvmap.find(msg.kvdb.key) == kvmap.end()) {
+    JTRACE("Lookup Failed, Key not found.");
+    remote << DmtcpMessage(DMT_KVDB64_GET_FAILED);
+  }
+
+  DmtcpMessage reply(DMT_KVDB64_GET_RESPONSE);
+  reply.kvdb.value = kvmap[msg.kvdb.key];
```
JainTwinkle, line 118: Will the code try to send another message to the remote (line `117-120`) after sending a failed message in the line `114`? Or does a reply to remote work as a `return`?

karya0: Good catch, @JainTwinkle !

src/lookup_service.cpp
karya0: Fixed the bug and added a comment.

src/lookup_service.cpp
```diff
@@ -150,11 +152,13 @@ LookupService::set64(const DmtcpMessage &msg)
     break;
 
   case DMTCP_KVDB_AND:
+    // If key isn't found, set the key to the given value.
     if (kvmap.find(msg.kvdb.key) == kvmap.end()) {
       kvmap[msg.kvdb.key] = msg.kvdb.value;
+      break;
```
gc00, line 158: Here also, I assume you intend to flatten the three commits into one, which is fine.

I do have a small quibble.  I think the intention of the code will be clearer, if you reorganize this as:  `if (kvmap.find(msg.kvdb.key) == kvmap.end()) {...} else { ...}`
and then add a single `break' at the end of the if-then-else, to indicate the end of this `case` statement.

gc00 (approved):
I see where you're going.  It's not critical, but I have one suggestion to make short piece of code slightly more readable.

Thanks for pointing me toward the Redis key-value store system.  I see where you're going.

For others who are following this, I found the following URL helpful:
https://redis.io/topics/data-types-intro
The rest of this review is to help others to see where we're going.

Because Redis supports values of keys that are lists or sets, and it supports 'and'ing' and 'or'ing' new values into the keys, this is likely to suffice for the logic of the MANA project.  Of course, we intend to create a small Redis-like subset for now.  But we hve the option of using a full Redis in the future if needed (e.g., for scalability).

In addition to that,  in case we ever need it. Redis is in the process of adding the possibility of creating a server-side function.  Putthing this all together should give us the functionality that we're looking for, to support such projects as MANA.

src/lookup_service.cpp
karya0: Good point. I'll update it in a followup PR to avoid re-approval of this PR.

PR #926, opened 2022-01-26
Fix a bug when vvar/vdso segment does not exist

fyshhh:
This pull request fixes a bug occurring in `mtcp_restart.c` when a vdso or vvar segment doesn't exist, which may result in undefined behaviour in some functions since they were written with the assumption that these segments are non-zero in size.


gc00 (approved):
LGTM.
This fixes a bug that was probably created when I extended the functionality of tc_restart in PR #915.

PR #930, opened 2022-02-02
Refactor pthread create

gc00:
Still a work-in-progress.

This wrapper for pthread_create() does not call clone().  This is important, since glibc-2.34 defines pthread_create to call clone3 instead of clone.   And the new pthread_create calls clone3 through an internal library-private symbol, instead of calling it through the general symbol "clone3".  So, we have no easy way to create a wrapper around clone3.

This should also produce a cleaner design, that will make it easier to manage the entanglement of libdmtcp.so and libdmtcp_pid.so in the future.

Right now, there's a bug that can be seen by:
[gdbinit.txt](https://github.com/dmtcp/dmtcp/files/7985326/gdbinit.txt)

and: `gdb -x gdbinit.txt --args bin/dmtcp_launch test/dmtcp1`

TODO:  We should create a wrapper around clone3(), and the wrapper should refuse to emulate clone() unless some environment variable like DMTCP_CLONE3 is defined.  Note that in dmtcp-2.6, there is a commit that provides a clone3 wrapper, and defines it in terms of clone, if we want a quick shortcut.

### DIFF
--- a/src/plugin/pid/pid_miscwrappers.cpp
+++ b/src/plugin/pid/pid_miscwrappers.cpp
@@ -153,10 +153,18 @@ vfork()
   return vforkPid;
 }
 
+// This struct must be kept in sync with src/threadwrappers.cpp
 struct ThreadArg {
-  int (*fn) (void *arg);  // clone() calls fn that returns int
-  void *arg;
-  pid_t virtualTid;
+  union {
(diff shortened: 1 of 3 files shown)


gc00:
PR #932 is a better implementation of this PR.  Thanks, @karya0.  I'm closing this PR.

PR #932, opened 2022-02-02
Removed clone wrappers.

karya0:


gc00:
This is still on my TODO list.  Something came up on my side, and I'm delayed in carefully coming back to this issue.

src/threadwrappers.cpp
```diff
@@ -223,19 +109,46 @@ pthread_create(pthread_t *thread,
...
+// Make sure __clone is not called without pthread_create.
+extern "C" int
+__clone(int (*fn)(void *arg),
+        void *child_stack,
+        int flags,
+        void *arg,
+        int *parent_tidptr,
+        struct user_desc *newtls,
+        int *child_tidptr)
+{
```
gc00, line 138: Let's add a FIXME comment here:
    `// FIXME:  Replace this with __clone_standalone_wrapper()`

By the way, note that `src/miscwrappers.cpp` and maybe other places.  But let's leave that in the code if we're agreed on replacing this with new wrappers for `__clone()`.

gc00 (commented):
There are two issues here:
  1.  I'd like to verify that this works with aarch64.  I can't do that currently, because a recent PR (PR #940) broke aarch64 even for compiling DMTCP.  PR #945 is essentially asking you to repair PR #940 to make it work for aarch64.  When it does work, I can come back here and verify that this new PR also works on aarch64.

2. I don't like the idea of removing the `clone()` wrapper, when it may be that some people are using it.  If we are agreed on this PR, then I can bring back the clone wrapper.  (See my inline comment for `__clone()`.)  I'd like to also review a replacement for the clone wrapper (on x86_64 and aarch64) before we push in this PR.

Aside from that, I like the architecture.  It's clean and easy to read.

By the way, fun fact:
  In x86_64, glibc has clone() calling `clone_start()`, and we interpose and replace it completely with our own version.
  In aarch64, glibc does not define `clone_start()`, and instead defines `pthread_start()`.  Essentially, glibc is passing `pthread_start()` down and removing their own version of a clone wrapper, because `pthread_create()` could call either clone() or clone3().  I guess that whoever did that, did not have the time or interest in similarly updating glibc for x86_64.

Since DMTCP currently depends on `__clone()` calling `clone_start()`, DMTCP was broken by recent versions of glibc when they removed `clone_start()` in the case of aarch64.  So, this PR will make aarch64 work again.  Therefore, this PR is very important.  I'll work closely with you on it.

karya0:
> There are two issues here:
> 
> 1. I'd like to verify that this works with aarch64.  I can't do that currently, because a recent PR (PR #940) broke aarch64 even for compiling DMTCP.  PR #945 is essentially asking you to repair PR #940 to make it work for aarch64.  When it does work, I can come back here and verify that this new PR also works on aarch64.

Created #946 to fix the AARCH64 issue.

> 2. I don't like the idea of removing the `clone()` wrapper, when it may be that some people are using it.  If we are agreed on this PR, then I can bring back the clone wrapper.  (See my inline comment for `__clone()`.)  I'd like to also review a replacement for the clone wrapper (on x86_64 and aarch64) before we push in this PR.

Sure, we can create a separate PR to handle __clone/clone3. It would be great to find an application that uses clone/clone3 directly instead of going through pthread_create. 

> Aside from that, I like the architecture. It's clean and easy to read.
> 
> By the way, fun fact:    In x86_64, glibc has clone() calling `clone_start()`, and we interpose and replace it completely with our own version.    In aarch64, glibc does not define `clone_start()`, and instead defines `pthread_start()`. Essentially, glibc is passing `pthread_start()` down and removing their own version of a clone wrapper, because `pthread_create()` could call either clone() or clone3(). I guess that whoever did that, did not have the time or interest in similarly updating glibc for x86_64.
> 
> Since DMTCP currently depends on `__clone()` calling `clone_start()`, DMTCP was broken by recent versions of glibc when they removed `clone_start()` in the case of aarch64. So, this PR will make aarch64 work again. Therefore, this PR is very important. I'll work closely with you on it.

Our clone_start function is library-private, so it doesn't have any interaction with the libc:clone_start. I am not sure if I understand the issue you are talking about here.

gc00:
> > There are two issues here:
> > 
> > 1. I'd like to verify that this works with aarch64.  I can't do that currently, because a recent PR (PR #940) broke aarch64 even for compiling DMTCP.  PR #945 is essentially asking you to repair PR #940 to make it work for aarch64.  When it does work, I can come back here and verify that this new PR also works on aarch64.

> Created #946 to fix the AARCH64 issue.

Great.  I'll test it soon on AARCH64.

>> 2. I don't like the idea of removing the `clone()` wrapper, when it may be that some people are using it.  If we are agreed on this PR, then I can bring back the clone wrapper.  (See my inline comment for `__clone()`.)  I'd like to also review a replacement for the clone wrapper (on x86_64 and aarch64) before we push in this PR.

> Sure, we can create a separate PR to handle __clone/clone3. It would be great to find an application that uses clone/clone3 directly instead of going through pthread_create. 

Agreed about a separate PR.  There are codes that use `clone()` with weird combinations of flags.  One that Rohan and I have been intrigued by is process-in-process by Hori, ..., Balaji, Ishikawa.  It seems to be a collaboration between Riken and Argonne.  Since we had a clone wrapper, there's no harm in spending a little time now to maintain a separate clone wrapper.  I can do that.

> > Aside from that, I like the architecture. It's clean and easy to read.
> > 
... 
> > Since DMTCP currently depends on `__clone()` calling `clone_start()`, DMTCP was broken by recent versions of glibc when they removed `clone_start()` in the case of aarch64. So, this PR will make aarch64 work again. Therefore, this PR is very important. I'll work closely with you on it.

> Our clone_start function is library-private, so it doesn't have any interaction with the libc:clone_start. I am not sure if I understand the issue you are talking about here.

You could be right on that.  I'll look at this again when I have time.  In this session, I spent most of my time on fixing smaller problems in previous PR's :-(.  You'll see that I have a new PR #947 to fix PR #937, and I have a smaller fix requested for PR #946 before we can approve it.

src/threadwrappers.cpp
```diff
@@ -37,13 +37,49 @@ using namespace dmtcp;
...
 
-// Invoked via pthread_create as start_routine
-// On return, it calls mtcp_threadiszombie()
-static void *
-pthread_start(void *arg)
+static bool
+prepareThreadCreate()
 {
-  Thread *thread = (Thread *) arg;
+  /* pthread_create() should acquire the thread-creation lock. Not doing so can
```
gc00, line 43: Minor:  `pthread_create()` -> `pthread_create and clone/clone3`

Can the first type of deadlock below still happen, now that pthread_create and clone are indpendent?

gc00 (approved):
This looks good to me.

Later, I'll create a DMTCP test for using clone() to create a thread.    But for now, I trust you on the functionality.  Let's approve it now, for the sake of efficiency.

By the way, I've now tested this PR on aarch64.  It does indeed work well.  There are a few tests on aarch64 that are failing for particular reasons.  dmtcp5 and shared_memory1.  Those random failures might possibly be due to the need for PR #957, but I'll test again later on aarch64 with that PR.

However, the interesting failure on aarch64 concerns the `vfork1` test.  Apparently, `__builtin_frame_address(0)` does not work the same on aarch64 as on x86_64.  So, in `execwrappers.cpp:vfork()`, the arithmetic with `stackStart` and `stackSize` produces a negative number for `stackSize`.  I'm still trying to do a little research to figure out in what way aarch64 is different.  Is there a way to do this part without `__builtin_frame_address(0)`?

PR #933, opened 2022-02-04
Refactored dmtcp_restart and shareddata.

karya0:


src/util_exec.cpp
```diff
@@ -522,86 +522,6 @@ Util::freePatchedArgv(void *ptr)
...
-  char *newArgs[] = {
-    (char *)mtcprestart.c_str(),
-    (char *)"--stderr-fd",
-    protected_stderr_fd_str,
-    (char *)"--fd",
-    buf,
-    NULL
-  };
-  JTRACE("launching mtcp_restart --fd")(fd)(path);
-
```
gc00, line 556: This FIXME could be needed in certain cases.  Removing it completely means mtcp_restart can break in some cases, and we won't have this FIXME on how to fix it.

See:  `mtcp/mtcp_check_vdso.ic` and look at `return setPersonalityEnv("MTCP_OLDPERS", oldpers_str, environ);`.

The issue is that we don't want to use libc in mtcp_restart.c.  So, we can't do `setenv()` there.  But we need the environment variable `"MTCP_OLDPERS"` in case we need to record the old personality and then exec into a new personality for things like removing the address space randomization.  So, we want to create the environment variable `"MTCP_OLDPERS"` in advance, inside `dmtcp_restart.cpp` using `setenv()`, so that the environment variable will later be available to `mtcp_restart`.

There's a comment  here in the `FIXME` about adding an explanation of the logic.  You might want to copy some of the above paragraph into a comment explaining the logic.

gc00 (commented):
See the inline comment.

src/util_exec.cpp
karya0: Hi @gc00, this code was stale and not called from anywhere. This was from an earlier model when we used to create child processes _after_ restoring the parent process. We switched to the current model of forking child processes _before_ running mtcp_restart 7-8 years ago but forgot to remove this code from util_exec.cpp. The personality handling code is currently in all inside src/mtcp/. If you like, we can add a new comment somewhere about personality restoration logic.

karya0:
> See the inline comment.
> 
> If you want, you can split this PR into two parts. I can approve the rest of it, while we discuss the code being removed in `util_exec.cpp` -- or you can address my concern directly in this PR.

Unfortunately untangling the two commits would require some non-trivial rebasing, so let's just continue to conversation here :).

src/util_exec.cpp
karya0: Here is the original commit for reference: https://github.com/dmtcp/dmtcp/commit/9bf02cf695e31077309c94f3f9235100d97fda0b

gc00 (commented):
I updated our inline conversation.  There's still something that I don't understand here.  Please look at the inline conversation.

src/util_exec.cpp
karya0: I looked at `setPersonalityEnv()` definition, and if I am reading it correctly, it _does_ have and equivalent of setenv in the code. Here is the relevant part: https://github.com/dmtcp/dmtcp/blob/master/src/mtcp/mtcp_check_vdso.ic#L386-L403

The current call sequence is: mtcp_check_vdso() -> setenv_oldpers() -> setPersonalityEnv(), so I don't think it's obsolete yet :).

gc00 (approved):
@karya0, I tested on aarch64 in PR #942.  I had to fix `src/Makefile.am` to make it compile (missing libatomic.so for aarch64).  Even so, I'm getting a runtime error:
  `[40000] ERROR at pidwrappers.cpp:118 in dmtcp_gettid; REASON='JASSERT(_real_gettid() == _real_getpid()) failed'`

But I assume that's independent of the current PR.  So, I'm approving this PR.

As we discussed, please leave a comment about the possibility of setting personality() on `mtcp_restart`.  But I agree, we don't need that currently, and we can modify `src/tlsutil.h` to remove the relevant code, and leave only the comment.

gc00:
@karya0, I tested on aarch64 in PR #942.  I had to fix `src/Makefile.am` to make it compile (missing libatomic.so for aarch64).  Even so, I'm getting a runtime error:
  `[40000] ERROR at pidwrappers.cpp:118 in dmtcp_gettid; REASON='JASSERT(_real_gettid() == _real_getpid()) failed'`

But I assume that's independent of the current PR.  So, I'm approving this PR.
[ Actually, this comment probably goes with PR #940. ]

PR #934, opened 2022-02-04
Some new APIs and functions

karya0:


src/util_misc.cpp
```diff
@@ -222,6 +222,22 @@ Util::skipBytes(int fd, size_t count)
   return totalSkipped;
 }
 
+ssize_t
+Util::readAllFromFile(const char *path, void *buf, size_t count)
```
gc00, line 226: Why not change `readAllFromFile` to `readAll`?  Won't C++ allow this as two signatures for the same function name?

jalib/jconvert.h
```diff
@@ -40,7 +40,7 @@ StdLibEC(const dmtcp::string &s, bool strict)
 {
```
gc00, line 40: I don't see where `StdLibEC<a, b, c>()` is used.  I only see `StdLibEC<a, b>()`.  If `StdLibEC<a, b, c>()` is not used, let's just delete it.  (Or can you point me to where it's used?)

src/processinfo.cpp
```diff
@@ -616,6 +616,23 @@ ProcessInfo::vdsoOffsetMismatch(uint64_t f1, uint64_t f2,
          (f3 != _gettimeofday_offset) || (f4 != _time_offset);
 }
 
+void
+ProcessInfo::addKeyValuePairToCkptHeader(const string &key, const string &value)
```
gc00, line 620: Why is this called `addKeyValuePairToCkptHeader()`?  When I think "CkptHeader", I imagine the header of a DMTCP checkpoint image file?  But if that's the case, where is this function used?

If you'll add new code later that uses it, can you add a comment for this function now, saying how it's supposed to be used?

gc00 (commented):
I'm asking for small changes here.  If you add the changes, I can review and approve this quickly.

jalib/jconvert.h
karya0: The <a,b> functions are used for `strtof/strtold` while the <a,b,c> variant is used for `strol` variants.

jalib/jfilesystem.cpp
```diff
@@ -266,19 +266,23 @@ jalib::Filesystem::FileExists(const dmtcp::string &str)
   }
 }
 
-dmtcp::vector<int>
-jalib::Filesystem::ListOpenFds()
+dmtcp::vector<dmtcp::string>
+LsInternal(const dmtcp::string& dir, int *procSelfFd = nullptr)
```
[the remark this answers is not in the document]
karya0: Changed it to ListDirEntries.

src/processinfo.cpp
karya0: The ProcessInfo object effectively acts as the checkpoint header for DMTCP. It is what dmtcp_restart reads after opening the ckpt image. The Mana plugin will use this API to add information that can be read/interpreted by MANA/MPI-specific code in dmtcp_restart without requiring any other modifications.

Most of the new functions exposed here will be used in the MANA branch, but since these are generic enough, I decided to  create a separate PR for them, just to keep things simpler for the upcoming MANA PR.

gc00: Thanks.  That's fair.  But we still need a comment to explain this to the casual reader.  I propose:

`// NOTE:  ProcessInfo object acts as the checkpoint header for DMTCP.`

gc00 (approved):
I'm approving it now for convenience.  But _please_ do add the one-line comment that I'm requesting (see in-line comment) in the function  `ProcessInfo::addKeyValuePairToCkptHeader()`.

PR #940, opened 2022-02-05
Updated/Refactored TLS save/restore handling.

karya0:
We don't need to restore TLS in mtcp_restart. Instead, we do it after
making a jump to ThreadList::postRestart.


gc00 (commented):
I'm concerned about what we lose in `mtcp/tlsutil.h`.  It would take me significant time to trace the full logic, and make sure that it has the same functionality.  I could do that a little later this week, if needed.

In the meantime, I have two basic questions:

1.  Where have you tested this?  Did you test on an i386 virtual machine,and/ or compile DMTCP with -m32, to verify the test suite?  If so, what was the O/S?  Also, did you test on WSL?  Finally, did you test on an aarch64 machine?  (If not, I could test on an aarch64 machine for you.)

2. Are you sure that the `gs` register is also restored?  Normally, most applications don't care.  But there are esoteric experiments using the `gs` register.  And one of them provided us with the gs patch to make sure that it properly restores the `gs` register.

karya0:
Hi @gc00, I have inlined my responses.

> I'm concerned about what we lose in `mtcp/tlsutil.h`. It would take me significant time to trace the full logic, and make sure that it has the same functionality. I could do that a little later this week, if needed.

mtcp/tlsutil.h contained macro definitions for tls_[gs]et_thread_area for fs/gs registers, etc.. What I have done now is wrapped them inside functions tls_[gs]et_thread_area. However, I have made separate copies of these funcitons for all architectures that we support, along with separate definitions of TLSInfo struct. This has simplified some of the #ifdefs/#defines that we had to do when we had a single stuct and single macro signature.

> 1. Where have you tested this?  Did you test on an i386 virtual machine,and/ or compile DMTCP with -m32, to verify the test suite?  If so, what was the O/S?  Also, did you test on WSL?  Finally, did you test on an aarch64 machine?  (If not, I could test on an aarch64 machine for you.)

Here is where I tested:
* WSL with 64-bit Ubuntu 20.04 with and without --enable-m32. 
* On login-faculty with and without --enable-m32
* There is also a Travis run with Ubuntu 20.04 (https://app.travis-ci.com/github/karya0/dmtcp/builds/245964340). 

I have not tested it on aarch64 so it would great if you could test it there.

> 2. Are you sure that the `gs` register is also restored?  Normally, most applications don't care.  But there are esoteric experiments using the `gs` register.  And one of them provided us with the gs patch to make sure that it properly restores the `gs` register.

The gs register is being restored. You can see that here: https://github.com/dmtcp/dmtcp/pull/940/files#diff-d248b0979c123acc429ab0f9829671539bf79001058e487481b62e82c6425a3fR359 .

gc00 (approved):
@karya0, I tested on aarch64 in PR #942.  I had to fix `src/Makefile.am` to make it compile (missing libatomic.so for aarch64).  The builtin `__atomic_compare_exchange` wasn't compiling.  Even so, I'm getting a runtime error:
  `[40000] ERROR at pidwrappers.cpp:118 in dmtcp_gettid; REASON='JASSERT(_real_gettid() == _real_getpid()) failed'`

But that's probably unrelated to his PR.  So, I'm approving.

PR #944, opened 2022-02-09
Fixes call by target of dlsym(RTLD_NEXT,...)

gc00:
I'm open to suggestions for possible improvemets:
1.  We probably want to move this code to somewhere in `libdmtcp.so`, but to which file?
2. This code improves on the previous `dlsym()` wrapper.  That code assumed that `RTLD_NEXT` is used only to wrap a function in libc.so.  However, it's not perfect.  Is there a __simple__ way to improve on this?  (If it's not simple, then this is good enough for now.  "The perfect is the enemy of the good.")

For convenience, I'm copying here the comments found in `src/plugin/svipd/sysvipcwrappers.cpp`:

```
//   We need to wrap dlsym to disable and re-enable ckpt.  But then,
// dlsym(RTLD_NEXT, ...) no longer works, since glibc:dlsym() looks up
// the stack to find the caller's library, and discovers a DMTCP library
// instead of the original target library.
//   It was originally placed in the svipc plugin so that RTLD_NEXT would
// point to glibc.  But some targets want RTLD_NEXT to point to a library
// before glibc.  This happened when ompt_start_tool called dlsym()
// to see if there was a preferred definition later in the search path.
// By default, dlsym() should return NULL.  But now, dlsym is called by
// DMTCP, and so dlsysm discovers the Intel OpenMP library (libiomp5.so)
// later in the search path.
// FIXME:  This code belongs in 'src' in libdmtcp.so, and not here.
// FIXME:  This code searches from the beginning of the link map, using
//         dl_iterate_phdr().  It assumes that the first match of symbol
//         is the wrapper function, and that RTLD_NEXT should assume that
//         the caller starts with that library, in order to find the next
//         match.  This does not have to be the case.
```

### DIFF
--- a/src/plugin/svipc/sysvipcwrappers.cpp
+++ b/src/plugin/svipc/sysvipcwrappers.cpp
@@ -28,6 +28,11 @@
 #include <sys/msg.h>
 #include <sys/shm.h>
 
+#ifndef _GNU_SOURCE
+# define _GNU_SOURCE
+#endif
+#include <link.h>
+
 #undef msgrcv
 
 #include "jassert.h"
@@ -40,18 +45,18 @@ using namespace dmtcp;
 static struct timespec ts_100ms = { 0, 100 * 1000 * 1000 };
 
 /*
- * In Open MPI 2.0, shmdt() is intercepted by modifying libraries' global offset
- * table, meaning that _real_shmdt() will be redirected into OpenMPI's hook
- * function, instead of libc's shmdt(). The hook function finally calls syscall()
- * with the corresponding syscall number. The inside_shmdt variable indicates
- * if the code is inside our shmdt() wrapper. If so, our syscall wrapper simply
- * calls _real_syscall(), avoiding the recursive call to the shmdt() wrapper.
- * See the wrapper of syscall() in miscwrappers.cpp and pid_miscwrappers.cpp.
- *
- * FIXME: for the long term, we need to think about the case where user code
- * modifies its own global offset table.
- *
- * */
+ * In Open MPI 2.0, shmdt() is intercepted by modifying libraries' global
+ * offset table, meaning that _real_shmdt() will be redirected into
+ * OpenMPI's hook function, instead of libc's shmdt(). The hook function
+ * finally calls syscall() with the corresponding syscall number. The
+ * inside_shmdt variable indicates if the code is inside our shmdt()
+ * wrapper. If so, our syscall wrapper simply calls _real_syscall(),
+ * avoiding the recursive call to the shmdt() wrapper.  See the wrapper
+ * of syscall() in miscwrappers.cpp and pid_miscwrappers.cpp.
+
+ * FIXME: for the long term, we need to think about the case where user
+ * code modifies its own global offset table.
+ */
 static __thread bool inside_shmdt = false;
 
 /******************************************************************************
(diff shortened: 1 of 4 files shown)


gc00:
@karya0 .  Ping!  This PR came about to fix a bug encountered by one of the DMTCP "customers".

Functionally, this PR works and could be pushed in now.  But I'd like your special input, to recommend how to integrate it cleanly into the DMTCP architecture.  So, I'm waiting for a review particularly from you.  Feel free to reach out to me, if you'd like me to lead you through the issues interactively.

src/plugin/svipc/sysvipcwrappers.cpp
```diff
@@ -183,9 +188,72 @@ shmdt(const void *shmaddr)
...
+// instead of the original target library.
+//   It was originally placed in the svipc plugin so that RTLD_NEXT would
+// point to glibc.  But some targets want RTLD_NEXT to point to a library
+// before glibc.  This happened when ompt_start_tool called dlsym()
+// to see if there was a preferred definition later in the search path.
+// By default, dlsym() should return NULL.  But now, dlsym is called by
+// DMTCP, and so dlsysm discovers the Intel OpenMP library (libiomp5.so)
+// later in the search path.
+// FIXME:  This code belongs in 'src' in libdmtcp.so, and not here.
+// FIXME:  This code searches from the beginning of the link map, using
```
karya0, line 245: @gc00: I am not sure if this logic would work. Even some of our plugins use RTLD_NEXT and if the same symbol is wrapped multiple times, we'll be in trouble if we always skip the first match only. I think what we need is to record the return-address and use that to figure out the _next_ symbol.

src/plugin/svipc/sysvipcwrappers.cpp
```diff
@@ -183,9 +188,72 @@ shmdt(const void *shmaddr)
...
+static const char *dlsym_symbol = NULL;
+static void *dlsym_retval;
+// extern "C"
+static int
+callback(struct dl_phdr_info *info, size_t size, void *data) {
+  void *handle = dlopen(info->dlpi_name, RTLD_LAZY);
+  dlerror(); // initialize for errors
+  void *address = dlsym(handle, dlsym_symbol);
+  if (address == NULL && dlerror() != NULL) {
+    dlsym_retval = dlsym(handle, dlsym_symbol); // Set dlerror to error string
```
karya0, line 201: This seems redundant. Wouldn't dlerror already be set to the right value because of the previous call?

src/plugin/svipc/sysvipcwrappers.cpp
```diff
@@ -183,9 +188,72 @@ shmdt(const void *shmaddr)
...
+    dlclose(handle);
+    return 0;
+  } else {
+    dlclose(handle);
+  }
+  if (address != NULL) {
+    dlsym_addr_instance++;
+    printf("Symbol %s in %s: found at address %p\n",
```
karya0, line 209: Let's convert printfs to JTRACEs.

src/plugin/svipc/sysvipcwrappers.cpp
```diff
@@ -183,9 +188,72 @@ shmdt(const void *shmaddr)
...
+// the stack to find the caller's library, and discovers a DMTCP library
+// instead of the original target library.
+//   It was originally placed in the svipc plugin so that RTLD_NEXT would
+// point to glibc.  But some targets want RTLD_NEXT to point to a library
+// before glibc.  This happened when ompt_start_tool called dlsym()
+// to see if there was a preferred definition later in the search path.
+// By default, dlsym() should return NULL.  But now, dlsym is called by
+// DMTCP, and so dlsysm discovers the Intel OpenMP library (libiomp5.so)
+// later in the search path.
+// FIXME:  This code belongs in 'src' in libdmtcp.so, and not here.
```
karya0, line 244: I agree. This code belongs in libdmtcp.so along with the code for dlopen (which currently sits in plugin/dl directory).

gc00:
I'll fix this PR, and add a test for it later when I have the cycles.  I haven't forgotten.

src/plugin/svipc/sysvipcwrappers.cpp
gc00: Done.

src/plugin/svipc/sysvipcwrappers.cpp
gc00: I've modified the comment for clarity:
>  // Clear any old errors; will return non-null only on new error

src/plugin/svipc/sysvipcwrappers.cpp
gc00: I've created `src/dlwrappers.cpp` now, and moved the 'dlsym' wrapper to there.

gc00:
Please note the FIXME in src/dlwrappers.cpp.  Should we move `_real_dlsym()` to `src`?  I'm guessing yes.

```
// FIXME:  _real_dlsym is currently in src/plugin/pid/pid_syscallsreal.c
//         Should we move it to src/syscallsreal.c ?  
````

File: src/plugin/svipc/sysvipcwrappers.cpp (hunk 1 above)
Comment (gc00) [reply src/plugin/svipc/sysvipcwrappers.cpp:245]: > ... we'll be in trouble if we always skip the first match only. I think what we need is to record the return-address and use that to figure out the next symbol.

I assume you meant to record the address of the symbol in the current library, and then figure out the next symbol.  But then, what if 'dlsym' is called in a library that doesn't define the target symbol?

I suppose we could add that heuristic on top of the current heuristic.  And we could even do things like save old values of dlsym, to decide if RTLD_NEXT is making progress or getting stuck at one library.  But before we go too deeply down that road, let's talk interactively.

For now, I think the perfect is the enemy of the good.  And I'll note that currently, this PR passes all of our tests.  I propose that talk about the roadmap for dlsym, and then decide what to commit now, and what to add a FIXME for.

In general, I don't see a simple fix for all cases.  And this is needed by one of the DMTCP "customers".  They define a function that is _not_ part of glibc.  And they use dlsym to test if their is a later definition of the same function.  If there is not, then they accept the definition of the current library as the "default".  I intend to add a test for this case.  We were failing before, for this case.

**Comment by gc00** [comment]
@karya0,
I think I've resolved all our conversations except one.  Please do mark those resolved.

The one discussion is about your comment `we'll be in trouble if we always skip the first match only`.  As I note there, this passes all tests, and it fixes the bug of a customer.  (I'll add a test to verify that later.)  Let's talk together.  I worry that we will otherwise get into a tight loop, where the perfect is the enemy of the good.

File: src/plugin/svipc/sysvipcwrappers.cpp (hunk 2 above)
Comment (karya0) [reply src/plugin/svipc/sysvipcwrappers.cpp]: I think you misread my comment. It was about the second dlsym() call on line 201.

File: src/plugin/svipc/sysvipcwrappers.cpp (hunk 1 above)
Comment (karya0) [reply src/plugin/svipc/sysvipcwrappers.cpp]: Let's assume A->B->C->libdmtcp.so->D->libc.so, where, A-D are plugin libraries. If A, B and D had wrappers for a function F(), would this code return the correct value when calling dlsym(RTLD_NEXT, "F") from B or D?

For handling cases where C calls dlsym(RTLD_NEXT, "F"), we should have a thread-local variable `curLib` which is used to record info about the library that called dlsym(RTLD_NEXT, "F") (e.g., by lookup up the return address). Then we continue returning `0`, until we see curLib in the callback, i.e.:
```
callback(struct dl_phdr_info *info, size_t size, void *data) {
  if (curLib != NULL && curLib != info) {
    return 0;
  }

  if (curLib == info) {
    curLib = NULL;
    return  0;
  }
 ...
}
 ```

gc00:
@karya0,
    I've now added a long comment at the beginning of `dlwrappers.cpp`.  I've done some searching to figure out the history.
    Probably, this all began with trying to support Open MPI, which patches 'shmdt'.  Now, with MANA, we no longer need to directly checkpoint Open MPI.

Note that internally, DMTCP uses:
  &nbsp;&nbsp;&nbsp;&nbsp;  _real_dlsym (macro) -> NEXT_FNC() (macro) -> dmtcp_dlsym() (fnc. ptr.) -> libc:dlsym().
So, the internal logic of DMTCP never goes through a DMTCP wrapper function for 'dlsym'.  I expand on that in my large comment at the beginning of `dlwrappers.cpp`.

In principle, it's possible that we don't need a wrapper for 'dlsym' at all, anymore.  I'm fine if we decide that, but then let's keep `dlwrappers.cpp`, and put a big `#ifdef 0` around this new code.  I guess we also don't need `__real_dlsym` either, in that case.  It's possible that we'll see this issue again, where the target application does something unusual with 'dlsym'.  Further, this wrapper does allow us to disable checkpointing in the middle of a call to 'dlsym'.  That could be useful.

Alternatively, we can keep the 'dlsym' wrapper for target executables in this PR, and improve it later, if needed for a new target.

karya0 (approved):

PR #945, opened 2022-02-10
Revert "Updated TLS save/restore handling."

gc00:
* This reverts commit 485b5232b2757fcc4bc05a1672cfbdced9978f73.
 * That commit does not compile on aarch64.  tlsAddr field is not defined.
   To see the problem, look at '#ifdef __aarch64__' in src/tls.cpp
     and look at it in src/mtcp/tls_util.h .
     You'll see that you're replacing a macro with a function and the
     arguments of the macro and function don't match.

Either we need to fix PR #940 or else back out of it.  That PR seems to have
been verified for x86-64 and i386, but not for aarch64.

This PR backs out of it.  And we cannot push in PR #942 until this is fixed.


karya0:
Looks like I made a copy-paste error in updating ThreadTLSInfo struct for ARM. I have fixed it in #946. We can close this PR now.

gc00:
Thanks for the fix.  I'll test on aarch64 to verify this (probably later tonight).  And then I'll close this PR.

In addition, this should unblock PR #942.  I'll then be able to push that in, and then return to reviewing PR #932.  As part of PR #932, I can take responsibility for adding a `clone()` wrapper as a separate commit.  And we can then push in PR #932 and a clone wrapper commit.

Looks like we have a plan!  Nice.

gc00:
Closing this PR as agreed (replaced by PR #946).

PR #937, opened 2022-02-04
Fixes to syscall-tester test.

karya0:

PR #947, opened 2022-02-11
syscall-tester.c: revert BasicMknod test changes

gc00:
The revision to the BasicMknod test of `syscall-tester.c` is failing on Suse Enterprise 15 (Linux 4.12, gcc-7.5, glibc-2.26).
This PR reverts the changes to BasicMknod that were made in PR #937, while keeping the other improvements.

```
--- Line 4154: BasicMknod() ---
 mknod(): path=/tmp/dmtcp-syscall-tester-FzrRG8, mode=0x21c0(020700), dev=0
                            ret = -1      
                Succeeded Phase 1
BasicMknod():4161
Failed Phase 2: This call did something unexpected.
BasicMknod():4161
Fail Phase 2: Aborting test because of catastrophic failure
Ending Test: [BasicMknod: Can I make pipes and not other stuff?]
------------------------------------------------------------- 
```

As a side comment, the previous PR #937 causes BasicMknod to always fail on Suse Enterprise 15.  On CentOS 7, _with this PR that reverts it_, it seems to always succeed on Suse Enterprise 15.  I have seen it fail once on CentOS 7 with this PR, but I can't reproduce that.  It usually passes on CentOS 7.

### DIFF
--- a/configure
+++ b/configure
@@ -661,6 +661,7 @@ HAS_PYTHON3
 HAS_PS
 HAS_EPOLL_CREATE1
 HAS_CMA
+WSL
 AARCH64_HOST
 AARCH64_HOST_FALSE
 AARCH64_HOST_TRUE
@@ -6535,13 +6536,15 @@ fi
 
 { $as_echo "$as_me:${as_lineno-$LINENO}: checking if this is WSL (Windows Subystem for Linux)" >&5
 $as_echo_n "checking if this is WSL (Windows Subystem for Linux)... " >&6; }
-if uname -r | grep Microsoft > /dev/null; then
+if uname -r | grep -i Microsoft > /dev/null; then
   is_wsl='yes'
   wsl_version=`uname -r`
 else
   is_wsl='no'
 fi
 if test "$is_wsl" == "yes"; then
+  WSL=yes
+
 
 cat >>confdefs.h <<_ACEOF
 #define WSL $wsl_version
--- a/configure.ac
+++ b/configure.ac
@@ -458,13 +458,14 @@ else
 fi
 
 AC_MSG_CHECKING([if this is WSL (Windows Subystem for Linux)])
-if uname -r | grep Microsoft > /dev/null; then
+if uname -r | grep -i Microsoft > /dev/null; then
   is_wsl='yes'
   wsl_version=`uname -r`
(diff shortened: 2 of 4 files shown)


karya0:
Hi @gc00, can you quickly check this test without dmtcp? If we revert the PR, it will fail on WSL.

gc00:
I've checked this in WSL (using Ubuntu 20.04).  I've gotten to the bottom of this.
1.  I used GDB on `syscall-tester` and did `(gdb) break mknod`
2. It does two tests on `mknod()`, first with `S_IFIFO` and then with `S_IFCHR `.  The device number is 0 in both cases.
3. Normally, according to the man page, we would expect it to return 0 (success) for `S_IFIFO`, and -1 (failure) for `S_IFCHR `.  Indeed, it does this in CentOS 7 (and presumably Susc Enterprise 15).
4. In WSL, `mknod()` returns 3 (sic!) for `S_IFIFO`, and 0 (success) for `S_IFCHR `.   I assume that WSL is lying to us in both cases, and especially for `S_IFCHR `, since we didn't give it a reasonable device number.  So, the conclusion is that WSL has not implemented `mknod()` and Ubuntu (or the Microsoft hypervisor) does not interpret the result correctly.

I'll modify this PR to add `#ifndef WSL` around the `BasicMknod` test, and re-submit it.  After that, you're free to simply approve this PR; or else create your own PR in which you apply my `#ifndef WSL` to the underlying code and fix it so that Enterprise Suse 15 can pass the test.

gc00:
Okay, this is ready for review.  I have tested on WSL, and it works now.
1.  The first commits adds `AC_SUBST()` for `WSL` in `configure.ac`, `test/Makefile.in`, etc.
2. The second commit, in addition to reverting BasicMkod()` in `syscall-tester.c`, it also adds an `#ifndef WSL` to omit that test for WSL.

@karya0, you have a choice.
1.  The simplest thing is simply to approve this.
2. But if you still like your modification of `BasicMknod()`, we'll have to modify it to work with Suse Enterprise 15 (and of course also keep the `#ifndef WSL` that I provide in the first commit).

karya0 (approved):
Sounds reasonable to me. Let's merge it in.

PR #948, opened 2022-02-11
Enhance and fix gdb-dmtcp-utils (data seg, perm)

gc00:
1. Improved 'help' message when it's first loaded
2. whereis-address also works with data segments and segments with no permission now.
3. It also now displays the permissions of the segment that it finds.


JainTwinkle (approved):
LGTM :)

PR #953, opened 2022-02-12
Remove '#ifdef WSL' and WSL comments

gc00:
* As of Windows 21H!, WSL is now mature.  We don't need the special cases.


karya0 (approved):
LGTM.

PR #957, opened 2022-02-15
volatile -> volatile __attribute((aligned))

gc00:
* ISO C does not guarantee that an int is aligned.  So, it won't be
   atomic if it crosses a cache boundary.  This adds ATOMIC_SHARED macro.

@gourryinverse convinced me that the ISO C standard says that C doesn't have to respect a 4-byte int (e.g., on x86_64) as atomic, and that it can still be misaligned.  ISO C says that a struct must respect the chosen alignments of its fields, but it doesn't seem to say what is the "natural" alignment of an int.

As an interesting example, consider this partial evidence provided by him.  An 8-byte pointer turns out to be only 2-byte aligned.  So, maybe the compiler normally aligns 4-byte int's correctly, but we can't depend on that.  This is all about evidence, since we have no way of directly proving what a compiler is required to do.

```
#include <assert.h>

int g_x;
int* g_x_p = &g_x;

int func(int a, int b, int c, int d)
{ return a + b + c + d; }

int main(void) {
  int l_x;
  int* l_x_p = &l_x;

  l_x = 2;
  *l_x_p = 3;
  g_x = 4;
  *g_x_p = 5;

  assert(func(l_x, *l_x_p, g_x, *g_x_p) == 16);
}
```

Note that `g_x_p` is only 2-byte alligned, when you run this, and print addresses with GDB.

### DIFF
--- a/jalib/jalloc.cpp
+++ b/jalib/jalloc.cpp
@@ -28,6 +28,8 @@
 #include <string.h>
 #include <unistd.h>
 
+#define ATOMIC_SHARED volatile __attribute((aligned))
+
 // Make highest chunk size large; avoid a raw_alloc calling mmap()
 // during /proc/self/maps
 #define MAX_CHUNKSIZE (4 * 1024)
@@ -269,14 +271,14 @@ class JFixedAllocStack
     };
     struct StackHead {
       uintptr_t counter;
-      FreeItem* volatile node;
+      FreeItem* ATOMIC_SHARED node;
     };
 
   private:
     StackHead _top;
     size_t _blockSize = 0;
     char padding[128];
-    int volatile _numExpands;
+    ATOMIC_SHARED int _numExpands;
 };
 } // namespace jalib
 
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -48,7 +48,10 @@
 extern "C" {
 #endif // ifdef __cplusplus
 
-#define LIB_PRIVATE              __attribute__((visibility("hidden")))
+#define LIB_PRIVATE __attribute__((visibility("hidden")))
+#define ATOMIC_SHARED_GLOBAL volatile __attribute((aligned))
+// Same as global macro, but by convnetion, use this for local variables:
+#define ATOMIC_SHARED volatile __attribute((aligned))
 
 typedef enum eDmtcpEvent {
   DMTCP_EVENT_INIT,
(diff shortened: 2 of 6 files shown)


jalib/jalloc.cpp
```diff
@@ -28,6 +28,8 @@
 #include <string.h>
 #include <unistd.h>
 
+#define ATOMIC_SHARED volatile __attribute((aligned))
```
karya0, line 31: Can we put this and other macros in dmtcp.h near https://github.com/dmtcp/dmtcp/blob/master/include/dmtcp.h#L51 ? This would allow other files to use it as needed.

jalib/jalloc.cpp
```diff
@@ -269,14 +271,14 @@ class JFixedAllocStack
     };
     struct StackHead {
       uintptr_t counter;
-      FreeItem* volatile node;
+      FreeItem* ATOMIC_SHARED node;
```
karya0, line 274: It's existing code with this reversed order, but as part of this PR, can you please change the order to `ATOMIC_SHARED <type> var;`? Here and below.

karya0 (approved):
Approved with some minor suggestions.

jalib/jalloc.cpp
gc00: I tested, and moving ATOMIC_SHARED changes the meaning.  Right now, it's a pointer to a volatile 'node'.  If we reverse it, it becomes a volatile pointer to 'node'.  And of course, with that different meaning, it's inconsistent with how it's used.

gc00:
I included the commit `Fix for commit 073e69, to support gcc-4.8` for ease of testing the current PR.  But that should be pushed in first as PR #958, before this PR is pushed in.

I've now made the requested changes, except for the "reversed order" mentioned in the inline conversation.

jalib/jalloc.cpp
gc00: Done.

gourryinverse:
Also just adding this comment for the record - the use of "volatile" and alignment does not imply atomicity of anything but a write instruction.   The compiler could very well break up an increment operation into 3 instructions.

i.e.

`*ptr++`
or
`*ptr += 1`

and easily produce the following assembly

```
mov rax, [rcx]
inc rax
mov [rcx], rax
```

which makes what appears to be an atomic operation into a blatant race condition.   This access mechanism is incredibly common if you're working with structures, and most people don't really think too much about it.

The inclusion of ATOMIC_GLOBAL gives you at least a chance at a later time to identify data which should probably use atomic intrinsics ( https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html ) should you find any of these race conditions being relevant.

PR #960, opened 2022-02-19
Removed volatile from StackHead definition.

karya0:
The __attribute((aligned)) causes the struct to grow beyond double-word,
causing CAS to fail. Since this field is updated safely, we don't need it to be
volatile.


gc00 (approved):
LGTM.

PR #962, opened 2022-02-20
Adding clone1 test (emulating child thread)

gc00:
@karya0,
    This `test/clone.c` segfaults on `bin/dmtcp_launch test/clone1`.  This has been the case since PR #932 .  Admittedly, prior to PR #932, we could launch, checkpoint, and resume clone1, but we were failing to restart clone1.
    Nevertheless, even if we comment out 'clone' in `atuotest.py`, this seems like a good test to have in reserve for the future.
    If we need to, we can simply check it in, with this test commented out.

But I'm hoping that this will help you to analyze your PR #932.  Maybe there's something simple there that you can easily fix.  To test, just do:
&nbsp;&nbsp;&nbsp;&nbsp; `make -j check-clone1 && bin/dmtcp_launch test/clone1`

### DIFF
--- a/test/clone1.c
+++ b/test/clone1.c
@@ -1,0 +1,65 @@
+/* Compile with:  gcc THIS_FILE -lpthread */
+
+#define _GNU_SOURCE
+#include <sched.h>
+#include <pthread.h>
+#include <semaphore.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
+int start_routine(void *);
+
+sem_t sem;
+
+
+int
+main()
+{
+  void *arg;
+  void *stack;
+
+  sem_init(&sem, 0, 0);
+
+  while (1) {
+    arg = malloc(10);
+    *(int *)arg = 42;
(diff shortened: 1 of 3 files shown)


karya0:
Looks like `make check-clone1` works fine on Ubuntu.
```
clone1         ckpt:PASSED; rstr:PASSED -> ckpt:PASSED; rstr:PASSED
```

I think that's a lucky run. We are creating a child thread with too-small a stack using malloc(4096). Please create a larger stack, at least 2MB, otherwise we run the risk of putting too much data on the stack in our own wrappers.

test/clone1.c
```diff
@@ -0,0 +1,65 @@
...
+{
+  void *arg;
+  void *stack;
+
+  sem_init(&sem, 0, 0);
+
+  while (1) {
+    arg = malloc(10);
+    *(int *)arg = 42;
+    stack = malloc(4096);
```
karya0, line 30: Please change to `2 * 1024 * 1024`.

karya0 (approved):
Looks good to me modulo the stack size.

test/clone1.c
gc00: Done.

PR #963, opened 2022-03-01
Fixed a bug when creating file backing shared data

xuyao0127:


gc00 (approved):
I reviewed it directly with @xuyao0127.  LGTM.

The comment in the code explains the bug.  Historically, @gc00 provided a partial bug fix in 2018, but missed this remaining case.

PR #964, opened 2022-03-01
Fix e37c9b9ba for 'core' for gdb-dmtcp-utils

gc00:
This fixes a bug where it was improved in one direction, but stopped working on core files (using `(gdb) info proc mapping`).  This has also been tested by @xuyao0127 for core dumps for debugging MANA.  It worked there.

The PR is also being added to MANA:master .


xuyao0127 (approved):
Tested with core dump files. LGTM.

PR #965, opened 2022-03-02
Check for env var DMTCP_ABORT_ON_FAILURE

gc00:
* JASSERT had recognized DMTCP_ABORT_ON_FAILED_ASSERT
 * Change behavior to abort on all failures, change var. name

Previously, there was no easy way to ask DMTCP to abort on exit with failure.  This generalization helps us to debug DMTCP internal errors, and perhaps it's useful also when an end user hands DMTCP some improperly structured data, causing DMTCP to abort.

This PR should be easy to review.  It was motivated by a previous effort with @xuyao0127 to debug a DMTCP internal error.

### DIFF
--- a/src/dmtcp_launch.cpp
+++ b/src/dmtcp_launch.cpp
@@ -432,6 +432,14 @@ main(int argc, const char **argv)
 
   initializeJalib();
 
+  // FIXME:  This was changed in Mar., 2022.  We can remove this msg in 2 or 3 years.
+  if (getenv("DMTCP_ABORT_ON_FAILED_ASSERT")) {
+    JNOTE("\n\n*********************************************\n"
+              "* DMTCP_ABORT_ON_FAILED_ASSERT is obsolete. *\n"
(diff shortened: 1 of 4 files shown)


src/dmtcp_launch.cpp
```diff
@@ -419,6 +419,13 @@ processArgs(int *orig_argc, const char ***orig_argv)
 int
 main(int argc, const char **argv)
 {
+  // FIXME:  This was changed in Mar., 2022.  We can remove this msg in 2 or 3 years.
+  if (getenv("DMTCP_ABORT_ON_FAILED_ASSERT")) {
+    fprintf(stderr, "*********************************************\n"
```
karya0, line 424: Please use JTRACE here and for that you'll need to move it _after_ `initializeJalib()` call.

karya0 (approved):
LGTM modulo the fprintf->JTRACE/JNOTE change.

PR #967, opened 2022-03-04
Fixed a bug in realToVirtualPath

xuyao0127:
Previously, the return value is the same as the input. Also in many places, the return value of the realToVirtualPath is not used. 

For this family of functions, it passes in a pointer to char as a buffer. But there is no buflen. This design seems fragile. Sometimes a function overwrites the buffer without checking the buflen. The current code seems to depend on PATH_MAX, but it has no way to verify the path length.

@karya0, could you improve on this PR?

### DIFF
--- a/src/wrappers.cpp
+++ b/src/wrappers.cpp
@@ -141,7 +141,7 @@ realToVirtualPath(char *path)
 
   PluginManager::eventHook(DMTCP_EVENT_REAL_TO_VIRTUAL_PATH, &data);
 
-  return path;
+  return data.realToVirtualPath.path;
 }
 
 static int


xuyao0127:
@karya0 We still have the problem that last two commits from you and Rohan @rohgarg stop MANA from working. Could either of you make a new PR to fix it? 

On MIT's engaging (centos 7) cluster, MANA can't launch MPI programs. On Cori, MANA can't restart.

We need to fix that before we can use this PR in MANA.

src/wrappers.cpp
```diff
@@ -141,7 +141,7 @@ realToVirtualPath(char *path)
 
   PluginManager::eventHook(DMTCP_EVENT_REAL_TO_VIRTUAL_PATH, &data);
 
-  return path;
+  return data.realToVirtualPath.path;
```
karya0, line 144: For real to virtual conversion, the `path` is updated in-place so at the end of the function, `data.realtoVirtualPath.path` and `path` point to the same location.

karya0: Is there any particular reason for this change?

karya0:
Thanks for the suggestion around path length. We should add an extra parameter and validate length against the buffer size.

src/wrappers.cpp
xuyao0127: `data` is an in and out parameter so that it's possible to change where `data.realtoVirtualPath.path` points to. So we think it could be more stable this way. Also, the `realtoVirtialPah` returns the result with both the `path` and return value, but the return value is never used. Can we simplify this?

karya0: We don't use the return value right now, but I can imagine situations similar to `virtualToRealPath`. The idea was to keep the two signatures similar.

Re. `data` being an in/out parameter, it's a good point. Let's return `data.realToVirtualPath.path` instead of `path`.

karya0 (approved):

PR #970, opened 2022-03-11
Fix commit f01c3009: bug in writememoryarea

gc00:
The DMTCP type 'Area' assumes the invariant that `area.addr + area.size == area.endAddr`.  This invariant was not satisfied in original commit f01c3009 from 2014.  `area.size` was updated, but not `area.endAddr`.

@xuyao0127 encountered this bug when working on MANA.  Earlier, @rohgarg encountered this bug in the initial commit for MANA, but did only a local fix by switching from `area.endAddr` to `area.addr + area.size`.

### DIFF
--- a/src/writeckpt.cpp
+++ b/src/writeckpt.cpp
@@ -78,6 +78,12 @@ static void writememoryarea(int fd, Area *area, int stack_was_seen);
 
 static void remap_nscd_areas(const vector<ProcMapsArea> &areas);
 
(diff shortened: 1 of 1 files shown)


xuyao0127:
@gc00 I tested this commit and it can fix the incorrect `area.endAddr`. Commit eff574d also has this bug.

xuyao0127 (approved):
LGTM

karya0 (approved):
LGTM.

PR #973, opened 2022-03-15
Fix for commit ad90917: unable to reset ckpt timer

xuyao0127:
In commit ad90917, the Resume_resume barrier is removed. Therefore, the minimum state when the checkpoint is complete is no longer WorkerState:RUNNING, it's WorkerState:CHECKPOINTED. So releaseBarrier will never reset the checkpoint timer.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -499,7 +499,7 @@ DmtcpCoordinator::releaseBarrier(const string &barrier)
     broadcastMessage(DMT_BARRIER_RELEASED,
                      prevBarrier.length() + 1,
                      prevBarrier.c_str());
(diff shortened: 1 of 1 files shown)


gc00 (commented):
@xuyao0127,

PR #975 now contains the improved version of your original commit, `Fix: src and dest of mremap_move may overlaps`.  And I left you as the author, since it was a joint work.
PR #975 has now been reviewed and approved.  Please push it in.

After that, please rebase the current PR  against the new master.  After that, I can review and approve this PR (with just the one commit).  In general, I guess the two commits are unrelated.  It would have been better as separate PR's anyway.

xuyao0127:
@gc00 I just merged PR #973 and cleaned this PR.

gc00 (approved):
LGTM.

PR #974, opened 2022-03-20
Do not call futex_wake on MutexUnlock if no waiters queued.

karya0:


src/mutex.cpp
```diff
@@ -4,51 +4,55 @@
...
-    if (waitVal == 0) {
-      // We successfully acquired the lock.
-      break;
+  do {
+    int oldval = __sync_val_compare_and_swap(&mutex->futex,
+                                              LOCK_ACQUIRED,
+                                              LOCK_ACQUIRED_WAITERS_QUEUED);
+    if (oldval != LOCK_FREE) {
+      int s = futex_wait(&mutex->futex, LOCK_ACQUIRED_WAITERS_QUEUED);
+      JASSERT(s != -1 || errno == EAGAIN)(JASSERT_ERRNO);
```
gc00, line 48: I believe you want to add a test for EINTR: `(s != -1 || errno == EAGAIN || errno == EINTR)`.

src/mutex.cpp
```diff
@@ -4,51 +4,55 @@
...
-    uint32_t waitVal =
-      __sync_val_compare_and_swap((uint32_t *)&mutex->owner, 0, owner);
-
-    if (waitVal == 0) {
-      // We successfully acquired the lock.
-      break;
+  do {
+    int oldval = __sync_val_compare_and_swap(&mutex->futex,
+                                              LOCK_ACQUIRED,
+                                              LOCK_ACQUIRED_WAITERS_QUEUED);
```
gc00, line 45: QUESTION: Suppose one thread is calling DmtcpMutexLock and another thread is releasing the lock, at the same time.  So, the first thread might see `oldval == LOCK_ACQUIRED`, and the second thread might then release the lock (putting it into state `LOCK_FREE`).  Our local variable `oldval` still thinks that someone else is holding the lock, and so it does `futex_wait`, even though the lock is now in state `LOCK_FREE`.

src/mutex.cpp
```diff
@@ -4,51 +4,55 @@
...
   }
 
-  while (1) {
-    uint32_t waitVal =
-      __sync_val_compare_and_swap((uint32_t *)&mutex->owner, 0, owner);
-
-    if (waitVal == 0) {
-      // We successfully acquired the lock.
-      break;
+  do {
```
gc00, line 42: A comment above the `do` would help.  Readers can figure out the code, but it takes time.  It's better to show the intention in the comment, and a reader will either temporarily trust that it's correct, or will validate against the spec in the comment.

I think the intention here is:
```
// If someone else is holding the lock (lock in state LOCK_ACQUIRED), then we register ourselves
// for waiting on the lock by putting it in state LOCK_ACQUIRED_WAITERS_QUEUED, and then
// calling futex_wait().
// When we wake up from this, we test if the lock is in state LOCK_FREE, and if so, we set it for LOCK_ACQUIRED_WAITERS_QUEUED, and we return, holding this lock.
// If the lock was originally in state LOCK_FREE, then we set the lock to state LOCK_ACQUIRED_WAITERS_QUEUED.
// We succeed and return only by leaving the lock in state LOCK_ACQUIRED_WAITERS_QUEUED.  Else, we loop.
```
QUESTION:  Suppose we are the only one who acquired the lock.  Shouldn't we then be allowed to return by leaving the lock in state LOCK_ACQUIRED?

src/mutex.cpp
```diff
@@ -99,8 +111,14 @@ DmtcpMutexUnlock(DmtcpMutex *mutex)
...
     mutex->owner = 0;
 
-    // TODO(Kapil): Call futex_wake only if there are futex waiters.
-    JASSERT(futex_wake((uint32_t*) &mutex->owner, 1) != -1) (JASSERT_ERRNO);
+    if (__sync_bool_compare_and_swap(&mutex->futex, LOCK_ACQUIRED, LOCK_FREE)) {
+      // No need to call futex_wake as we don't have any waiters.
+    } else {
+      JASSERT(__sync_bool_compare_and_swap(&mutex->futex,
+                                           LOCK_ACQUIRED_WAITERS_QUEUED,
+                                           LOCK_FREE));
```
gc00, line 119: QUESTION:  Suppose we succeed in setting the lock to LOCK_FREE, above.  Then two other threads come in.  One acquires the lock and sets the state to LOCK_ACQUIRED.  A second thread comes in and sets the state to LOCK_ACQUIRED_WAITERS_QUEUED and does `futex_wait()`.  It appears that we are now going to send a _spurious_ wakeup to the thread that is waiting on the lock.
&nbsp;&nbsp; That's okay, if your intention is that the thread receiving the spurious wakeup will simply go around the loop, and return to waiting.  But in that case, it would be nice to add a comment here, such as:
```
// After setting the lock to LOCK_FREE, two other threads may have tried to grab the lock.
// One succeeds and one is waiting.  In this case, we are doing a spurious wakeup.
// The logic for handling a spurious wakeup is contained in the function 'DmtcpMutexLock()'.

**Review by gc00 — COMMENTED** [review]
@karya0,  I'm concerned especially about my first two questions, that are inline in the code, for `DmtcpMutexLock()`.  If need be, we can call each other and analyze this interactively.

My worry is the following:
&nbsp;&nbsp;    For many cases, including MANA, we are using this logic to view the checkpoint thread as asking for an exclusive lock, while each use thread is at most asking for a shared lock (via `DISABLE_CKPT()`).  So, we have one writer and multiple readers.
 &nbsp;&nbsp;  This situation has a particularly simple, efficient, and easy-to-analyze solution  Because you are trying to handle also the case of multiple writers, you may be making this either overly complicated, or possibly even inefficient.
 &nbsp;&nbsp;   I'm happy to be corrected on this.  :-)

File: src/mutex.cpp (hunk 4 above)
Comment (gc00) [reply src/mutex.cpp:119]: And I'll add one further comment on the simplicity of a rslock with readers preferred.  This is a quote from:
`glibc-2.32/nptl/pthread_rwlock_common.c`:
>   This lock is built so that workloads of mostly readers can be executed with           low runtime overheads.  This matches that the default kind of the lock is             PTHREAD_RWLOCK_PREFER_READER_NP.  Acquiring a read lock requires a single             atomic addition if the lock is or was previously acquired by other                    readers; releasing the lock is a single CAS if there are no concurrent                writers.

And of course, `CAS`, above, stands for Compare-And-Swap.

File: src/mutex.cpp [hunk 5]
```diff
@@ -4,51 +4,55 @@
...
-      __sync_val_compare_and_swap((uint32_t *)&mutex->owner, 0, owner);
-
-    if (waitVal == 0) {
-      // We successfully acquired the lock.
-      break;
+  do {
+    int oldval = __sync_val_compare_and_swap(&mutex->futex,
+                                              LOCK_ACQUIRED,
+                                              LOCK_ACQUIRED_WAITERS_QUEUED);
+    if (oldval != LOCK_FREE) {
```
gc00, line 46: Perhaps we could add a comment here:
```
        // futex_wait() returns immediately if the lock is not in state LOCK_ACQUIRED_WAITERS_MAY_BE_QUEUED

**Review by gc00 — APPROVED** [review]
1, Let's change `LOCK_ACQUIRED_WAITERS_QUEUED` to `LOCK_ACQUIRED_WAITERS_MAY_BE_QUEUED`, as we discussed.
2. Should we change `futex_wait()` to `futex_wait_if()`?  I could go either way.  The new name helped me in understanding the code.  (We wait on the futex only if it has the value given by the second argument.)
3. There are minor inline comments to address (extra comments in code, EINTR, indent spaces).  Please review the inline comments.

Otherwise, everything looks good.

File: src/mutex.cpp [hunk 6]
```diff
@@ -4,51 +4,55 @@
...
+      int s = futex_wait(&mutex->futex, LOCK_ACQUIRED_WAITERS_QUEUED);
+      JASSERT(s != -1 || errno == EAGAIN)(JASSERT_ERRNO);
     }
+  } while (__sync_val_compare_and_swap(&mutex->futex,
+                                        LOCK_FREE,
+                                        LOCK_ACQUIRED_WAITERS_QUEUED)
+          != LOCK_FREE);
```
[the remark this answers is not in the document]
karya0: Fixed

src/mutex.cpp
karya0: Fixed.

src/mutex.cpp
karya0: I think futex_wait is better. Futex_wait_if might indicate some additional processing we might be doing.

src/mutex.cpp
karya0: Fixed.

src/mutex.cpp
karya0: The comment for lock is sufficient. We won't have spurious wakeups, so no need for this comment here.

PR #975, opened 2022-03-23
Fix: src and dest of mremap_move may overlap in mtcp_restart.c

xuyao0127:
Newer Linux kernels may place vdso and vvar in rather arbitrary locations.  On restart, we must mremap the vdso/vvar of the current mtcp_restart.c to the vdso/vvar address from the original pre-checkpoint process.  But if the current vdso/vvar overlaps with the mremap destination, then the mremap fails.

This commit mremap's to a temporary mmap'ed memory region that doesn't overlap, and then it mremap's again to the final destination.

* In addition, there is a second commit that changes `-Wl,-Ttext-segment=11100000` to `-Wl,-Ttext-segment=11200000`.  The address must be a multiple of the pagesize, and a 2M hugepage is `0x200000`.  So, this is needed to be compatible with hugepages.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -901,7 +901,7 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
     // Since vdso will use randomized addresses (unlike the standard practice
     // for vsyscall), this implies that kernel calls on __x86__ can go through
     // randomized addresses, and so they need special treatment.
-    void *vdso = mmap_fixed_noreplace(statingVdsoStart, vdsoEnd - vdsoStart,
+    void *vdso = mmap_fixed_noreplace(stagingVdsoStart, vdsoEnd - vdsoStart,
                                       PROT_EXEC | PROT_WRITE | PROT_READ,
                                       MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED,
                                       -1, 0);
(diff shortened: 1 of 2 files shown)


xuyao0127:
Currently dmtcp `make -j check` fails on the second checkpoint. MANA fails to restart.

xuyao0127:
The bug in the previous commit is fixed now. It was the staging area overlaps with the new vvar region and the overlapping region is unmapped when releasing the staging area. Tested on Cori.

xuyao0127:
Found another bug. This fix only works when the old and new vvar/vdso areas overlap. If they are not overlapping, this could fail because the staging area overlaps with the new vvar/vdso.

gc00:
Hi Yao,
    Thanks!  I'm sure there is a clean fix for this.  I'll think about it overnight.
Best,
- Gene

xuyao0127:
Fixed and tested on Cori with and without MANA.

src/mtcp/mtcp_restart.c
```diff
@@ -867,7 +901,7 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
     // Since vdso will use randomized addresses (unlike the standard practice
     // for vsyscall), this implies that kernel calls on __x86__ can go through
     // randomized addresses, and so they need special treatment.
-    void *vdso = mmap_fixed_noreplace(vdsoStart, vdsoEnd - vdsoStart,
+    void *vdso = mmap_fixed_noreplace(statingVdsoStart, vdsoEnd - vdsoStart,
```
JainTwinkle, line 904: statingVdsoStart -> stagingVdsoStart ?

xuyao0127: Yes, it's I typo. Thank you.

gc00 (approved):
@xuyao0127, I've now fixed the minor typo (which had been my fault).  I've left you as the author of these commits and the PR.  Thanks for this joint effort, @xuyao0127.

Everything looks good now.   @xuyao0127, could you push in this PR?

PR #976, opened 2022-04-04
Changes from feature/mana

xuyao0127:
@karya0 @gc00 There are three commits in the feature/mana branch now. We need to take a look at some of these commits before we can take them to the DMTCP.
1. 1df553b97f60608d46b40839b4fd793ed2ddc3b4: This commit added more keywords in the `FileConnList::prepareShmList()` function to skip some MANA related regions. I remember Kapil said he will work on a way that plugins can add more keywords without modifying DMTCP.
2. 08e4c8d2469f9bbe986744d94bafd2071c39847a: This commit revert a previous commit 764a55f917c7a4abaa0183cada8b9f1934bfab99 that added a compile option that is no longer needed.
3. 4bd4fa6dc3dbe9ceaa7ad9669d3fab7eaecb99dc: This commit is part of a commit in MANA's master branch that renames `libmpidummy.so` to `libmpistub.so`. `dmtcp_launch.cpp` is also modified because if dmtcp_launch is run with the `--mpi` option, the function `setLDLibraryPathForMPI()` will be called to setup the `LD_LIBRARY_PATH` environment variable. If DMTCP was not compiled with MANA, but a user added the `--mpi` option, should we print some warning messages? Like the restart plugin issue I mentioned in our previous meeting, how can we distinguish if DMTCP was built with or without MANA? What should we do if a wrong build or flag is used?

### DIFF
--- a/src/dmtcp_launch.cpp
+++ b/src/dmtcp_launch.cpp
@@ -40,7 +40,6 @@ static int testJava(const char **argv);
 static bool testSetuid(const char *filename);
 static void testStaticallyLinked(const char *filename);
 static bool testScreen(const char **argv, const char ***newArgv);
-static void setLDLibraryPathForMPI(bool is32bitElf);
 static void setLDPreloadLibs(bool is32bitElf);
 
 // gcc-4.3.4 -Wformat=2 issues false positives for warnings unless the format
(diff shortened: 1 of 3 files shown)


gc00:
@karya0 , I discussed this with Yao.  Some items (especially item 3 of the description) requires input from you.  Either github comments or a joint meeting would be fine.

karya0 (approved):
LGTM.

PR #977, opened 2022-04-21
Cherry-pick 2 fixes for `gdb-dmtcp-utils`

jungan:
- Doesn't work with coredump file with getpid() == 1
- Minor fix from MANA 0f8b15eb8


gc00 (approved):
LGTM.

PR #980, opened 2022-05-02
Disabled vfork tests in autotest.py.

karya0:
Minor update to .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,8 @@ test/testconfig.py
 
 ckpt_*.dmtcp
 dmtcp_restart_script*
+
+am--include-marker
+\#
+*.tmp
+.vscode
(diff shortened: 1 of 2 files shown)


JainTwinkle:
Could you provide some reasoning behind disabling tests?

karya0:
Vfork support and tests are broken currently and we intend to fix it later
when we have someone using vfork.

JainTwinkle:
Do we have any existing Github issue or bug to reference to show what exactly is broken? If not, then could you please create one? We can label it as TODO to ensure that we address the issue in the future.

karya0:
> Do we have any existing Github issue or bug to reference to show what exactly is broken? If not, then could you please create one? We can label it as TODO to ensure that we address the issue in the future.

I created https://github.com/dmtcp/dmtcp/issues/983. Since we don't have a precedence for tracking github issues in code, I'll skip the TODO for now.

JainTwinkle (approved):
Thanks for creating #983; approving this one.

PR #982, opened 2022-05-02
Fixed area->addr typecasting for JTRACE.

karya0:
The segfault occurred because area.addr is `char *` and JTRACE treats `char *` pointers as string.


JainTwinkle (approved):
LGTM!

PR #984, opened 2022-05-03
Fixed Sys V Semaphore bug related to delete.

karya0:
Replaces a dynamically allocated array with a vector.

The delete operation in destructor should have been JALIB_FREE instead.

We don't need to use a dynamic array anyway, since a vector works just as well.


xuyao0127 (approved):
LGTM

PR #987, opened 2022-05-12
Bugfix - abort() is being called before jalib_init() because of DMTCP_FAIL_RC macro evaluation.

tarunsmalviya:
The definition of DMTCP_FAIL_RC is changed in [this](https://github.com/dmtcp/dmtcp/commit/14932be2ab71fc0bc5b700b618c3f12e0d128b62) commit. If getenv(DMTCP_ABORT_ON_FAILURE) returns 1, DMTCP_FAIL_RC will call abort() immediately. However, DMTCP_FAIL_RC is used as the last parameter of the function jalib_init. Therefore, before calling jalib_init, DMTCP_FAIL_RC is expanded and evaluated in order to get the value of the parameter. Then because the environment variable DMTCP_ABORT_ON_FAILURE is defined as 1, the program will call abort() immediately.

### DIFF
--- a/include/dmtcp.h
+++ b/include/dmtcp.h
@@ -690,12 +690,15 @@ uint64_t dmtcp_dlsym_lib_fnc_offset(const char *libname, const char *symbol);
  * TYPICAL USAGE:  exit(DMTCP_FAIL_RC)
  * Use this to distinguish DMTCP failing versus the target application failing.
  */
+#define DMTCP_FAIL_RC_PARAM                                      \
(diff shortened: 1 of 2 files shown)


tarunsmalviya:
I tested this branch on Cori. All tests were passed except one, i.e, `emacs`.

Test Screenshot 1
Test Screenshot 1

Test Screenshot 2
Test Screenshot 2

xuyao0127 (approved):
LGTM. jalib_init seems to be the only place that DMTCP_FAIL_RC is used as a parameter.

PR #989, opened 2022-05-18
Cherry-pick dmtcp change from commit a1583a "Deterministic p2p".

dahongli:
This change cherry-pick the warning for checkpoint file size change from commit a1583a "Deterministic p2p".


gc00 (commented):
@dahongli, can you write a short comment describing the scenario where this should only be a warning, because there is a valid use case?  Thanks.

dahongli:
> @dahongli, can you write a short comment describing the scenario where this should only be a warning, because there is a valid use case? Thanks.

Yes, I have added a comment describing the scenario. Thanks Gene for your review!

gc00:
the its size saved in checkpoint ...:
the its -> its
in checkpoint -> in the checkpoint
After fixing that LGTM.

dahongli:
> the its size saved in checkpoint ...:
> the its -> its
> in checkpoint -> in the checkpoint
> After fixing that LGTM.

Done.

gc00 (approved):
LGTM.

PR #991, opened 2022-05-23
Add GitHub Actions file.

karya0:
Testing GitHub Actions CI/CD pipeline.


JainTwinkle (approved):
Could you add a small comment on why we have disabled the mknod test?
Otherwise, LGTM!

karya0:
Mknod has been flaky on certain platforms so we disabled it a while back but then re-enabled it or something like that. IIRC, part of the problem is with the less-than-strict API.

Issue #990, opened 2022-05-20
DMTCP: current break != saved break


cyssi-cb:
Hello, 

I've been trying out DMTCP, version 3.0.0 on Ubuntu 22.04.

After installation "make check" will pass all the checkpoints, but will fail for every single restore. 
>$ make check 
dmtcp1         ckpt:PASSED; rstr:FAILED (first process rec'd signal 11); retry:
***** Copied checkpoint images to /tmp/dmtcp-cyssi@Cyssi-ASUS-Laptop/dmtcp-autotest-248999978
FAILED
               root-pids: [56088] msg: restart error, 1 expected, 0 found, running=0
dmtcp2         ckpt:PASSED; rstr:FAILED (first process rec'd signal 11); retry:
***** Copied checkpoint images to /tmp/dmtcp-cyssi@Cyssi-ASUS-Laptop/dmtcp-autotest-248999978
FAILED
               root-pids: [56102] msg: restart error, 1 expected, 0 found, running=0
dmtcp3         ckpt:PASSED; rstr:FAILED (first process rec'd signal 11); retry:
***** Copied checkpoint images to /tmp/dmtcp-cyssi@Cyssi-ASUS-Laptop/dmtcp-autotest-248999978
FAILED

If I'm running DMTCP on my own little test program (just a c program counting up numbers and printing them to stdout) I'm able to create a checkpoint and a restart script, but running the restart script will fail with the output below: 
>$ ./dmtcp_restart_script.sh
[56273] ERROR at dmtcp_restart.cpp:859 in main; REASON='JASSERT(stat(restorename.c_str(), &buf) != -1) failed'
dmtcp_restart (56273): Terminating...

If I try to restart it using dmtcp_restart [.dmtcp file] it'll terminate due to segmentation fault, as shown below. 
>$ dmtcp_restart ckpt_dmtcp_c_test_d7cac449be66ffb-40000-40d59db424b.dmtcp
[56283] mtcp_restart.c:347 restore_brk:
  error: new/current break (0x11213000) != saved break (0x557a1ba36000)
Segmentation fault (core dumped)

Any help will be highly appreciated, thank you already in advance!

gc00:
@JainTwinkle or @karya0 , do you have time to take a look at DMTCP under Ubuntu 22.04?  It seems that there may be a small incompatibility that we will need to fix.  Thanks.

karya0:
I’ll take a look.

karya0:
[...]

PR #992, opened 2022-05-24
Use -static when compiling mtcp_restart.

karya0:
Without -static, gcc generates a dynamically linked mtcp_restart on
Ubuntu 22.04.

Addresses #990.


xuyao0127 (approved):
LGTM

PR #998, opened 2022-06-09
Moved static members from dmtcpplugin to ProcessInfo.

karya0:
Initializing static members relied on untimely usage of new, causing deadlock due to malloc wrappers, etc..

### DIFF
--- a/src/dmtcpplugin.cpp
+++ b/src/dmtcpplugin.cpp
@@ -160,10 +160,7 @@ dmtcp_get_ckpt_signal(void)
 EXTERNC const char *
 dmtcp_get_tmpdir(void)
 {
-  static char tmpdir[PATH_MAX];
-
-  JASSERT(SharedData::getTmpDir(tmpdir, sizeof(tmpdir)) != NULL);
-  return tmpdir;
+  return SharedData::getTmpDir();
 }
 
 // EXTERNC void dmtcp_set_tmpdir(const char* dir)
@@ -176,8 +173,7 @@ dmtcp_get_tmpdir(void)
 EXTERNC const char *
 dmtcp_get_ckpt_dir()
 {
-  static string *tmpdir = new string(ProcessInfo::instance().getCkptDir());
-  return tmpdir->c_str();
+  return ProcessInfo::instance().getCkptDir().c_str();
 }
 
 EXTERNC int
@@ -198,17 +194,13 @@ dmtcp_set_ckpt_file(const char *filename)
 EXTERNC const char *
 dmtcp_get_ckpt_filename(void)
 {
-  static string *filename =
-    new string(ProcessInfo::instance().getCkptFilename());
-  return filename->c_str();
+  return ProcessInfo::instance().getCkptFilename().c_str();
 }
 
 EXTERNC const char *
 dmtcp_get_ckpt_files_subdir(void)
 {
-  static string *tmpdir =
-    new string(ProcessInfo::instance().getCkptFilesSubDir());
-  return tmpdir->c_str();
+  return ProcessInfo::instance().getCkptFilesSubDir().c_str();
 }
 
 EXTERNC int
@@ -232,33 +224,25 @@ dmtcp_get_executable_path(void)
 EXTERNC const char *
 dmtcp_get_uniquepid_str(void)
 {
-  static string *uniquepid_str =
-    new string(UniquePid::ThisProcess(true).toString());
-  return uniquepid_str->c_str();
+  return ProcessInfo::instance().upidStr().c_str();
 }
 
 EXTERNC DmtcpUniqueProcessId
 dmtcp_get_uniquepid(void)
 {
-  return UniquePid::ThisProcess().upid();
+  return ProcessInfo::instance().upid().upid();
 }
 
 EXTERNC DmtcpUniqueProcessId
 dmtcp_get_computation_id(void)
 {
-  return SharedData::getCompId();
+  return ProcessInfo::instance().compGroup().upid();
 }
 
 EXTERNC const char *
 dmtcp_get_computation_id_str(void)
 {
-  static string *compid_str = NULL;
-
-  if (compid_str == NULL) {
-    UniquePid compId = SharedData::getCompId();
-    compid_str = new string(compId.toString());
-  }
-  return compid_str->c_str();
+  return ProcessInfo::instance().compGroupStr().c_str();
 }
 
 EXTERNC DmtcpUniqueProcessId
--- a/src/processinfo.h
+++ b/src/processinfo.h
@@ -70,10 +70,6 @@ class ProcessInfo
 
     void serialize(jalib::JBinarySerializer &o);
 
-    UniquePid compGroup() { return _compGroup; }
-
-    void compGroup(UniquePid cg) { _compGroup = cg; }
-
     uint32_t numPeers() { return _numPeers; }
 
     void numPeers(uint32_t np) { _numPeers = np; }
@@ -106,9 +102,40 @@ class ProcessInfo
 
     const string &hostname() const { return _hostname; }
 
-    const UniquePid &upid() const { return _upid; }
+    const UniquePid &upid() {
+      // Temporary fix until we remove the static members from UniquePid.cpp.
+      if (_upid == UniquePid()) {
+        _upid = UniquePid::ThisProcess(true);
+      }
+      return _upid;
+    }
 
-    const UniquePid &uppid() const { return _uppid; }
+    const string &upidStr() {
+      if (_upidStr.empty()) {
+        _upidStr = upid().toString();
+      }
+      return _upidStr;
+    }
+
+    const UniquePid &uppid() {
+      // Temporary fix until we remove the static members from UniquePid.cpp.
+      if (_uppid == UniquePid()) {
+        _uppid = UniquePid::ParentProcess();
+      }
+      return _uppid;
+    }
+
+    UniquePid compGroup() const { return _compGroup; }
+
+    const string &compGroupStr() {
+      if (_compGroupStr.empty()) {
+        _compGroupStr = _compGroup.toString();
+      }
+      return _compGroupStr;
+    }
+
+    void compGroup(UniquePid cg) { _compGroup = cg; }
+
 
     bool isOrphan() const { return _ppid == 1; }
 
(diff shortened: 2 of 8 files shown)


JainTwinkle:
Hi @karya0 
On my machine(openSUSE), two tests (`frisbee` and `nocheckpoint`) are consistently failing.
With `frisbee`, I'm seeing the following error:
```
[42000] WARNING at socketconnection.cpp:720 in postRestart; REASON='JWARNING(_real_bind(_fds[0], (sockaddr *)&_bindAddr, _bindAddrlen) == 0) failed'
     (strerror((*__errno_location ()))) = Address already in use
     id() = 0-0-0(99004)
Message: Bind failed.
[41000] WARNING at socketconnection.cpp:720 in postRestart; REASON='JWARNING(_real_bind(_fds[0], (sockaddr *)&_bindAddr, _bindAddrlen) == 0) failed'
     (strerror((*__errno_location ()))) = Address already in use
     id() = 0-0-0(99004)
Message: Bind failed.
```
And, with `nocheckpoint`, I'm seeing the following error:
```
'unexpected number of checkpoint files, 1 procs, 2 files'
```
I am yet to debug further. I'll update this thread as I get more information.

karya0:
I did a force push with a fix for the frisbee bug. Can you do a pull and
try again?

JainTwinkle:
Thanks, @karya0! All tests now pass on my machine. I'll review the PR in a while.

src/processinfo.h
```diff
@@ -25,6 +25,7 @@
 #include <sys/types.h>
 #include "../jalib/jalloc.h"
 #include "uniquepid.h"
+#include "uniquepid.h"
```
JainTwinkle, line 28: Included twice

contrib/snooze/snooze.c
```diff
@@ -27,7 +27,6 @@ pre_ckpt()
            dmtcp_get_generation());
 
   dmtcp_set_ckpt_dir(newCkptDir);
-  dmtcp_set_coord_ckpt_dir(newCkptDir);
```
JainTwinkle, line 30: Do we not need to update the `ckpt_dir` info in the coordinator anymore?

include/dmtcp.h
```diff
@@ -281,41 +285,6 @@ int dmtcp_enable_ckpt(void) __attribute__((weak));
...
- *     libc_execvp_ptr = dmtcp_get_libc_addr("execvp");
- *   }
- *   if (libc_execvp_ptr != DMTCP_NOT_PRESENT) {
- *     // Don't preload DMTCP libraries when you exec to a new program.
- *     char *ld_preload = getenv("LD_PRELOAD");
- *     ld_preload[0] = '\0';
- *     (*libc_execvp_ptr)(...);
- *   }
- */
-void * dmtcp_get_libc_addr(const char* libc_fnc) __attribute__((weak));
```
JainTwinkle, line 311: Do we not need `dmtcp_get_libc_addr` function at all?
If we have to, I'd suggest either putting it under `#if 0 ...  #endif` or removing the remaining references (in `src/execwrappers.cpp` and `src/dmtcpnohijackstubs.cpp`) in DMTCP. What do you think @gc00?

src/coordinatorapi.cpp
```diff
@@ -351,40 +351,6 @@ connectAndSendUserCommand(char c,
...
-}
-
-void
-updateCoordCkptDir(const char *dir)
-{
-  if (noCoordinator()) {
-    return;
-  }
-  JASSERT(dir != NULL);
-  DmtcpMessage msg(DMT_UPDATE_CKPT_DIR);
```
JainTwinkle, line 384: Again, do we not need to get and set `ckpt_dir` info in the coordinator anymore?

src/dmtcpplugin.cpp
```diff
@@ -39,7 +39,6 @@
 #undef dmtcp_get_local_status
 #undef dmtcp_get_uniquepid_str
 #undef dmtcp_get_ckpt_filename
-#undef dmtcp_set_coord_ckpt_dir
 #undef dmtcp_get_coord_ckpt_dir
```
JainTwinkle, line 42: If we decide to remove `dmtcp_get_coord_ckpt_dir` then let's remove this line as well.

src/dmtcpplugin.cpp
```diff
@@ -216,81 +195,13 @@ dmtcp_set_ckpt_file(const char *filename)
...
 EXTERNC const char *
 dmtcp_get_ckpt_files_subdir(void)
 {
-  static string *tmpdir =
-    new string(ProcessInfo::instance().getCkptFilesSubDir());
-  return tmpdir->c_str();
-}
-
-EXTERNC void
-dmtcp_get_bin_dir(char dmtcp_bin_dir[]) {
```
JainTwinkle, line 233: @gc00 was the one who put all the functions related to `dmtcp_get_libc_addr` in #837. I would like to get his opinion on removing these functions. We can move them or comment out temporarily.

src/processinfo.h
```diff
@@ -106,9 +103,40 @@ class ProcessInfo
 
     const string &hostname() const { return _hostname; }
 
-    const UniquePid &upid() const { return _upid; }
+    const UniquePid &upid() {
+      // Temporary fix until we remove the static members from UniquePid.cpp.
```
JainTwinkle, line 106: Minor: `UniquePid` is fine for the clarity but `uniquepid.cpp` is the actual name.

src/processinfo.h
```diff
@@ -106,9 +103,40 @@ class ProcessInfo
...
+
+    const string &upidStr() {
+      if (_upidStr.empty()) {
+        _upidStr = upid().toString();
+      }
+      return _upidStr;
+    }
+
+    const UniquePid &uppid() {
+      // Temporary fix until we remove the static members from UniquePid.cpp.
```
JainTwinkle, line 121: `UniquePid` -> `uniquepid`

include/dmtcp.h
gc00: I think we still need dmtcp_get_libc_addr.  We're not actively using it right now.  But we will need it in a scenario where several plugins (or a DMTCP plugin and the target application) are interposing on a call in glibc.  At some point, one of these plugins will want to directly call the glibc low-level function, without fear of the target application or another plugin interposing on it.
@karya0 ,  If you don't want it to interfere for now, then please enclose it in `#if 0 ... #endif`.  But this should not be dead code.

src/coordinatorapi.cpp
gc00: I agree with @JainTwinkle .  Let's keep this functionality for now.

If we want to remove some functionality, then that should be the topic of a separate PR -- not combined with this one.

contrib/snooze/snooze.c
gc00: I agree with @JainTwinkle .  If we want to discuss removing some existing functionality, let's make that a separate PR, and not mix it with this one.

src/coordinatorapi.h
```diff
@@ -79,9 +79,6 @@ char *connectAndSendUserCommand(char c,
                                 int *isRunning = NULL,
                                 int *ckptInterval = NULL);
 
-void updateCoordCkptDir(const char *dir);
-string getCoordCkptDir(void);
-
```
gc00, line 84: As before, any discussion of removing some functionality should be a separate PR.

src/dmtcpplugin.cpp
```diff
@@ -191,22 +186,6 @@ dmtcp_set_ckpt_dir(const char *dir)
...
-
-EXTERNC int
-dmtcp_set_coord_ckpt_dir(const char *dir)
-{
-  if (dir != NULL) {
-    CoordinatorAPI::updateCoordCkptDir(dir);
-  }
-  return DMTCP_IS_PRESENT;
-}
-
```
gc00, line 209: Same comment as above.

src/dmtcpplugin.cpp
gc00: See my earlier comment on the reasons for keeping `dmtcp_get_libc_addr`.  Sometimes, we really want the low-level glibc fnc, with no interposition.  We either need this or an equivalent mechanmism.

src/dmtcpplugin.cpp
```diff
@@ -216,81 +195,13 @@ dmtcp_set_ckpt_file(const char *filename)
...
-    close(0);
-    dup2(pipefd[1], 1); // command will write offset to stdout (fd: 1)
-    // char *ld_preload = getenv("LD_PRELOAD");
-    // char ld_preload_first_char = ld_preload[0]; // Unused; see below;
-    // ld_preload[0] = '\0'; // No LD_PRELOAD on next program.
-    environ[0] = NULL;
-    execvpe(command[0], command, environ);
-    // ld_preload[0] = ld_preload_first_char; // Not needed; never reached.
-    exit(0); // to satisfy the compiler
-  }
```
gc00, line 293: Same comment as earlier.  If we're removing functionality, that should be a separate PR, unrelated to this one.

karya0:
@gc00 @JainTwinkle : Should we do a quick phone call to go over these points? I think it'll be easier than a bunch of emails.

Here is a high-level reasoning behind removing those functions:

**dmtcp_set_coord_ckpt_dir** -- not used by any known plugin, etc. In fact, this is a rather dangerous call since it allows a user thread to send a message to the coordinator which is something we don't support yet as it interferes with the ckpt threads' socket communication with the coordinator.

**dmtcp_get_libc_addr** -- this also is dead code with no active callers. I am not sure if I am sold on the notion of allowing a third-party plugin to bypass other plugins/dmtcp to call libc directly. Do we have a good usecase yet? If not, I'd suggest to remove it from dmtcp.h and debate whether we want to allow (and have a need for) first-party plugins like MANA, etc., to call libc directly. My concern is that it might modify some state that will be hard to capture/debug later.

If we have a good use case, we should reimplement this function to use the @gc00's new dlsym functionality instead of the fork+exec model. The latter is troublesome due to mismatch between libc version loaded in the process vs the one that's going to be loaded in the child process -- we have seen this happen in past when ckpt images were moved from one machine to another with slight mismatches in libc versions.

JainTwinkle (commented):
Thanks, @karya0, for splitting the PR. Please update this PR to reflect only the changes that are not in #999. Otherwise, LGTM!

JainTwinkle (approved):

PR #999, opened 2022-06-10
Removed some deadcode.

karya0:
This is a split from #998.


JainTwinkle (approved):
@gc00, I'm approving this PR in advance. We can add these functionalities back to DMTCP if required.

gc00:
Hi @karya0 and @JainTwinkle ,
It's okay to remove this code for now.

But in the future, I suspect that I'll want to bring some of this code back.  There are complex cases where some strapper function wants to call, for example, fork for a short-lived process, and doing the extra stuff in our current fork wrappers really does get in the way.  But we can jointly discuss the right architecture for that on the future, when we need it for some case.

PR #1002, opened 2022-06-20
Save ckpt thread's TLS during preckpt.

karya0:
TLS state (fs/gs registers, etc.) for a thread can change as a result of dlopen calls. For all user-threads, we already query the TLS state during the pre-ckpt phase but not doing so for ckpt-thread causes an invalid TLS state to be saved during checkpoint.

In this fix, we simply query ckpt-thread's TLS state during preckpt phase and restore it as usual during post-restart processing.


JainTwinkle (approved):
Thanks, @karya0!
Could you add the same comment to the code as in the PR description? It would be a good reminder for anyone who wants to remove the repetitious save during checkpoints.

I'm approving in advance, but I have a question. Now that we save the TLS state of checkpoint-thread during ckpt, can we remove saving the TLS state during creation (L382)? Is it redundant, or is it important?

karya0:
Good catch. I removed the redundant call and also moved SaveSigState inside the loop along with a comment.

PR #1003, opened 2022-06-25
Allow application to munmap SysV SHM area.

karya0:
Application may call shmat at some area followed by a munmap call. While unconventional, this is allowed, and we should simply remove that area from our tracking map.


karya0:
@JainTwinkle : This is now ready to go. Can you please approve it? I'll then update mana branch as well.

JainTwinkle (approved):
LGTM.

PR #1005, opened 2022-07-02
Always use static TLS TID offset.

### LINKED ISSUE #994: Warning: Couldn't find offsets of tid/pid in thread_area
With `gcc-9`, dmtcp_launch throws the following warning:
```
$ ../bin/dmtcp_launch ./dmtcp1
[40000] WARNING at tls.cpp:256 in TLSInfo_GetTidOffset; REASON='JWARNING(false) failed'
     tid_pid.tid = 1442
Message: Couldn't find offsets of tid/pid in thread_area.
  Now relying on the value determined using the
  glibc version with which DMTCP was compiled.
[40000] WARNING at tls.cpp:256 in TLSInfo_GetTidOffset; REASON='JWARNING(false) failed'
     tid_pid.tid = 1442
Message: Couldn't find offsets of tid/pid in thread_area.
  Now relying on the value determined using the
  glibc version with which DMTCP was compiled.
[40000] WARNING at tls.cpp:256 in TLSInfo_GetTidOffset; REASON='JWARNING(false) failed'
     tid_pid.tid = 1442
Message: Couldn't find offsets of tid/pid in thread_area.
  Now relying on the value determined using the
  glibc version with which DMTCP was compiled.
[40000] WARNING at tls.cpp:256 in TLSInfo_GetTidOffset; REASON='JWARNING(false) failed'
     tid_pid.tid = 1448
Message: Couldn't find offsets of tid/pid in thread_area.
  Now relying on the value determined using the
  glibc version with which DMTCP was compiled.
```

System information:
```
gcc (GCC) 9.3.1 20200408 (Red Hat 9.3.1-2)
ldd (GNU libc) 2.17
OS: CentOs 7.9.2009
```

The comment at [tls.cpp#L249](https://github.com/dmtcp/dmtcp/blob/138dbe086d147b0255a4b3e03a53f67b5dd4d0ba/src/tls.cpp#L249) mentions about glibc-2.25 but I'm using glibc 2.17.

karya0:
Dynamic offset calculation fails sometimes due to a false hit (when tid
matches some bytes in the tcbhead struct).

Fixes #994.

### DIFF
--- a/src/threadlist.cpp
+++ b/src/threadlist.cpp
@@ -78,24 +78,6 @@ static int restarthread(void *threadv);
 static void Thread_SaveSigState(Thread *th);
 static void Thread_RestoreSigState(Thread *th);
 
-// Copied from src/plugin/pid/pid_syscallsreal.c
-// Without this, libdmtcp.so will depend on libdmtcp_plugin.so being loaded
-static pid_t _real_getpid(void)
-{
-  JWARNING("_real_getpid")
-          .Text("FIXME: _real_getpid returning virtual pid, not real pid.");
-  // libc caches pid of the process and hence after restart, libc:getpid()
-  // returns the pre-ckpt value.
-  return (pid_t)_real_syscall(SYS_getpid);
-}
-// Copied from src/plugin/pid/pid.c and .../pid/pid_syscallsreal.c
-// Without this, libdmtcp.so will depend on libdmtcp_plugin.so being loaded
-LIB_PRIVATE
-pid_t dmtcp_get_real_pid()
-{
-  return _real_getpid();
-}
-
 /*****************************************************************************
  *
  * Lock and unlock the 'activeThreads' list
--- a/src/tls.cpp
+++ b/src/tls.cpp
@@ -67,6 +67,9 @@ static int glibcMinorVersion()
   return minor;
 }
 
+/*****************************************************************************
+ *
+ *****************************************************************************/
 /* Offset computed (&x.pid - &x) for
  *   struct pthread x;
  * as found in:  glibc-2.5/nptl/descr.h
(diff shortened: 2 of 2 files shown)


karya0:
Hi @JainTwinkle, this is also good to go. Let me know if you have any questions.

JainTwinkle:
@karya0, 
The issue [Warning: Couldn't find offsets of tid/pid in thread_area](https://github.com/dmtcp/dmtcp/issues/994#) is related. Even with the static offset, there can be a PID mismatch because the `THREAD_REAL_PID` function returns a virtual PID. It'd be great if we could improve this PR to fix the `THREAD_REAL_PID` function. What do you think?

karya0:
@JainTwinkle : Thanks for bringing it up. I have updated the code with proper offsets. We shouldn't need dynamic offset calculations until glibc updates size of tcbhead_t.

src/tls.cpp
```diff
@@ -142,192 +145,47 @@ static int glibcMinorVersion()
...
-  if (pid_offset == -1) {
-    int tid_offset = TLSInfo_GetTidOffset();
-    pid_offset = tid_offset + (char *)&(tid_pid.pid) - (char *)&tid_pid;
-    JTRACE("") (pid_offset);
-  }
   return pid_offset;
 }
 
-static char *
-memsubarray(char *array, char *subarray, size_t len)
```
JainTwinkle, line 308: This seems like a utility function we could use. Instead of removing it, can we move it either in mtcp or util?

src/threadlist.cpp
```diff
@@ -78,24 +78,6 @@ static int restarthread(void *threadv);
 static void Thread_SaveSigState(Thread *th);
 static void Thread_RestoreSigState(Thread *th);
 
-// Copied from src/plugin/pid/pid_syscallsreal.c
-// Without this, libdmtcp.so will depend on libdmtcp_plugin.so being loaded
```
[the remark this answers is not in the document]
karya0: It wouldn't because we have:
```
#define THREAD_REAL_PID() \
  (dmtcp_get_real_pid != NULL ? dmtcp_get_real_pid() : getpid())
```

src/tls.cpp
karya0: We can use std::find or friends if we need similar functionality later on.

PR #1008, opened 2022-07-17
Handle /proc/self/task/* files.

karya0:
Updated procfd1 test with /proc/self/task/<tid>/maps file example.


JainTwinkle (approved):
LGTM!

PR #1011, opened 2022-08-02
Util to test for FSGSBASE patch; info on use

gc00:
This is a useful quick test with the MANA plugin, to see if the user-space FSGSBASE patch is present, when calling the lower half.

### DIFF
--- a/util/fsgsbase-enabled.c
+++ b/util/fsgsbase-enabled.c
@@ -1,0 +1,31 @@
+#include <sys/auxv.h>
+#include <elf.h>
+#include <stdio.h>
+
+/* The user-space FSGSBASE patch was added to Linux kernel version 5.9. */
+
+/* Will be eventually in asm/hwcap.h */
+#ifndef HWCAP2_FSGSBASE
+# define HWCAP2_FSGSBASE (1 << 1)
+#endif
+
+int main() {
+  unsigned val = getauxval(AT_HWCAP2);
+  if (val & HWCAP2_FSGSBASE) {
+    printf("FSGSBASE enabled (both in CPU and Linux kernel)\n");
+    return 0;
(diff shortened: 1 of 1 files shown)


JainTwinkle:
@gc00 
While this is a nice way to test the user-space FSGSBASE feature, shouldn't this be a part of MANA or CRAC? Also, we can add this test to `configure.ac` to configure MANA or CRAC accordingly.

xuyao0127:
I agree with Twinkle that this probably should be part of MANA or CRAC, instead of DMTCP. DMTCP doesn't need to set/get the FS register.

JainTwinkle:
On second thought, DMTCP is common for both MANA and CRAC projects. Adding the utility program here will reflect on both projects. Moreover, DMTCP can also utilize this feature as well. For example, to get and set thread area in `tls.cpp`.

gc00:
As twinkle says:
> DMTCP is common for both MANA and CRAC projects. Adding the utility program here will reflect on both projects.

So, a MANA developer will have this utility available immediately, since they will see the util directory.  And MANA does not have its own util directory.  Right now, I can't see creating a special MANA util directory.

And if we place it somewhere in MANA, then the DMTCP programmer will not see this.

I could do either, but if we're not sure, I suggest putting this in DMTCP for now.

JainTwinkle (approved):
Currently, MANA has two util folders: first in top-level dir (`mana/util`) and the second in DMTCP submodule (`mana/dmtcp/util`). This PR will add the test program in the `DMTCP's util` dir, but not the `top-level util` dir.

I am approving this PR. But, we can consider creating a symlink (mana/util -> mana/dmtcp/util) in MANA or adding the same test program in MANA's top-level dir to keep it identical.

PR #1012, opened 2022-08-02
configure.ac/configure: check FSGSBASE in kernel

gc00:
`include/config.h` will define or undefine `HAS_FSGSBASE` if the CPU/Linux kernel supports it.  This is needed right now for MANA.


JainTwinkle (approved):
LGTM!

PR #1015, opened 2022-08-14
Fix syntax error in defining JTRACE

dahongli:


JainTwinkle (approved):
Thanks for catching this one. LGTM!

PR #1017, opened 2022-08-26
Handle deleted shm area case

JainTwinkle:
If a shared memory segment is marked deleted, the kernel gives a **different** shmID with the same key. It creates more than shm regions in the process address space with the same key but different shmIDs.
However, DMTCP doesn't mark deleted while restoring the shm areas, and the kernel gives the **same** shmid to DMTCP. This PR handles this use case.
[Clarification added by gc00:  "If a shared memory segment is marked deleted," means: `shmctl(shmid, IPC_RMID, NULL);` was called.]


karya0 (approved):
LGTM

PR #1018, opened 2022-08-26
Remove SYSV segments from bookeeping when detached from all processes…

chirag-singh-memverge:
1. Before shmdt call, update the SHM Segment's local flags on number of attached process and if it is marked for deletion
2. If shmdt is successful, the previous nattch=1, and it is marked from deletion, remove from SysVIPC bookeeping

This ensures we don't try to restore any segments that have been deleted already.


chirag-singh-memverge:
This is already handled in the leader election

PR #1019, opened 2022-08-31
Fixed handling of unlinked and deleted files on restart

chirag-singh-memverge:
Instead of writing to a file and remapping it, if a file has been unlinked prior to checkpointing, there's no reason to write back to that file. Instead open and immediately unlink the file, then refill the memory segment contents

### DIFF
--- a/src/plugin/ipc/file/fileconnlist.cpp
+++ b/src/plugin/ipc/file/fileconnlist.cpp
@@ -473,30 +473,58 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
   JASSERT(createDirectoryTree(area.name)) (area.name)
   .Text("Unable to create directory in File Path");
 
-  /* Now try to create the file with O_EXCL. If we fail with EEXIST, there
-   * are two possible scenarios:
-   * - The file was created by a different restarting process with data from
-   *   checkpointed copy. It is possible that the data is "in flight", so we
-   *   should wait until the next barrier to compare the data from our copy.
-   * - The file existed before restart. After the next barrier, abort if the
-   *   contents differ from our checkpointed copy.
-   */
-  int fd = _real_open(area.name, O_CREAT | O_EXCL | O_RDWR,
-                      S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
-  JASSERT(fd != -1 || errno == EEXIST) (area.name);
+  // Unlinking /libhugetlbfs.tmp is different from unlinking an ordinary file.
+  // The file is a virtual file.  It is not really unlinked, and the memory
+  // segment was originally backed by this solely to force the use of hugepages.
+  if (strstr(area.name, "/libhugetlbfs.tmp")) {
+    JTRACE("Unlinked huge page region detected") (area.name)
+      ((void *)area.addr);
 
-  if (fd == -1) {
-    fd = _real_open(area.name, O_RDWR);
-    JASSERT(fd != -1) (JASSERT_ERRNO);
+    int fd = _real_openat(AT_FDCWD, area.name, O_RDWR|O_CREAT|O_EXCL, 0600);
+    JASSERT(fd != -1) (JASSERT_ERRNO) (area.name);
+
+    // Unlink (the area was originally unlinked)
+    unlink(area.name);
+
+    // Restore SHM area mapping
+    // The mmap in restoreShmArea will clear the content in the area because the
+    // file has been unlinked, so we have to save the content first into a
+    // temp buffer
+    void *tempbuf = mmap(NULL, area.size, PROT_READ | PROT_WRITE,
+                        MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    JASSERT(tempbuf != MAP_FAILED) (JASSERT_ERRNO) (area.size);
+    memcpy(tempbuf, area.addr, area.size);
+    restoreShmArea(area, fd);
+    // Restore the content from temp buffer
+    memcpy(area.addr, tempbuf, area.size);
+    munmap(tempbuf, area.size);
   }
+  else {
+    /* Now try to create the file with O_EXCL. If we fail with EEXIST, there
+    * are two possible scenarios:
+    * - The file was created by a different restarting process with data from
+    *   checkpointed copy. It is possible that the data is "in flight", so we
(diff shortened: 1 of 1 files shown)


src/plugin/ipc/file/fileconnlist.cpp
```diff
@@ -490,13 +490,22 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
...
+  // Unlink (the area was originally unlinked)
+  unlink(area.name);
+
+  // Restore SHM area mapping
+  // The mmap in restoreShmArea will clear the content in the area because the
+  // file has been unlinked, so we have to save the content first into a
+  // temp buffer
+  void* tempbuf = mmap(NULL, area.size,
+                       PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, 0, 0);
+  JASSERT(tempbuf > (void*)0) (MAP_FAILED);
```
JainTwinkle, line 502: Please change it to `JASSERT(tempbuf != MAP_FAILED) (JASSERT_ERRNO) (area.size);`

src/plugin/ipc/file/fileconnlist.cpp
```diff
@@ -490,13 +490,22 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
...
-    (JASSERT_ERRNO);
+  // Unlink (the area was originally unlinked)
+  unlink(area.name);
+
+  // Restore SHM area mapping
+  // The mmap in restoreShmArea will clear the content in the area because the
+  // file has been unlinked, so we have to save the content first into a
+  // temp buffer
+  void* tempbuf = mmap(NULL, area.size,
+                       PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, 0, 0);
```
JainTwinkle, line 501: Shouldn't `fd` (second last argument in the mmap) be `-1`?

src/plugin/ipc/file/fileconnlist.cpp
```diff
@@ -490,13 +490,22 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
...
-    (JASSERT_ERRNO);
+  // Unlink (the area was originally unlinked)
+  unlink(area.name);
+
+  // Restore SHM area mapping
+  // The mmap in restoreShmArea will clear the content in the area because the
+  // file has been unlinked, so we have to save the content first into a
+  // temp buffer
+  void* tempbuf = mmap(NULL, area.size,
+                       PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
```
JainTwinkle, line 501: Could you please add space before and after the OR operation?

src/plugin/ipc/file/fileconnlist.cpp
```diff
@@ -490,13 +490,22 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
...
-  JASSERT(Util::writeAll(fd, area.addr, area.size) == (ssize_t)area.size)
-    (JASSERT_ERRNO);
+  // Unlink (the area was originally unlinked)
+  unlink(area.name);
+
+  // Restore SHM area mapping
+  // The mmap in restoreShmArea will clear the content in the area because the
+  // file has been unlinked, so we have to save the content first into a
+  // temp buffer
+  void* tempbuf = mmap(NULL, area.size,
```
JainTwinkle, line 500: `void * tempbuf` -> `void *tempbuf`

src/plugin/ipc/file/fileconnlist.cpp
```diff
@@ -490,13 +490,22 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
...
+                       PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
+  JASSERT(tempbuf != MAP_FAILED) (JASSERT_ERRNO) (area.size);
+  memcpy(tempbuf, area.addr, area.size);
   restoreShmArea(area, fd);
+
+  // Restore the contentfrom temp buffer
```
JainTwinkle, line 506: contentfrom -> content from the

src/plugin/ipc/file/fileconnlist.cpp
```diff
@@ -490,13 +490,22 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
...
-  // Now populate file contents from memory.
-  JASSERT(Util::writeAll(fd, area.addr, area.size) == (ssize_t)area.size)
-    (JASSERT_ERRNO);
+  // Unlink (the area was originally unlinked)
+  unlink(area.name);
+
+  // Restore SHM area mapping
+  // The mmap in restoreShmArea will clear the content in the area because the
+  // file has been unlinked, so we have to save the content first into a
+  // temp buffer
```
gc00, line 499: If I remember correctly, this bug appeared because area.name was originally a file based on hugepages.  Is this bug fix specific to a bug that appeared with a memory segment of hugepages,  Or does the bug appear with an ordinary mmap using MAP_SHARED also?

If it's specific to hugepages, we need to add something about that in the comment.  If it's about an ordinary mmap with MAP_SHARED, can we further simplify the example that led to this bug?

chirag-singh-memverge: This does not seem to be related to the hugepages at all.

src/plugin/ipc/file/fileconnlist.cpp
```diff
@@ -473,30 +473,55 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
   JASSERT(createDirectoryTree(area.name)) (area.name)
   .Text("Unable to create directory in File Path");
```
gc00, line 475: I suggest an added comment on the next line:
```
// NOTE:  Unlinking /libhugetlbfs.tmp is different from unlinking an ordinary file.
//             The file is a virtual file.  It is not really unlinked, and the
//             memory segment was originally backed by this solely to force
//             the use of hugepages.
```
Feel free to re-phrase anything if you think it's not clear.

gc00 (approved):
Please add the comment in the code, as suggested (or something similar).  After that, it LGTM.
I'm approving this in advance.

PR #1021, opened 2022-09-14
For gcc-4.8, std::put_time was failing

gc00:
std::put_time was introduced in commit b2823e39 (July, 2022)

Apparently, gcc-4.8 (and therefore CentOS 7) doesn't support it.  I added a conditional around it: `#if __GCC__ > 4`

### DIFF
--- a/jalib/jassert.cpp
+++ b/jalib/jassert.cpp
@@ -87,7 +87,14 @@ jassert_internal::JAssert::JAssert(const char* type, bool exitWhenDone)
     Print("\n");
   }
 
-  ss << "[" << std::put_time(&localTime, "%F, %T.") << ms << ", "
+  ss << "[" 
+#if __GCC__ > 4
+    // put_time introduced in commit b2823e39 (July, 2022)
+    // In CentOS 7 using gcc/g++ 4.8.5, this fails with:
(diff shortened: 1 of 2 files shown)


jalib/jassert.cpp
```diff
@@ -87,7 +87,14 @@ jassert_internal::JAssert::JAssert(const char* type, bool exitWhenDone)
...
   }
 
-  ss << "[" << std::put_time(&localTime, "%F, %T.") << ms << ", "
+  ss << "[" 
+#if __GCC__ > 4
+    // put_time introduced in commit b2823e39 (July, 2022)
+    // In CentOS 7 using gcc/g++ 4.8.5, this fails with:
+    //      error: ‘put_time’ is not a member of ‘std’ 
+     << std::put_time(&localTime, "%F, %T.")
+#endif
```
JainTwinkle, line 96: Why are we not adding an else condition with a C version of std::put_time?

karya0:
We can replace it with a C-style implementation. I’ll generate a separate
PR.

JainTwinkle (approved):
@karya0 promises to generate a separate PR for the c-style implementation. This PR is ready to get merged.

karya0 (approved):

PR #1022, opened 2022-09-16
Use curThread->virtual_tid in dmtcp_gettid.

karya0:


dahongli (approved):
The change looks good. It would save a lots of syscalls for gettid. Thanks Kapil!

gc00:
@karya0 , I agree that this looks good.  But here is an interesting web page to be aware of:
  https://stackoverflow.com/questions/63589571/gcc-common-variable-attribute-on-tls-model

The bug reported there probably doesn't affect your code.  But could you briefly review that bug report?  Thanks.

It says that:
```
This is a bug, possibly in the C++ frontend, and/or in attribute merging code. If you move or duplicate the attribute from the declaration on the definition, it works as intended. If you eliminate the definitions, and leave only the declarations, it also works according to the documentation. Basically, each definition does not appear to properly inherit the attribute from the preceding declaration. This is worthy of a bug report at https://gnu.gnu.org/bugzilla
```

karya0:
Thanks for the pointer, @gc00! This is an interesting bug but yes, it doesn't affect this PR because we are using the attributes in the definition, not declaration.

karya0:
I just realized that @gc00 didn't approve the PR formally, but the comment provided the intention.

gc00:
Correct.  I intended to approve PR #1022 (this one).  Thanks for merging it.

PR #1023, opened 2022-09-20
Simulate vfork using fork.

karya0:


gc00 (changes requested):
@karya0 ,
The man page for vfork says:
```
 vfork()  differs  from  fork(2) in that the calling thread is suspended
       until the child terminates (either normally, by  calling  _exit(2),  or
       abnormally,  after  delivery  of a fatal signal), or it makes a call to
       execve(2).  Until that point, the child shares all memory with its par‐
       ent,  including  the stack.  The child must not return from the current
       function or call exit(3), but may call _exit(2).
```
If we're going to emulate vfork, shouldn't the parent process call waitpid() on the child, and wait for the child to finish?  This version allows them both to execute at the same time.

gc00 (approved):
I was doing too much multitasking earlier.  Thanks, @karya0.  Yes, vfork waits only for the exec.

I'm approving this now.  I assume that you can override the CodeFactor complaint about security issues.  It doesn't apply in our case.

PR #1024, opened 2022-09-20
Added ability to truncate append-only/RW files on restart.

karya0:
A plugin can define dmtcp_should_truncate_file_on_restart() and return '0' if it wants the file contents to be preserved. Otherwise it can defer to the next plugin (eventually to dmtcpplugin.cpp) and the file will be truncated by default.


JainTwinkle:
I'm seeing the following conflict in the signature:
```
dmtcpplugin.cpp: In function ‘int dmtcp_should_truncate_file_on_restart()’:
dmtcpplugin.cpp:219:1: error: conflicting declaration of C function ‘int dmtcp_should_truncate_file_on_restart()’
 dmtcp_should_truncate_file_on_restart(void)
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from uniquepid.h:28:0,
                 from dmtcpmessagetypes.h:29,
                 from coordinatorapi.h:27,
                 from dmtcpplugin.cpp:25:
../include/dmtcp.h:401:5: note: previous declaration ‘int dmtcp_should_truncate_file_on_restart(const char*)’
 int dmtcp_should_truncate_file_on_restart(const char* path);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

karya0:
@JainTwinkle : I pushed it too soon. Fixed now.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -351,9 +351,17 @@ FileConnection::refill(bool isRestart)
           // source at restart, the requests are also saved after checkpoint.
           // The log file size at restart is larger than its size saved in the
           // checkpoint image. Just give a warning here and continue.
-          JWARNING(false) (_path) (_st_size) (statbuf.st_size)
-          .Text("Setting saved size to the current file size");
-          _st_size = statbuf.st_size;
+          if (dmtcp_should_truncate_file_on_restart(_path.c_str())) {
```
JainTwinkle, line 354: It should be another way around. By default, DMTCP should truncate and it should avoid truncating only when the plugin instructs it to avoid.

dahongli (approved):
Thanks Kapil for fixing this issue!

JainTwinkle:
@karya0 Thanks for the last fix. Since we have introduced a new env variable and a new feature, let's update the PR description stating the usage. Also, it would be helpful to have a comment stating the same in the code.

include/dmtcp.h
```diff
@@ -398,6 +398,7 @@ const char *dmtcp_get_ckpt_filename(void) __attribute__((weak));
 const char *dmtcp_get_ckpt_files_subdir(void);
 int dmtcp_should_ckpt_open_files(void);
 int dmtcp_allow_overwrite_with_ckpted_files(void);
+int dmtcp_should_truncate_file_on_restart(const char* path);
```
JainTwinkle, line 401: Do we need __attribute__((weak)) just like dmtcp_skip_memory_region_ckpting? 
Ref: https://github.com/dmtcp/dmtcp/blob/9aea31b3dd974f5d2634b3161dfdd9844038bb63/include/procmapsarea.h#L121

karya0: We don't need weak because we have the terminating function explicitly defined in dmtcpplugin.cpp. We should eventually update other similar APIs as well.

karya0:
> @karya0 Thanks for the last fix. Since we have introduced a new env variable and a new feature, let's update the PR description stating the usage. Also, it would be helpful to have a comment stating the same in the code.

I ended up removing the env-var because it doesn't make sense to provide a value '0' to "enable" this thing.

karya0:
@JainTwinkle: I have updated the logic to truncate only if the file is opened in O_APPEND mode. For a regular file with O_WRONLY or O_RDWR permissions, it doesn't make sense for us to truncate them. Thoughts?

JainTwinkle:
I disagree. I think we should truncate them. Setting just the offset is not enough. Consider the following example:
File -> 1 2 3 4 5 (manually ckpt'ed here) 6 7 8 9 10 11 12 13 14 15
The file restarted from the offset after 5, and someone cancels/terminates the execution after 10. The file content will be 1 through 15 even then. Moreover, if some application wants to write other than existing content then it would be a mixture of the new and the old content.

karya0:
That's a valid concern. A proper solution would require detecting file size differences between startup, ckpt-time, and restart-time to determine if the file is a good candidate for truncation. I'm worried about truncating user-data that was not generated by the current process.

JainTwinkle:
I think it is okay to truncate them, as any operation related to those files would give them misleading results, for example, the total file size, end of the file offset, and many other things. Also, when we talk about restoring the application, theoretically, DMTCP should restore the files' state as they were at the checkpoint. 

One can disable the truncate feature, and for that, I think you should reintroduce an env-var because adding a new plugin would be too much work for a simple option flip. We can call the function `dmtcp_skip_truncating_file_at_restart` instead. When the env-var is set, DMTCP won't truncate any files. Someone can have a plugin if they wish to exclude truncation of a specific set of files.

karya0:
Readded the env var. Let's merge it in.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -342,7 +342,9 @@ FileConnection::refill(bool isRestart)
 
       if (stat(_path.c_str(), &statbuf) == 0 && S_ISREG(statbuf.st_mode)) {
         if (statbuf.st_size > _st_size &&
-            ((_fcntlFlags & O_WRONLY) || (_fcntlFlags & O_RDWR))) {
+            ((_fcntlFlags & O_APPEND) ||
```
JainTwinkle, line 345: Please change the order of flags. It should be O_WRONLY, O_RDWR, and then O_APPEND. In a weird setting (O_APPEND + O_RDONLY), truncate would fail.

JainTwinkle (approved):
LGTM! I'm asking for a very minor change. You can change if you agree. Otherwise, please go ahead and merge it.

src/plugin/ipc/file/fileconnection.cpp
JainTwinkle: I meant if a file was open in an append mode and with read-write permission. However, the application chooses to change the access mode.

karya0: We are reading flags from fcntl call so should have the right values. Even if there were some weird combination, it would be okay to truncate the file.

PR #1026, opened 2022-09-23
Updated zero-page restoration support.

karya0:
Instead of mmapping zero-page segments in chunks, we now save a "parent" header in the ckpt-image followed by several "child" headers. The parent header is used only for making the appropriate mmap call. The data-restoration relies on child headers. Each child header is either marked as zero-page segment, in which case, there is no further action needed. For non-zero child segments, the restoration reads the saved data from ckpt image.


dahongli:
@karya0 @Twinklejain  Is this PR for fixing MANA-250 crash issue? Could you add more detail in the PR description about the root cause of the bug and how this fix can solve it?

xuyao0127 (approved):
LGTM.

PR #1027, opened 2022-09-24
Add flag, --enable-fsgsbase-override

gc00:
This is for a cross-configure.  DMTCP tests the local machine for the FSGSBASE kernel patch.  But the FSGSBASE kernel patch might be installed on the build computer, but not on the login computer. This flag will define a macro, HAS_FSGSBASE_OVERRIDE, and any package using DMTCP as a submodule can inherit this macro with a '#include "config.h"'.

### DIFF
--- a/configure
+++ b/configure
@@ -707,6 +707,8 @@ HAS_READLINE
 M32
 CONFIG_M32_FALSE
 CONFIG_M32_TRUE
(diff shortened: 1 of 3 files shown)


xuyao0127 (approved):
LGTM.

PR #1029, opened 2022-09-29
Avoid PATH_MAX strings on stack

karya0:
Some applications use a much smaller signal stack and putting several PATH_MAX strings on stack can cause overflow.

We use mmap to allocate memory for now; in future, we will start using a pool of pre-allocated PATH_MAX strings.

### DIFF
--- a/include/pathbuffer.h
+++ b/include/pathbuffer.h
@@ -1,0 +1,51 @@
+#ifndef __DMTCP_BUFFER_H__
+#define __DMTCP_BUFFER_H__
+
+#include <limits.h>  // for PATH_MAX
+#include <sys/mman.h>
+#include "jalloc.h"
+#include "jassert.h"
+
+namespace dmtcp
+{
+class PathBuffer
+{
+  public:
+#ifdef JALIB_ALLOCATOR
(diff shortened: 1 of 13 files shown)


karya0:
> The changes related to get rid of using PATH_MAX buffer on stack look good to me.
> 
> Kapil, Do we need to move to C++17 right now? Change of moving to c++17 should be a separate PR.

Removed C++17 commit. It'll be part of a separate PR

karya0:
> @karya0 Can we leverage the information in `/proc/sys/kernel/pid_max` to decide the max_pid digits and the fact that we have MAX_VIRTUAL_PID? It would be 10 digits (chars) max.

@JainTwinkle : The MAX_VIRTUAL_PID trick will work only for /proc/<PID>/xyz. For that we are already using `dmtcp::string` now, which will use the static buffer for strings less than 20 chars or so. The virtual-to-real api can be used for path translations unrelated to virtual-pid and that's where we don't know the full size.

I'm contemplating on moving to strings for that api as well. That way we can rely on std::string to do the heavy lifting for us. We can discuss it in more detail.

yuelimv (approved):

karya0:
Abondoning this PR since it's not needed right now.

PR #1030, opened 2022-10-03
Save Coordinator KeyValue database at ckpt/rst.

karya0:
Cleaned up LookupService to allow dumping of key-value data into a json file for future debugging. One of the primary change is to use Base64 strings instead of binary data in the key-value store. This simplifies the coordinator API and makes jsons more legible. Also removed some stale API that was introduced for the Infiniband plugin but not used otherwise.

We can incorporate more coordinator information in the database in followup PRs.


karya0:
Closing in lieu of #1032

PR #1033, opened 2022-10-06
Replaced std::puttime with Util::getTimestampStr().

karya0:
It resolves the issue related to missing std::put_time on older compilers.

### DIFF
--- a/jalib/jassert.cpp
+++ b/jalib/jassert.cpp
@@ -75,27 +75,13 @@ jassert_internal::JAssert::JAssert(const char* type, bool exitWhenDone)
   , JASSERT_CONT_B(*this)
   , _exitWhenDone(exitWhenDone)
 {
-  struct timeval tv;
-  struct tm localTime;
-
-  gettimeofday(&tv, NULL);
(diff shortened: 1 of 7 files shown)


jalib/jassert.cpp
```diff
@@ -302,7 +288,8 @@ void jassert_internal::open_log_file()
   if (theLogFileFd != -1) {
     dmtcp::ostringstream a;
 
-    a << "[" << getpid() << "] INFO at " << JASSERT_FILE << ":" << JASSERT_LINE
+    a << "[" << jalib::getTimestampStr() << ", " << getpid() << ", "
```
JainTwinkle, line 291: @karya0, should we also add the keyword PID(or P) or TID while logging? It would be clear to anyone who's not familiar with the logging format.

JainTwinkle (approved):
Please see my suggestion in the comment. The rest looks good to me. Thanks!

jalib/jassert.cpp
karya0: Putting Pid/Tid in log lines is a fairly standard practice so we don't have to worry about people unfamiliar with it :).

PR #1034, opened 2022-10-10
Added FSGSBASE check and env var.

karya0:


gc00 (commented):
I don't understand this.
First of all, it appears to test rdfsbase (assembly instruction).  But what about wrfsbase?  we could always read the fsbase.  It's writing the fsbase that became possible with Linux 5.9.

Second, could we add a comment for ENV_VAR_FSGSBASE_ENABLED?  Is this supposed to guarantee that wrfsbase will work?  I believe that we could always read the fs register without a kernel call.

karya0:
`rdfsbase` doesn't work without the kernel patch -- I confirmed this on cori head node where this program will fail with SIGILL. In any case, I have now added wrfsbase as well (without checking the return value). The instruction will fail with SIGILL if it's not supported. If we feel strongly about it, we can come up with a more extensive check (setting fsbase and reading it back, etc.).

gc00 (approved):
@karya0 , I'm approving this in advance, and leaving it to your discretion.  Thanks for `wrfsbase'.  Being able to read it back in might be nice, but not essential.

QUESTION:  If this code can produce SIGILL, do we have a SIGILL signal handler in dmtcp_launch?  If not, won't dmtcp_launch simply crash?

karya0:
@gc00: SIGILL is a pain to handle, that's why we are doing this test in a child process and signaling parent via exit code.

PR #136, opened 2015-07-08
Test make check for Travis CI

karya0:

PR #1037, opened 2022-11-07
Several fixes for better MANA support

karya0:
Commits in this PR are largely independent but placed together in this PR for quicker review.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -1112,38 +1112,7 @@ read_one_memory_area(int fd, VA endOfStack)
         mmap_fixed_noreplace(area.addr, area.size, area.prot | PROT_WRITE,
                             area.flags, imagefd, area.offset);
 
-      if (mmappedat == MAP_FAILED) {
-        MTCP_PRINTF("error %d mapping %p bytes at %p\n",
-                mtcp_sys_errno, area.size, area.addr);
-        if (area.properties & DMTCP_ZERO_PAGE_PARENT_HEADER) {
-          size_t parentSize = area.size;
-          while (parentSize > 0) {
-            mtcp_readfile(fd, &area, sizeof area);
-            if (area.properties & DMTCP_ZERO_PAGE) {
-              // Zero page; nothing to do.
-              DPRINTF("Skipping zero page at %p with %p bytes\n", area.addr, area.size);
-            } else {
-              DPRINTF("Skipping non-zero page at %p with %p bytes\n", area.addr, area.size);
-              // Skip this segment.
-              MTCP_ASSERT(area.properties & DMTCP_ZERO_PAGE_CHILD_HEADER);
-              mtcp_skipfile(fd, area.size);
-            }
-            parentSize -= area.size;
-          }
-        } else {
-          DPRINTF("Skipping region at %p with %p bytes\n", area.addr, area.size);
-          mtcp_skipfile(fd, area.size);
-        }
-
-        if (imagefd >= 0) {
-          mtcp_sys_close(imagefd);
-        }
-
-        return 0;
-      } else if (mmappedat != area.addr) {
-        MTCP_PRINTF("area at %p got mmapped to %p\n", area.addr, mmappedat);
-        mtcp_abort();
-      }
+      MTCP_ASSERT(mmappedat == area.addr);
 
   #if 0
       /*


src/pluginmanager.cpp
```diff
@@ -93,6 +94,8 @@ PluginManager::eventHook(DmtcpEvent_t event, DmtcpEventData_t *data)
 
   case DMTCP_EVENT_VIRTUAL_TO_REAL_PATH:
 
+  case DMTCP_EVENT_PRESUSPEND:
+  case DMTCP_EVENT_PRECHECKPOINT:
```
gc00, line 98: I suggest adding the comment, below.  Events for ckpt/resume/restart are subtle,
since plugins must be called in _reverse_ order during resume/restart.  Let's
document this in the code, so that others will be clearly aware of this issue.
I suggest the following comment:
```
    // The plugins can be thought of as implementing a layered software
    // architecture.  All of the events here occur before writing the checkpoint
    // file.  The plugins are invoked for these events in the natural order.
    // For the resume/restart events below, the plugins are invoked
    // in _reverse_ order.  This is required to support layered software.
    // For an analogous case, see 'man pthread_atfork' with the handlers:
    // (i) prepare, (ii) parent, and (iii) child.
    // Those are analogous to our events for:
    // (i) pre-checkpoint, (ii) resume event, and (iii) restart; respectively.
```

src/pluginmanager.cpp
```diff
@@ -114,46 +117,16 @@ PluginManager::eventHook(DmtcpEvent_t event, DmtcpEventData_t *data)
...
   case DMTCP_EVENT_RESUME:
-    for (int i = pluginManager->pluginInfos.size() - 1; i >= 0; i--) {
-      if (pluginManager->pluginInfos[i]->event_hook) {
-        pluginManager->pluginInfos[i]->event_hook(event, data);
-      }
-    }
-  break;
-
   case DMTCP_EVENT_RESTART:
+  case DMTCP_EVENT_THREAD_RESUME:
```
gc00, line 122: 1i suggest the following comment, to correspond to the one, above.
```
    // The plugins are invoked in _reverse_ order during resume/restart.  This is
    // required to support layered software.  See the related comment, above.
```

src/pluginmanager.cpp
```diff
@@ -114,46 +117,16 @@ PluginManager::eventHook(DmtcpEvent_t event, DmtcpEventData_t *data)
...
+
```
gc00, line 123: Here, we added a blank line after the case statements.  But in the PRESUSPEND case states, we do not have a blank line.  For consistency, we should either have a blank line in both situations, or no blank line in both situations.

src/coordinatorapi.cpp
```diff
@@ -755,11 +755,12 @@ kvdb::KVDBResponse
 kvdbRequest(DmtcpMessage const& msg,
             string const& key,
             string const& val,
-            string *oldVal)
+            string *oldVal,
+            bool useNsSock)
```
gc00, line 759: Can we have a more meaningful name than `useNsSock`?  :-)
Longer variable names are good if they help people like me to read the code.  "Sock" probably means "Socket", and the extra two characters are a small cost for readability.
    I guess you were already using `nsSock`, but I would have to study the code further to figure out what an `nsSock` is.  Can we also change that name to something more meaningful?

src/mtcp/mtcp_restart.c
```diff
@@ -1112,38 +1112,7 @@ read_one_memory_area(int fd, VA endOfStack)
...
-          DPRINTF("Skipping region at %p with %p bytes\n", area.addr, area.size);
-          mtcp_skipfile(fd, area.size);
-        }
-
-        if (imagefd >= 0) {
-          mtcp_sys_close(imagefd);
-        }
-
-        return 0;
-      } else if (mmappedat != area.addr) {
```
gc00, line 1143: This is an "else" branch for `if (mmappedat == MAP_FAILED)`.  It diagnoses a different error, prints information, and aborts o this different error.
    We should not be removing this.

src/mtcp/mtcp_restart.c
```diff
@@ -1112,38 +1112,7 @@ read_one_memory_area(int fd, VA endOfStack)
         mmap_fixed_noreplace(area.addr, area.size, area.prot | PROT_WRITE,
                             area.flags, imagefd, area.offset);
 
-      if (mmappedat == MAP_FAILED) {
```
gc00, line 1115: I remember you saying that we hit an error in mtcp_restart.c, and we print the error, but try to continue anyway.
But I'm not sure that this code is doing what you said.  For example, I'm reading that if this is a zero page, then it's not really an error, and we really should be continuing.
Please remove this commit from the current PR.  If there's some worthwhile cleanup to do, please submit this single commit as a new PR, but not part of the current PR.

gc00 (approved):
I'm approving this in advance, but _only_ if you promise to do the recommendations that I'm including here.  Otherwise, I worry that we're taking code form the past that was non-obvious, and then writing extra code on top of it.  You and I understand the non-obvious code.  So, we will eventually figure it out when we see the extra code on top of it.  But it's a problem for others.

Specifically:
1.  Let's finally document in the comments about layers and invoking plugins in reverse order on resume/restart.
2. Let's add a `FIXME` comment about change the variable name `nsSock`.  Apparently, it used to mean `name_service_socket`, and we now need a more up-to-date name.  But a `FIXME` comment to this effect is fine for now.
3. Please do _not_ include the last commit:  "MTCP: Assert on mmap MAP_FIXED failure."  I think there's a lot to discuss there.  But I don't want to delay this PR.

Also, please note that we found that DMTCP was in a state where it was not compiling without this additional PR.  The first commit of this PR fixes that.  But we need to be more careful.

And finally, on a related subject, I just noticed:
 PR #136: Removed some stale code related to --no-coordinator.

Thanks for explaining the history to me.  Apparently, this PR#136 was quickly reviewed without a full discussion.  But the story is that the `--no-coordinator` code was already broken in multiple way.  Rather than fix it for the new kvdb architecture, it is faster and cleaner to produce a new implementation of `--no-coordinator` in a future PR.

PR #1038, opened 2022-11-08
Bugfix - Seek to the correct offset while restoring the "libhugetlbfs.tmp" area.

tarunsmalviya:


src/plugin/ipc/file/fileconnlist.cpp
```diff
@@ -483,6 +483,9 @@ FileConnList::recreateShmFileAndMap(const ProcMapsArea &area)
     int fd = _real_openat(AT_FDCWD, area.name, O_RDWR|O_CREAT|O_EXCL, 0600);
     JASSERT(fd != -1) (JASSERT_ERRNO) (area.name);
 
+    // Get to the correct offset
```
gc00, line 486: Suggest alternate wording:
```
    // Set the correct offset
```

gc00 (approved):
Thanks for this great PR.  I'm approving it in advance.  But please make the following changes.  You can then force push them into your branch ('git push -f ...'), and you will still have the approval to push in the PR after that.

1.  Please note my inline comment about re-phrasing your comment.  (Or maybe you already fixed that.)
2.  Separately, could you clean up some of the poor style formatting from the original PR.  Specifically, as below.
FROM:
```
}
    else {
```
TO:
```
} else {
```

And please correct the formatting of the comment in the "else" branch.  Starting at line 507, the original PR had moved the `*` on that line and below to appear under the `/` instead of under the `*` in `/*`.  The style formatting was correct earlier, and the PR changed it to the wrong style.

Thanks for adding these fixes.

dahongli (approved):
Good finding. Thanks Tarun!

PR #1040, opened 2022-11-10
Several fixes for MANA support

karya0:
Splitting #1037


gc00 (approved):
I'm approving this in advance.  But please see comments for PR #1037.  There are some suggested comments to add to the code.  Please add the comments before pushing this in.

Specifically:
1.  pluginmanager.cpp: Invoke plugins in natural order before ckpt; but in reverse order for plugins on resume/restart
2. nsSock - Used to be `name_service_socket`.  Please add a `FIXME` comment saying that this used to be called `name_service_socket` for an older architecture, and that we need a better name.

PR #1044, opened 2022-12-07
Document dependencies.

gogo2464:


gogo2464:
@gc00  Tiny PR (very incremental :) ). Ready to merge.

gc00 (approved):
Thanks.  Documentation is always useful!

And perhaps you want to include 'python3'.  We don't strictly need it to build DMTCP, but one would need to do a special configuration to avoid 'python3', and these days, Python is a small ask.

I'm approving this in advance.

gogo2464:
Change done. Ready to merge.

PR #1054, opened 2023-05-20
DMTCP CI fixes and enhancements

johnmuth81:
The DMTCP CI stopped working last December. The root cause was github updated `ubuntu-latest` from 20.04 to 22.04. A side effect of this change is default 'python` was upgraded from 3.8 to 3.10. Somewhere between python 3.8 and 3.10, the python `pty` module was changed in a way that breaks the DMTCP test driver (`test/sutotest.py`).

In addition, there are a few programs that aren't included in the default ubuntu 22.04 image that that will be used by DMTCP tests if present.

This change forces the CI to use python 3.8 to run the tests. This changes also installs some programs on the test runner to increase DMTCP test coverage.


gc00 (approved):
LGTM.
I assume that we will have to modify autotest.py in the future to work with Python 3.10.

When I look at Ubuntu 22.04 on my machine, I'm seeing:
  python3.10
  python3.8-minimal

gc00:
@karya0 , I've approved PR #1050. 

or PR #1054 here, should this be pushed in?  (It still requires a maintainer, you or me, to approve and run this,) I assume that the more principled solution will be to make autotest.py compatible with the pty module for both Python-3.8 and Python-3.10

johnmuth81:
> @karya0 , I've approved PR #1050.
> 
> or PR #1054 here, should this be pushed in? (It still requires a maintainer, you or me, to approve and run this,) I assume that the more principled solution will be to make autotest.py compatible with the pty module for both Python-3.8 and Python-3.10

In any case I believe you should at take the part the increases your test coverage:
```
    - name: add extra programs for tests
      run: |
        sudo apt-get update
        sudo apt-get install emacs emacs-nox gdb screen tcsh zsh
```

johnmuth81:
> Hi @johnmuth81, can you please update the PR with just the test coverage bits?

Of course. See commit aad2ace8eb52b66fdf6902848a2cddf477d805e1.

PR #1055, opened 2023-06-01
Use unordered_map in VirtualIdTable.

karya0:
This creates an unordered map in virtualidtable.h.  It can take a previous map, and create a new unordered map. Unordered map allows quicker lookup and also provide a facility to reserve space in advance. That way we can avoid multiple calls to malloc/free.

### DIFF
--- a/include/virtualidtable.h
+++ b/include/virtualidtable.h
@@ -23,6 +23,7 @@
 #define VIRTUAL_ID_TABLE_H
 
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


gc00 (approved):
Looks good to me.
I assume that one of the goals here is to create a set with 'capacity', so as to void the frequent small malloc's and free's.  I'll write some of that in the conversation, and you're free to improve on it.

PR #1057, opened 2023-06-15
Fix compiler warnings

gc00:
Fix a bunch of random compiler warnings


gc00:
@karya0, I believe I've now fixed the two issues for PR #1057 that you pointed to.  Thanks for the comments.  I'm pushing it in now.

PR #1058, opened 2023-06-16
Use MAP_FIXED_NOREPLACE flag

dahongli:
Perlmutter supports MAP_FIXED_NOREPLACE flag. Use this flag to guarantee the memory reservation is at the specified place.

### DIFF
--- a/src/mtcp/mtcp_util.c
+++ b/src/mtcp/mtcp_util.c
@@ -829,6 +829,7 @@ void* mmap_fixed_noreplace(void *addr, size_t len, int prot, int flags,
   if (flags & MAP_FIXED) {
     flags ^= MAP_FIXED;
   }
+  flags |= MAP_FIXED_NOREPLACE;
   void *addr2 = mtcp_sys_mmap(addr, len, prot, flags, fd, offset);
   if (addr == addr2) {
     DPRINTF("Mapped %p bytes at %p\n", len, addr);
--- a/src/mtcp/mtcp_util.c
+++ b/src/mtcp/mtcp_util.c
@@ -25,6 +25,7 @@
  *
  *****************************************************************************/
 
+#include <linux/version.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


gc00 (approved):
@dahongli ,
    I modified your PR branch to add a little code to your PR, to check if the MAP_FIXED_NOREPLACE flag was available.  It was introduced in Linux 4.17, and I believe that the version on Perlmutter is equal or later than this.

You should be able to push this in directly, now.

QUESTION:  What was the error that you were seeing on Perlmutter?  I would have thought that `map_fixed_noreplace()` would already correctly emulating `MAP_FIXED_NOREPLACE`.  I'd be curious to know how the two versions differ in their behavior.

dahongli:
> @dahongli , I modified your PR branch to add a little code to your PR, to check if the MAP_FIXED_NOREPLACE flag was available. It was introduced in Linux 4.17, and I believe that the version on Perlmutter is equal or later than this.
> 
Thanks Gene for adding the version checking code!


> QUESTION: What was the error that you were seeing on Perlmutter? I would have thought that `map_fixed_noreplace()` would already correctly emulating `MAP_FIXED_NOREPLACE`. I'd be curious to know how the two versions differ in their behavior.
Since map_fixed_noreplace() does not have any flag to tell the kernel the memory address should be at the specified fix address, so the kernel doesn't try hard enough to give the specified address. This function returns error if the mapped address is not the same as we specified. From logistic point of view, it does try to emulate the fixed_noreplace. However, the chance for map_fixed_noreplace to failure is much higher than using MAP_FIXED_NOREPLACE in which kernel works much harder to return the specified address. 

Here is error we saw on Perlmutter.
After getcontext[17631] mtcp_util.c:845 mmap_fixed_noreplace:
  error 12 mapping 0xfffffffb40958000 bytes at 0x7fffbffd3000, flags: 0x100022, prot :0x0

PR #1060, opened 2023-06-23
Dlsym cleanup to remove initialization issues.

karya0:
Removed JTRACE call from within dlsym since it can lead segfault due to recursive calls. As part of the cleanup, Pid plugin is also sanitized to remove dead code, etc.

### DIFF
--- a/src/dmtcp_dlsym_wrappers.cpp
+++ b/src/dmtcp_dlsym_wrappers.cpp
@@ -30,11 +30,11 @@ print_debug_messages(dt_tag tags,
     JTRACE("** symbol version: ")
           (version_name(tags.versym[default_symbol_index], &tags));
   }
-#endif /* ifdef VERBOSE */
   if (!default_symbol_index) {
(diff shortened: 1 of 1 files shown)


gc00 (approved):
I'm approving this.

But I'll note that on github, the test for dlopen1 is failing.
I tried this on WSL (Ubuntu 22.04), and it failed there, too.  But it seems that `./dlopen1` is failing even when I run it natively, with:
>  dlopen failed: libdlopen-lib1.so: cannot open shared object file: No such file or directory

or when I change the relative pathname in dlopen() to absolute, I see:
> dlopen failed: /home/gene/dmtcp.git/test/dlopen1: cannot dynamically load position-independent executable

When I modify dlopen1.c to use absolute pathnames, then the DMTCP test works fine.  It appears that the recent Linux policies don't allow us to set LD_LIBRARY_PATH when donig 'dlopen()' with relative pathnames.  We should fix this test, separately.

dahongli (approved):

PR #1061, opened 2023-06-25
Use virtual-tid in libc pthread struct (fixes recursive mutexes, etc.)

karya0:
Fixes recursive/error-checking mutex support.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -323,7 +323,7 @@ restore_brk(VA saved_brk, VA restore_begin, VA restore_end)
     // Now unmap the just mapped extended heap. This is to ensure that we don't
     // have overlap with the restore region.
     if (mtcp_sys_munmap(current_brk, new_brk - current_brk) == -1) {
-      MTCP_PRINTF("***WARNING: munmap failed; errno: %d\n", mtcp_sys_errno);
+      DPRINTF("***WARNING: munmap failed; errno: %d\n", mtcp_sys_errno);
     }
   }
   if (new_brk != saved_brk) {
@@ -333,10 +333,10 @@ restore_brk(VA saved_brk, VA restore_begin, VA restore_end)
               new_brk, saved_brk);
     } else {
       if (new_brk == current_brk) {
-        MTCP_PRINTF("error: new/current break (%p) != saved break (%p)\n",
+        DPRINTF("error: new/current break (%p) != saved break (%p)\n",
                     current_brk, saved_brk);
       } else {
-        MTCP_PRINTF("error: new break (%p) != current break (%p)\n",
+        DPRINTF("error: new break (%p) != current break (%p)\n",
                     new_brk, current_brk);
       }
 


gc00:
@karya0 , For PR #1061, note that the CI is failing on `glibc_pthread.cpp`.  The reason is that you have an extra `3include <stdatomic.h>` after your `#ifdef .. #endif`.

gc00:
@karya0 , I've now reviewed the first 6 commits out of 9. for PR #1061.  I'll finish the review later.

gc00:
@karya0 , I have now reviewed the first 8 of the 9 commits.  I will review the last commit next.

I'm happy to begin a discussion of my minor comments on the first 8 commits even while I look closely at the 9th commit.

src/plugin/timer/timerwrappers.cpp
```diff
@@ -124,13 +125,14 @@ clock_getcpuclockid(pid_t pid, clockid_t *clock_id)
 extern "C" int
 pthread_getcpuclockid(pthread_t thread, clockid_t *clock_id)
 {
-  DMTCP_PLUGIN_DISABLE_CKPT();
+  // We need to acquire exclusive lock here because the corresponding Pid plugin
```
gc00, line 128: Picky, but:
'to acquire _an_ exclusive lock'
'plugin wrapper requires _an_ exclusive lock'

gc00 (approved):
I'm requesting some changes.  But I'll approve this in advance.

QUESTION:  We're considering an Apache license for DMTCP.  glibc uses the LGPL license.
                      Are we copying some glibc code here, beyond a few lines here and there?
                      In the worst case, it would be easy to rewrite any lines taken from glibc.  So, I'm now worried.

test/autotest.py
```diff
@@ -606,12 +592,12 @@ def testKill():
...
   def testCheckpoint():
     #start checkpoint
-    coordinatorCmd(CKPT_CMD)
+    runDmtcpCommand(CKPT_CMD)
 
     #wait for files to appear and status to return to original
     # b'Kc' input to dmtcp_coordinator is equivalent to 'dmtcp_command -kc'
     WAITFOR(lambda: getNumCkptFiles(ckptDir)>0 and \
-                 (CKPT_CMD == b'Kc' or doesStatusSatisfy(getStatus(), status)),
+                 (CKPT_CMD == "Kc" or doesStatusSatisfy(getStatus(), status)),
```
gc00, line 600: Same here. We don't need to break Python 2.7 for no special reason.
'Kc' and "Kc" are equivalent. But removing b is not equivalent.
Please do: b'Kc' -> b"Kc".

test/autotest.py
```diff
@@ -626,7 +612,7 @@ def testCheckpoint():
           "unexpected number of checkpoint files, %s procs, %d files"
           % (str(status[0]), numFiles))
 
-    if SLOW > 1 and CKPT_CMD != b'Kc':
+    if SLOW > 1 and CKPT_CMD != "Kc":
```
gc00, line 615: Same here. We don't need to break Python 2.7 for no special reason.
'Kc' and "Kc" are equivalent. But removing b is not equivalent.
Please do: b'Kc' -> b"Kc".

test/autotest.py
```diff
@@ -860,7 +846,7 @@ def saveResultsNMI():
 # program will try to unlink the file once again, but the unlink operation will
 # fail, causing the test to fail.
 old_ckpt_cmd = CKPT_CMD
-CKPT_CMD = b'Kc' # Equivalent to 'dmtcp_command -kc'
+CKPT_CMD = "kc" # Equivalent to 'dmtcp_command -kc'
```
gc00, line 849: Same here. We don't need to break Python 2.7 for no special reason.
'Kc' and "Kc" are equivalent. But removing b is not equivalent.
Please do: b'Kc' -> b"Kc".

Also, why change `Kc` to `kc` (lower-case)?

src/dmtcp_coordinator.cpp
```diff
@@ -239,6 +240,14 @@ static pid_t _nextVirtualPid = INITIAL_VIRTUAL_PID;
 static int theNextClientNumber = 1;
 vector<CoordClient *>clients;
 
+static inline void
+ltrim(string &s)
+{
+  s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](unsigned char ch) {
```
gc00, line 246: Can we rewrite this as a standard iteration.  Making this a lambda function just helps to ensure that the majority of people with a B.S. in computer science will have to start googling to figure out what this code means.

And using a lambda function does not seem to add greater clarity.  When I think about trimming space from the beginning of a string, my first thought is not that i need to write an extra helper function in order to reduce the complexity of the task.  :-)

src/dmtcp_coordinator.cpp
```diff
@@ -94,15 +94,16 @@ using namespace dmtcp;
```
gc00, line 94: I think this commit ("Added prompt for dmtcp_coordinator.") seems to be about using a different programming style (e.g., lambda function instead of iteration;  'if...else if' instead of 'switch...case').

If this is the case, let's break this out into an independent PR, so it doesn't delay the many other important commits.

If there is something here that changes the semantics, can you isolate that into a separate commit for the current PR, and push the remaining changes into a separate PR?

Thanks.

gc00 (changes requested):
I've now reviewed everything.  And there are a lot of important commits here that should not be delayed.

Because of the many changes, I'd like to look at this again, after the changes are make.

I promise to re-review this quickly after the changes, but  If I forget to re-review it quickly, please do remind me.

gc00 (approved):
I'm approving this in advance.  Please ntoe that the check is failing due to `syscall-tester`.

I'm sure it's something simple, but I didn't have time to look at it.

If you can fix syscall-tester in one more commit, that's nice.  If you don't have the time, then you can push it in as is, and we'll examine syscall-tester later.  :-(

PR #1063, opened 2023-06-28
dmtcp_launch: Add --mpi flag: personality(ADDR_NO_RANDOMIZE)

gc00:
* When the '--mpi' flag is set, do 'personality(ADDR_NO_RANDOMIZE)'.
 * When restarting an MPI application, it is easier if the libraries, stack, vdso, etc., were loaded all at the same address for each MPI process/rank, prior to the checkpoint.


dahongli (approved):
The change looks good to me. Thanks Gene!

PR #1075, opened 2023-08-24
readdmtcp.sh: Prefer mtcp_restart of same dir tree

gc00:
Some sites might have an old system-wide copy of mtcp_restart in `/usr/bin/mtcp_restart`.

When invoking `DMTCP_ROOT/util/readdmtcp.sh`, the script now checks for a relative binary in `DMTCP_ROOT/util/../bin/mtcp_restart`, instead of blindly trying `which mtcp_restart` first.

### DIFF
--- a/util/readdmtcp.sh
+++ b/util/readdmtcp.sh
@@ -16,13 +16,13 @@ fi
 
 dir=`dirname $0`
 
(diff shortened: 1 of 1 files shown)


JainTwinkle (approved):
LGTM :)

PR #1077, opened 2023-09-11
Unmap mtcp_restart and heap areas explicitly.

karya0:
We want to unmap mtcp_restart and heap regardless of what a mtcp_restart plugin might suggest.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -92,6 +92,7 @@ static int hasOverlappingMapping(VA addr, size_t size);
 static int mremap_move(void *dest, void *src, size_t size);
 static void remapMtcpRestartToReservedArea(RestoreInfo *rinfo);
 static void mtcp_simulateread(int fd, MtcpHeader *mtcpHdr);
+static void unmap_one_memory_area_and_rewind(Area *area, int mapsfd);
 static void unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo);
 static void compute_vdso_vvar_addr(RestoreInfo *rinfo);
 
@@ -694,6 +695,26 @@ compute_vdso_vvar_addr(RestoreInfo *rinfo)
 
 NO_OPTIMIZE
 static void
+unmap_one_memory_area_and_rewind(Area *area, int mapsfd)
+{
+  int mtcp_sys_errno;
+  DPRINTF("***INFO: munmapping (%p..%p)\n", area->addr, area->endAddr);
+  if (mtcp_sys_munmap(area->addr, area->size) == -1) {
+    MTCP_PRINTF("***WARNING: %s(%x): munmap(%p, %d) failed; errno: %d\n",
+                area->name, area->flags, area->addr, area->size,
+                mtcp_sys_errno);
+    mtcp_abort();
+  }
+
+  // Since we just unmapped a region, the size and contents of the
+  // /proc/self/maps file has changed. We should rewind and reread this file to
+  // ensure that we don't miss any regions.
+  // TODO(kapil): Create a list of all areas that we want to munmap and then unmap them all at once.
+  mtcp_sys_lseek(mapsfd, 0, SEEK_SET);
+}
+
+NO_OPTIMIZE
+static void
 unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
 {
   /* Unmap everything except this image, vdso, vvar and vsyscall. */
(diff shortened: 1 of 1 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -692,6 +692,23 @@ compute_vdso_vvar_addr(RestoreInfo *rinfo)
   mtcp_sys_close(mapsfd);
 }
 
+NO_OPTIMIZE
+static void
+unmap_one_memory_area(Area *area, int mapsfd)
```
JainTwinkle, line 697: Let's declare this function in the beginning along with other internal routines.

JainTwinkle (approved):
See my comment about the declaration. The rest LGTM.

src/mtcp/mtcp_restart.c
```diff
@@ -738,21 +755,19 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo)
...
+    } else if (mtcp_strcmp(area.name, "[heap]") == 0) {
+      // Unmap heap.
+      unmap_one_memory_area(&area, mapsfd);
+    } else if (mtcp_strendswith(area.name, BINARY_NAME) ||
+               mtcp_strendswith(area.name, BINARY_NAME_M32)) {
+      // Unmap original mtcp_restart regions.
+      unmap_one_memory_area(&area, mapsfd);
     } else if (mtcp_plugin_skip_memory_region_munmap(&area, rinfo)) {
       // Skip memory region reserved by plugin.
       DPRINTF("***INFO: skipping memory region suggested by plugin (%p..%p)\n",
```
gc00, line 767: While we're cleaning up, can we do a slight change?
FROM:
> DPRINTF("***INFO: skipping memory region suggested by plugin (%p..%p)\n",
TO:
>DPRINTF("***INFO: skipping memory region as requested by plugin (%p..%p)\n",

src/mtcp/mtcp_restart.c
```diff
@@ -692,6 +692,23 @@ compute_vdso_vvar_addr(RestoreInfo *rinfo)
...
+  int mtcp_sys_errno;
+  DPRINTF("***INFO: munmapping (%p..%p)\n", area->addr, area->endAddr);
+  if (mtcp_sys_munmap(area->addr, area->size) == -1) {
+    MTCP_PRINTF("***WARNING: %s(%x): munmap(%p, %d) failed; errno: %d\n",
+                area->name, area->flags, area->addr, area->size,
+                mtcp_sys_errno);
+    mtcp_abort();
+  }
+
+  // Rewind and reread maps.
```
gc00, line 708: I know that the comment `Rewind and reread maps.` was there earlier.

But let's take this opportunity to document our design decisions.  Normally, we just view the memory regions in order, and unmap them.  Why do we unmap and then start reading again from the beginning?
Perhaps you can give a _brief_ example where this is needed.

src/mtcp/mtcp_restart.c
gc00: I agree:
> Let's declare this function in the beginning along with other internal routines.

gc00, line 697: Maybe call this `unmap_one_memory_area_and_reread`.

See my other comment, about documenting the reason for re-reading.

gc00 (changes requested):
While you're polishing this file, I suggest changing MTCP_RESTART, etc., by adding a `/` so that `strendswith` will not confuse it:
```
#define BINARY_NAME     "/mtcp_restart"
#define BINARY_NAME_M32 "/mtcp_restart-32" 
```

src/mtcp/mtcp_restart.c
```diff
@@ -692,6 +693,26 @@ compute_vdso_vvar_addr(RestoreInfo *rinfo)
...
+  if (mtcp_sys_munmap(area->addr, area->size) == -1) {
+    MTCP_PRINTF("***WARNING: %s(%x): munmap(%p, %d) failed; errno: %d\n",
+                area->name, area->flags, area->addr, area->size,
+                mtcp_sys_errno);
+    mtcp_abort();
+  }
+
+  // Since we just unmapped a region, the size and contents of the
+  // /proc/self/maps file has changed. We should rewind and reread this file to
+  // ensure that we don't miss any regions.
```
gc00, line 711: How could we miss a region?  We're calling unmap_one_memory_area in order on the areas.
Is this simply "defensive programming", or is there a scenario where we would miss something?  We're not closing and re-opening mapsfd -- so, lseek to offset 0 might simply show us the same stale data.

gc00:
@karya0 , Sorry to slow you down on this.  But I'm truly confused about why we need to rewind to offset 0.  If this is simply defensive programming due to some unknown danger, then please add the phrase "defensive programming" in the comment.  If you know of a scenario where we need this, please add it to the comment.

    And finally, I _do_ want to change the name to `unmap_one_memory_area_and_rewind` (or something similar.  Someone casually reading the code would miss this, since the definition of the function is not close to where it's actually used.

Thanks.

gc00 (approved):
Perfect!  Thanks very much.

PR #1068, opened 2023-07-12
WIP: gdb-dmtcp-utils: Create command: (gdb) load-symbols # for GDB v8, etc.

gc00:
**[ Could someone test this with split processes for GDB version 8 or later?  I'll wait for that test  and review before pushing this in.  This should show all symbols of the upper half in the stack after: 
  (gdb) source gdb-dmtcp-utils;
 (gdb) load-symbols
]**

 * Somewhere around GDB-v8 (or some Linux kernel), (gdb) add-symbol-files-all no longer sufficed to see symbols in GDB.
 * GDB-v8 introduced the option '[-o OFF]' for the GDB commands 'symbol-file' and 'add-symbol-file'.  'load-symbols'uses '[-o OFF]' with 'symbol-file EXEC_FILE' to replace all symbols. With no argument, 'load-symbols' guesses the EXEC_FILE.
 * A later commit will also support 'add-symbol-file -o OFF' for the sake of debugging split processes on restart: (gdb) symbol-file -o OFF UPPER_HALF_EXEC_FILE (gdb) add-symbol-file -o OFF LOWER_HALF_LIBRARY.so

PR #1078, opened 2023-09-14
Pass curated environ to mtcp_restart.

karya0:
This would allow us to publish ProcessInfo::KVDB dictionary to be consumed by mtcp_restart and mtcp_restart plugins.


src/dmtcp_restart.cpp
```diff
@@ -404,8 +404,42 @@ getMtcpArgs()
   return mtcpArgs;
 }
 
+vector<string>
+getMtcpEnviron(RestoreTarget *restoreTarget)
```
gc00, line 408: I suggest, everywhere:
Envrion -> Evironment

The 'main' routine calls this 'envp', and in most other cases, it is called 'environment'.  It's confusing to add a third term.

src/mtcp/mtcp_util.c
```diff
@@ -73,6 +73,37 @@ unsigned long mtcp_strtol (char *str)
   return v;
 }
 
+unsigned long mtcp_strtol (const char *str)
+{
+  return (unsigned long) mtcp_strtoll(str);
+}
+
+void mtcp_utoa(char* buffer, unsigned int n)
```
gc00, line 81: How about:
mtcp_utoa -> mtcp_ultoa

'ltoa' is easily recognized as the opposite of 'atol'.  In 'printf' formants, we use %ud to represent an unsigned decimal value.  By analogy, 'ul' should be 'unsigned long'.

src/mtcp/mtcp_util.h
```diff
@@ -162,7 +162,10 @@ void mtcp_printf(char const *format, ...);
 ssize_t mtcp_read_all(int fd, void *buf, size_t count);
 int mtcp_readfile(int fd, void *buf, size_t size);
 void mtcp_skipfile(int fd, size_t size);
-unsigned long mtcp_strtol(char *str);
+unsigned long mtcp_strtol(const char *str);
+unsigned long long mtcp_strtoll(const char *str);
+void mtcp_utoa(char* buffer, unsigned int n);
+void mtcp_itoa(char* buffer, int n);
```
gc00, line 168: Even though you don't need it here, I've sometimes missed:
> mtcp_strcmp

We have 'mtcp_strcpy', but not 'mtcp_strcmp'.  It's often a convenient utility.

src/mtcp/mtcp_util.c
```diff
@@ -42,10 +42,10 @@
 #include "mtcp_util.h"
 #include "../membarrier.h"
 
-unsigned long mtcp_strtol (char *str)
+unsigned long long mtcp_strtoll (const char *str)
```
gc00, line 45: We seem to have a mixed formatting convention in this file.  Sometimes it's `FUNCTION (arg)` and dometimes it's `FUNCTION(arg)`.  It's optional, but you might want to consider an extra commit that cleans this up everywhere:
   `FUNCTION (arg)` ->  `FUNCTION(arg)`

gc00 (changes requested):
As seen in the comments, I'd like to omit the new functions getMtcpEnvironment(t) and StringVectorToCharPtrVector().

There is an optional suggestion to unify the formatting of function definitions and function calls in mtcp_util.h in an additional commit.  But the request for changes is based on the two new functions mentioned above.

src/dmtcp_restart.cpp
```diff
@@ -404,8 +404,42 @@ getMtcpArgs()
   return mtcpArgs;
 }
 
+vector<string>
+getMtcpEnvironment(RestoreTarget *restoreTarget)
```
gc00, line 408: This is an example of highly non-local program design.  I see that you want this in combination with the new design of restart plugins.  But let's look for something better.  Why can't we simply take the entire environment from the global 'environ' environment variable, for example.

1. The purpose of this function seems to be to hardwire in certain environment variables, while removing all other environment variables.
2. So, when testing or debugging, if we call 'dmtcp_restart', we see one set of environment variables, and if we call 'mtcp_restart' in the 'mtcp' directory we see different environment variables.  This is confusing when extending 'mtcp_restart' in the mtcp directory and then testing it.
3. There is no harm in keeping the extraneous environment variables.  We are going to call 'mtcp_restart', which will restore the checkpoint image.  So, all extraneous environment variables will disappear, except those which 'mtcp_restart' treats specially (like 'DMTCP_RESTART_PAUSE'.
4. If you really want the reader to be aware of special-purpose environment variables like 'DTMCP_RESTART_PAUSE', then adding a _local_ comment near where the reader sees it (e.g., in the mtcp dir or in the restart_plugin dir of MANA) is a better option.

(While I like the benefits of C++, this is one place where I feel that C++ encourages a non-local design paradigm.  I feel that this is a misfeature of C++.
    Some examples are:  a C++ method might be defined in the .h or the .cpp file.  One should look in both places.  A namespace or class may sometimes be omitted.  Look elsewhere to discover the true signature (because this saves a small amount of typing that a good editor would be glad to insert automatically).  An operator may be what it looks like, or there may be a separate file redefining it.  There is no syntactic indicator.  (This is different from C macros, where macros are capitalized.)
    This non-local design happens at the most basic level of C++, even before we talk about inheritance. )

Anyway, enough complaints from me.  It's probably because this is late at night.  But I do want to avoid more non-local design.

src/dmtcp_restart.cpp
```diff
@@ -461,7 +483,7 @@ runMtcpRestart(int fd, ProcessInfo *pInfo)
   mtcpArgs.push_back((char *) fdBuf.c_str());
 
   mtcpArgs.push_back(NULL);
-  execvpe(mtcpArgs[0], &mtcpArgs[0], environ);
+  execvp(mtcpArgs[0], &mtcpArgs[0]);
```
gc00, line 486: I see that codefactor.io warned about including an untested args (or untested environment variables) in the execvp.  I don't think we have to worry.  We could be influenced by the user injecting DMTCP_RESTART_PAUSE, and the DEBUG variant.  But that is part of the intended design.  So, I'm not worried here.

gc00 (approved):
@karya0 , Thanks for the changes.  I'm approving it.

In addition, I'm sure your intention is to update MANA to use the most recent DMTCP.  Before you do that, could you please review:
* PR #1064 -<br> If a coordinator has no client process for 12 hours, the coordinator exits;  This fixes an annoying problem where someone sits on the default DMTCP 7779 port with a stale process.  It can also hurt our reputation if a user falsely ascribes this to a DMTCP bug or misfeature.
* PR #1068 -<br> Latest `gdb-dmtcp-utils` version; much improved and already pushed into the corresponding MANA utils directory

Especially, PR #1064 is important to push in.
And after reviewing, please give me time to take a final look and push in the PRs.  Then when you update tMANA's DMTCP submodule, these will be inherited.

PR #1079, opened 2023-09-18
WIP: Use new stack early, before restart_fast_path

gc00:
This is important for MANA, to avoid address conflicts between the temporary stack of mtcp_restart and restoring the upper half memory inside mtcp_plugin_hook.


JainTwinkle:
@gc00, I see that you have only added code but not moved from later to earlier in the execution. Will the existing code again try to relocate the stack later in the mtcp_restart execution?

PR #1082, opened 2023-09-22
WIP: Reclaim the restore buffer region before anyone can allocate in the reserved area

JainTwinkle:
A jalloc allocation can result in an mmap call. Although it is rare, the kernel can place the new region in the free space between the heap and stack of mtcp_restart i.e., in the restoreBuf region. It is better to reclaim the restore buffer before it's too late.

### DIFF
--- a/src/threadlist.cpp
+++ b/src/threadlist.cpp
@@ -28,6 +28,7 @@
 #include "uniquepid.h"
 #include "util.h"
 #include "shareddata.h"
+#include "processinfo.h"
 
 // For i386 and x86_64, SETJMP currently has bugs.  Don't turn this
 // on for them until they are debugged.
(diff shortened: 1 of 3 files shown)


src/threadlist.cpp
```diff
@@ -644,6 +645,13 @@ ThreadList::waitForAllRestored(Thread *thread)
 
     JTRACE("before DmtcpWorker::postRestart()");
 
+    // A jalloc allocation can result in an mmap call. Although it is rare,
+    // the kernel can place the new region in the free space between the
+    // heap and stack of mtcp_restart i.e., in the restoreBuf region.
+    // It is better to reclaim the restore buffer now before it's too late.
+    // FIXME: can we do any better?
```
gc00, line 652: I like this PR a lot.  I can't imagine a cleaner design.  I suggest removing the FIXME.

gc00 (commented):
Since it's still a 'WIP', I'm reviewing this as a comment.  Please tell me whenever you like, and I (or someone else) can change this to "Approve".
In the meantime, I will modify our testing branch for MANA to point to this PR for the DMTCP submodule.

JainTwinkle:
> Hi @JainTwinkle: I believe #1085 subsumes this change. Please take a look.

Yes, it does! I can close this PR. This was anyway an ad-hoc solution.

After looking at your PR, I realized that the stack of mtcp_restart was never mapped with MAP_GROWSDOWN. This means we can reserve the whole area. While working on this solution, I was under the impression that we intentionally left some space in case mtcp_restart's stack grows.

PR #1083, opened 2023-09-23
Modularized dmtcp_restart and mtcp_restart.

karya0:
This would allow custom binaries to be created by MANA without needing any relinking within DMTCP.

It subsumes #1084 which should be reviewed and merged before this PR.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -38,6 +38,7 @@
  */
 
 #define _GNU_SOURCE 1
+#include <elf.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <sched.h>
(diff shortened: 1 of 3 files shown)


gc00:
I'll review this carefully later.  But:
1 The `pthreads` test is failing.
2. As a reminder, `src/mtcp/stdlibfnc.c` seems to be an extraneous file.  It is linked into `mtcp_restart`, but I don't see why it's needed.  It replaces `__libc_start_main` by a trivial function.  If we're going to keep it, can you add a comment why it's needed?

src/dmtcp_restart_main.cpp
```diff
@@ -0,0 +1,36 @@
...
+ *  <http://www.gnu.org/licenses/>.                                         *
+ ****************************************************************************/
+
+#include "dmtcp_restart.h"
+
+using namespace dmtcp;
+
+#define MTCP_RESTART_BINARY "mtcp_restart"
+
+int
```
gc00, line 28: Some comments would help.  Normally, dmtcp_XXX.cpp is for an executable of the same name.

Here, you're breaking the convention, and choosing dmtcp_restart_main.cpp.  This is confusing.

As an alternative, how about naming this file:  `dmtcp_restart.cpp`  (to keep the convention)

Then, you can rename the current `dmtcp_restart.cpp` into something like `dmtcprestartinternal.cpp` or `libdmtcprestart.cpp` (or else `dmtcp_librestart.cpp`, but I prefer the first version, since it's used for `libdmtcprestart.a`), since you're now going convert all of that code into a library that can be linked in more than one way.  After all, you are building a libdmtcprestart.a, similar to `libdmtcpinternal.a`.

gc00 (approved):
See PR #1095 for a comment on which tests are failing for Rocky Linux 8.8, and for Ubuntu 20.04 (WSL).  However, this PR is not the cause of the failed tests.

Please rename `dmtcp_restart_main.cpp` to `dmtcp_restart.cpp`, etc., along with the comments on that issue.  But after that, it's fine to approve this.  (Sorry it took me so long to review this.  The file naming issue was something that initially confused me.)

PR #1090, opened 2023-10-13
Fix "stale timeout" again (commit d487881)

gc00:
Add another fix for "stale timeout"

* Also add FIXME comment about cleaner future logic

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -1470,7 +1470,24 @@ calcLocalAddr()
 static void
 resetCkptTimer()
 {
-  if (theCheckpointInterval > 0) {
+  // FIXME:  Fix timeout logic to include setStaleTimeout, resetStaleTimeout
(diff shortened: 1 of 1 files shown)


jiamingz9925:
@gc00  There are two things during my test:
1. This commit does not resolve the issue that the checkpoint did not happen when we set '--interval(-i)' flag for dmtcp_coordinator.
2. This commit seems introduce more issues to the checkpoint procedure. Manually checkpoint with mana_status seems to hang with this commit.

I realized that the changes are on updateCheckpointInterval() and resetCkptTimer(). Correct me if I am wrong, but I doubt if I set the --interval flag for coordinator these two functions should be involved at least in the beginning of first checkpoint. But the checkpoint does not happen at all currently with --interval flag on coordinator. Maybe there is some other methods we should investigate?

gc00:
@karya0 ,
    Could you review this PR #1090.  It fixes a bug/misfeature that I had created in a previous PR for the ``--stale`` timeout.  Thanks.

PR #1091, opened 2023-10-29
Jalloc Improvements

karya0:
* Use fewer mmap calls.
* Use run-time object construction instead of static construction.


gc00 (approved):
Thanks for the extra comment, reviewing the logic of different level arenas.
LGTM.

PR #1092, opened 2023-11-07
Coordinator prompt with one-command-per-line interface

karya0:
This is a step in the direction of bringing the coordinator interface closer to the standard readline-like interface where an entire line is interpreted as a single command as opposed to potentially multiple commands. For example, earlier, you could have typed `<l><s><newline>` in the coordinator, and it would first execute `<l>` and then `<s>`. This caused us to use things like "k" for kill but "K" as a prefix for kill after ckpt and so on. With this change, "ls" will be considered an invalid command. If one wants to execute "l" and "s" both, they will do `<l><newline><s><newline>`. 

A semi-related change is to use dmtcp_command in autotest.py to interact with the coordinator. This would allow us to improve autotest.py by having dmtcp_command output machine-readable output.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -94,15 +94,16 @@ using namespace dmtcp;
 
 static const char *theHelpMessage =
   "COMMANDS:\n"
-  "  l : List connected nodes\n"
-  "  s : Print status message\n"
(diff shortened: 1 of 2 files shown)


gc00 (approved):
LGTM.  Thanks for isolating this as a separate PR.

PR #1095, opened 2023-11-30
Initiate ckpt on receiving ckpt-signal.

karya0:
This PR makes it possible for a user to use `kill -SIGUSR2 <pid>` to initiate a checkpoint in addition to the existing methods of using `dmtcp_coordinator` terminal and `dmtcp_command`.

This functionality also allows a resource manager such as slurm to request checkpoint by sending a custom signal (e.g., 15) to all processes in the distributed computation. Once a process receives this custom signal (along with `--ckpt-signal 15` flag for `dmtcp_launch`), it will request the coordinator to initiate a ckpt. Only first request will be honored, the rest will be safely ignored by the coordinator.

### DIFF
--- a/src/threadlist.cpp
+++ b/src/threadlist.cpp
@@ -12,22 +12,22 @@
 #include <sys/prctl.h>
 #endif  // if LINUX_VERSION_CODE >= KERNEL_VERSION(2, 6, 11) ||
 // defined(HAS_PR_SET_PTRACER)
-#include "jalloc.h"
-#include "jassert.h"
 #include "ckptserializer.h"
+#include "coordinatorapi.h"
 #include "dmtcpalloc.h"
 #include "dmtcpworker.h"
+#include "jalloc.h"
+#include "jassert.h"
 #include "mtcp/mtcp_header.h"
 #include "pluginmanager.h"
 #include "shareddata.h"
 #include "siginfo.h"
 #include "syscallwrappers.h"
 #include "threadlist.h"
-#include "tls.h"
 #include "threadsync.h"
+#include "tls.h"
 #include "uniquepid.h"
 #include "util.h"
-#include "shareddata.h"
 
 // For i386 and x86_64, SETJMP currently has bugs.  Don't turn this
 // on for them until they are debugged.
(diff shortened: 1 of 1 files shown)


src/threadlist.cpp
```diff
@@ -572,6 +572,13 @@ stopthisthread(int signum)
    * sigaction(STOPSIGNAL, stopthisthread) to discard all pending signals.
    */
 
+  // If the user sent a ckpt signal, it can be caught by either ckpt-thread or
```
gc00, line 575: In above line, let's clean up the comment now;
1.  Case 3 above: insert newline before it
2. Case 4 above:  ckpt-threads -> ckpt-thread

The line just above here, is not clear.  "If the user sent a ckpt signal":  The "user" can send a command to the coordinator for a checkpoint.  The "user" should not directly send a ckpt signal to some random thread.  Does this comment mean that we are in Case 4 here.  If so, please just say that we're in Case 4, above.

gc00 (approved):
I tested on Ubuntu 20.04.  The CI is saying that pthread5 and syscall-tester both failed.
But it is not failing here.  For Ubuntu 20.04 (WSL), I'm seeing:
    popen1 and hellompich-nl failing.
But it's clearly not because of this PR.

On the other hand, on Rocky Linux 8.8, I'm seeing:\
Test ssh1:
    "Coordinator not found. Please check port and host." repeatedly.

Similarly, on syscall-tester, I'm seeing:
> syscall-tester      ckpt:Coordinator not found. Please check port and host.     PASSED; rstr:PASSED -> ckpt:PASSED; rstr:Coordinator not found. Please check port and host.                                                                     PASSED  

Anyway, we should pass this PR.

gc00:
**PLEASE SEE MY COMMENT ABOVE.  COULD YOU FIX UP THE COMMENT, BEFORE PUSHING THIS IN?  THANKS.**

(In general, a description for the PR would have been nice.  :-) )

gc00:
Please note that this was pushed in on Aug. 13.  But I first expanded some of the comments.
So, the original commit message was PR #1095 (now PR #1121 ).
> Initiate ckpt on receiving ckpt-signal

It is now:
>  Coord initiates ckpt if user receives ckpt-signal
with commit hash:  835b557181713c226deb78cddac703c70bb2b9ed 

Another way to convince yourself is to cherry-pikc the commit of this PR on top of 'main'.  Git will say that it's an empty commit.

PR #1102, opened 2024-03-18
Restore deleted file if opened in O_WRONLY.

karya0:


src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -371,7 +368,15 @@ FileConnection::refill(bool isRestart)
         }
       }
       tempfd = openFile();
+    } else if (_fcntlFlags & O_WRONLY) {
+      // File doesn't exist. If it's a WRONLY file, we'll create a new one.
+      tempfd = _real_open(_path.c_str(), O_CREAT|O_WRONLY|O_TRUNC);
+      ASSERT_NE(-1, tempfd);
```
gc00, line 374: Isn't `JASSERT(tmpfd != -1)` easier to read than `ASSERT_NE()`?  Also, it seems confusing that we have two variants:  `JASSERT` and `ASSERT_XX`, in addition to the normal `assert()`.

src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -371,7 +368,15 @@ FileConnection::refill(bool isRestart)
...
+      ASSERT_EQ(0, ftruncate(tempfd, _st_size));
```
gc00, line 375: Isn't `JASSERT(ftruncate(tempfd, _st_size) == 0)` easier to read than ASSERT_EQ()? Also, it seems confusing that we have two variants: JASSERT and ASSERT_XX, in addition to the normal assert().  Also, ASSERT_XX reminds me of Fortran's `.EQ.`, `.NE.`, etc.

gc00 (commented):
If a write-only file was deleted between checkpoint time and restart time, do we really want to truncate it to zero?
I would have assumed that we would prefer to do something like `./dmtcp_launch --checkpoint-open-files`.  If we had an open file descriptor to the file at ckpt time, then this seems like the right thing to do.

If we didn't have the file open anymore, then we need a new feature to save the file only.  I'm proposing a feature like that based on precious files.  See my comments in the latest version of `syscall-tester.c`.  (Search on "precious" in that file.)

gc00 (approved):
Okay, I'm approving this, as is.

Whenever we can talk, I can explain to you a concept of precious files that I'm thinking about.  We saw an application for which a directory kept many small files ans a kind of ad hoc database.  Not all files had an open file descriptor, but the application still wnated them preserved.  I'm thinking of a precious-files design that allows users to cleanly declare their intent.

I had to implement something just like that in syscall-tester.c, to fix a bug that we were seeing.

PR #1105, opened 2024-04-11
Add alignas(16) for dwcas in jalloc.cpp

gc00:
This is an easy one-line code to be reviewed.

Recall that dwcas (double-word-compare-and-swap) is used in our lock-free algorithm for a stack.

In jalloc.cpp, we use bool_atomic_dwcas .  On x86_64, with gcc, it can probably do a workaround if its arguments are not double-world aligned (128-bit aligned).  But we can't guarantee it for all CPUs and all compilers.  And misaligned makes it slower.

This fixes it, by ensuring that all `struct StackHead` are double--word aligned.

### DIFF
--- a/jalib/jalloc.cpp
+++ b/jalib/jalloc.cpp
@@ -275,7 +275,8 @@ class JFixedAllocStack
         char buf[N];
       };
     };
-    struct StackHead {
+    // alignas(16) to support 128-bit StackHead, used in dwcas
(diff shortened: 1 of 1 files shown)


gc00:
@xuyao0127 or @karya0 ,
    Could you review this old PR now?  It's trivial.  (And now, the newest DMTCP says "All checks have passed"

PR #1106, opened 2024-04-14
jalloc.h: getAllocArenas: JALIB_ALLOCATOR always

gc00:
* Bug fix for commit e64459ccf
 * If JALIB_ALLOCATOR is not defined, it fails to compile.  We move `getAllocArenas` outside of the conditionalizatio


aayushi363 (approved):
Looks good

PR #1108, opened 2024-07-16
Fix syscall-tester.c; Remove BasicStat test

gc00:
* In BasicState, it did: fd=open(file); close(fd) and unlink(file) When checkpointing between close() and unlink(), resume would still unlink 'file'.  And then restart fails to do 'unlink(file)'
 * See comment in syscall-tester.c about TODO: // TODO:  DMTCP should offer:  `dmtcp_precious_file(file)`
     //        At checkpoint time, even if the fd is closed, if the
     //        file still exists on disk, then pretend that
     //          ./dmtcp_launch --checkpoint-open-files ...
     //        was used, but only for this file.

This bug is causing DMTCP to fail its status check on github.  Let's review this quickly, so as to get the testing on github working again.

### DIFF
--- a/test/syscall-tester.c
+++ b/test/syscall-tester.c
@@ -4838,7 +4838,15 @@ testall()
 
     /*            {BasicIOV, "BasicIOV: Basic vector reads and writes"},*/
     { BasicFreopen, "BasicFreopen: Does freopen return something sensible?" },
-    { BasicStat, "BasicStat: Does [fs]tat return correct simple info?" },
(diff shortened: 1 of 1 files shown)


xuyao0127 (approved):
LGTM

gc00:
And a quick comment.
Apparently syscall-tester.c calls `unlink()` a lot (and also `rename()`.  So, this bug would appear a lot.

We have two choices:
 1.  Implement dmtcp_precious_file(); OR
 2.  Force DMTCP to exit on checkpoint, and not to try to resume (maybe with a plugin for the RESUME event?).

PR #1112, opened 2024-07-19
Fix compiler warnings; bug fix for vfork1, problem w/ plugin-example-db

gc00:
The github commits say it all (with oldest commits first):

* Clean up autotest for syscall-tester
* Fix some compiler warnings; Polish autotest.py
* plugin-example-db test never did as intended
* Fix vfork[12] tests: They had a bug from the start
* Lower timeout for pthread[345] tests
  (Empirically, the continuous integration seems to succeed more often with this.)


xuyao0127 (approved):
LGTM

PR #1113, opened 2024-07-22
Improve gdb dmtcp utils

gc00:
Two commits for util/gdb-dmtcp-utils
  1. load-symbols-library now handles addresses or filenames of executable file, not just for library files.
  2. load-symbols-library now handles a default with no arguments:
      With no arguments, it will now load symbols from the newest call frame that does not yet have loaded symbols (that does not yet show a function name).


xuyao0127 (approved):
LGTM

PR #1118, opened 2024-07-31
Fix for aarch64 (ARM64)

gc00:
DMTCP-3.0.0 worked for aarch64.  But since then, there were some commits that broke aarch64, and some commits that exposed long-standing bugs in DMTCP on the aarch64 platform.

This should fix all of them.  It has been tested on other people's an aarch64 computer,

This no longer reverts the VDSO commit.  Instead, it uses the commit for tid's from @karya0 .


gc00:
@karya0 ,
    Can you review this PR?  It now has your commit for `[Update glibc_pthread.h for ARM and AARCH64.]`.  And it no longer does the vdso revert.
    All of these commits should be non-controversial.  I think it should fix ARM64 now.  But even if we discover some small bug later, at least this brings us very close to it working.

gc00:
Hi Kapil,
    How about in this case, I just revert DMTCP master and push in the PRs in the order that I had planned?

I guess at Microsoft, they review a PR in a day or tow, and so the approver  can push it in without consulting the author.

But we don't have those resources for DMTCP.  If we're going to push things in three weeks later, and juggle five different PRs simultaneously,, then we really need the author to do a double check, and then the author pushes it in.

Best,
- Gene

gc00:
Hi Kapil,
    I've now re-ordered the commits to make it easy to avoid conflicts.  I believe that I've now created all the commits that we had before.

Since I was rewriting the 'master' branch, I decided to take advantage of the rewrite to also rename it to the 'main' branch.  It was time for DMTCP to join the convention for other repos.  :-)

If you still have that ARM64 VM, could you do a sanity check and see if the new DMTCP 'main' branch runs on ARM64?  If there are no serious bugs, then I'll do a DMTCP 3.1.0 release.

Thanks,
- Gene

gc00:
Hi Kapil,
    Sorry for the delay.  I was travelling yesterday.  We're visiting my daughter Sarah in Brooklyn right now.

I did include it, but I changed the git message to expand it, and I expanded some comment in the code.

I've now added a comment to PR #1095 for clarity:
====

Please note that this was pushed in on Aug. 13.  But I first expanded some of the comments.
So, the original commit message was PR #1095 (now PR #1121 ).
> Initiate ckpt on receiving ckpt-signal

It is now:
>  Coord initiates ckpt if user receives ckpt-signal
with commit hash:  835b557181713c226deb78cddac703c70bb2b9ed

Another way to convince yourself is to cherry-pikc the commit of this PR on top of 'main'.  Git will say that it's an empty commit.

PR #1121, opened 2024-08-14
Initiate ckpt on receiving ckpt-signal.

karya0:
This PR makes it possible for a user to use kill -SIGUSR2 <pid> to initiate a checkpoint in addition to the existing methods of using dmtcp_coordinator terminal and dmtcp_command.

This functionality also allows a resource manager such as slurm to request checkpoint by sending a custom signal (e.g., 15) to all processes in the distributed computation. Once a process receives this custom signal (along with --ckpt-signal 15 flag for dmtcp_launch), it will request the coordinator to initiate a ckpt. Only first request will be honored, the rest will be safely ignored by the coordinator.

Recreating #1095


gc00:
@karya0 , This commit was included in 'main', but I changed the commit message.
Please see my comment for PR #1095 for details.

PR #1123, opened 2024-08-16
Fix pthread1 for compiling

gc00:
On CentOS 7 (glibc-2.17),, this would not even compile.  This fixes it.  It's trivial to review.


xuyao0127 (approved):
LGTM

PR #1124, opened 2024-08-17
Fix bugs in clone1

gc00:
There were several bugs in clone1.
A race condition usually triggered on CentOS 7, and the race could be forced on all platforms by adding `sleep(1)` after `sem_post()`.  The race condition happens with native 'clone1`' (without DMTCP).
 * There are now two tests in autotest.py, and they both work in DMTCP:
    [ make check-clone1 && make check-clone2 ]
   + "clone1" (./clone1 0)  # clone1 emulates a non-detached thread
   + "clone2" (./clone1 1)  # clone1 emulates a detached thread 

This also serves as an example of using clone() to emulate a detached or non-detached thread, including how to wait for the child to exit.  With this example, DMTCP could use this model to support vfork()`


gc00:
Let's talk about this offline.  If we're not supporting 'clone', then should we even have a 'clone1' test?

gc00:
@karya0 ,
    You wrote:
>Right now, if someone calls clone and it ends up in our __clone wrapper, we simply assert out.
On the other hand, as we see from the tests, this code calls our 'clone' wrapper, and the check now succeeds.  And the reason why it was failing before is because we were not properly reaping the previous cloned thread before starting a new one.

I can go either way on this PR:
1. We can use the PR and keep the clone1 test here so that others can see how to use clone in a manner that is compatible with DMTCP.  This shows how to create detached threads and regular threads using clone.  A similar strategy would work for vfork and posix_spawn.
2.   Or else we can decide that DMTCP should not be supporting 'clone' at all. and we shouldn't encourage them.

I'm happy with either decision.  What do you prefer?

gc00:
@karya0 ,
    I've now added a commit on top of this that removes `test/clone1.c`, and removes it from `autotest.py`, as agreed.
    When you approve this, I'll push it in.  (Also, please note that this PR also has a minor commit here, for `test/Makefile.in`.)  I was finding that errors in a test file would go off-screen, and I wouldn't see them.  So, I re-arranged the order of commands in the Makefile default target.

PR #1129, opened 2024-08-19
jalloc.cpp: bool_atomic_dwcas(): use mutex if sync_bool not available

gc00:
If `__sync_bool_compare_and_swap()` is not available for a CPU, then fall back to using a mutex for atomicity.  This will make codes slower only if they _very_ frequently use DMTCP-internal operations, such as thread create/destroy for OpenMP.

This is important for newer CPU architectures that do not support an assembly instruction for 128-bit atomic compare and swap.  A current example is the RISC-V CPU.


xuyao0127 (approved):
LGTM

PR #1130, opened 2024-08-27
Mostly lock-free RW lock.

karya0:
Replaces mutex inside RWlock with atomics.

PR #1133, opened 2024-09-01
Revert "Mostly lock-free RW lock."

karya0:
Reverts dmtcp/dmtcp#1130


gc00 (approved):
LGTM.  I'm approving this revert PR.
Thanks for your cooperation.  We do want some of these commits, but let's revert for now, and then go slow as we push things in.  Thanks.

PR #1137, opened 2024-09-09
We cannot trust ARM64 for HAS_ATOMIC

gc00:
* FIXME:  We should use futex_wait, futex_post instead of pthread_mutex_lock/unlock for ARM64 for the reasons below. Add another commit.
 * See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70814 for the problem with /lib/aarch64-linux-gnu/libatomic.so.1 on ARM64 as of Sept., 2024.  (tested on libc-2.36)

@maxwellpirtle, Does this PR now fix the bug you saw with McMini during startup?

@karya0 ,
    This was a weird bug.  jalloc.cpp had HAS_ATOMIC defined, and used the C++ atomics. That caused C++ to delegate to libatomic.so.  But in libatomic.so, they did _not_ use the atomic assembly instruction.  Instead, they called `pthread_mutex_lock`.  Unfortunately, McMini had its own wrapper around  `pthread_mutex_lock`.  And that led to DMTCP getting control again and DMTCP trying to do an atomic in jalloc.cpp.
    So, for ARM64, I'm not allowing it to use HAS_ATOMIC.  The URL above has a full explanation.

### DIFF
--- a/configure
+++ b/configure
@@ -11590,7 +11590,7 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 printf "%s\n" "$gccSyncBuiltins" >&6; }
 if test "$gccSyncBuiltins" = "yes"; then
 
-printf "%s\n" "#define HAS_SYNC_BOOL 1" >>confdefs.h
+printf "%s\n" "#define HAS_128_SYNC_BOOL 1" >>confdefs.h
 
 fi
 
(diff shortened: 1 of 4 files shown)


gc00:
@karya0 ,
    Thanks for approving.  I assume that I can push this PR in, as is. (no further changes needed)
 
You noted the comment:
> We need to create proper DMTCP wrappers/functions for DWCAS, etc., and make them available for DMTCP core as well as plugins.

I completely agree with this.  Would you like to either create a PR, or point me in the right direction and I'll do the PR?

PR #1138, opened 2024-09-09
ARM64: add comment, __NR_getpgrp doesn't exist in glibc-2.36

gc00:
So, __NR_getpgrp is  not defined for ARM64, even though glibc-2.36 does have the getprgrp symbol.

I've tested this now on an ARM64 Raspberry Pi with glibc-2.36.


aayushi363 (approved):

gc00:
The change was added in a different commit.

PR #1142, opened 2024-09-18
Remove unmaintained plugins.

karya0:
These plugins are no longer maintained/supported, so removing them.
* Infiniband
* IB2TCP
* Batch-queue
* Ptrace
* HBICT


gc00 (approved):
LGTM.

PR #1143, opened 2024-09-18
Minor updates related to launch-refactor.

karya0:


gc00 (changes requested):
I'm worried about the first commit: [Use std=gnu11 for C.]
Could you write in the PR description what this is doing?  Are we going to have to start supporting m4 files, even as upstream changes those files?

If you'd like to split this commit into two, then I can approve the second and third commits right now.  And then we can review the first commit at our leisure.

gc00:
@karya0 , I'm still nervous about the first commit.  Can we split this PR, so that we can push in the last two commits, while we discuss the first one?

For the first commit, my concerns are several.  (And maybe we should just talk it over by Zoom.)  Mainly, it takes us one step further from new PhD students understanding the code.  They would have obvious questions:  I thought DMTCP is using C++14 now.  Why is gnu11 also important?  Why are we recommending to use C11 in the diff for this commit, when we also write in configure.ac:
> `AX_CXX_COMPILE_STDCXX([14], [noext], [mandatory])`

Also, why does the configure.ac not explain the need for the m4 files.  And if gnu11 is important, won't a future autoconf also support an option for gnu11.

Maybe we can add comments to answer these questions.  But I'd like to talk this over.  Meanwhile, let's push in the last two commits.

gc00 (approved):
Thanks for moving the first commit to a different PR, where it will be motivated betterr.  And yes, I would enjoy a Zoom call to discuss it.  :-)

Meanwhile, I'm approving the rest, of course.

PR #1146, opened 2024-09-20
Added tid offset for M32 mode.

karya0:
This is required to compile 32-bit binaries on x64 machine.

### DIFF
--- a/src/plugin/pid/glibc_pthread.h
+++ b/src/plugin/pid/glibc_pthread.h
@@ -110,10 +110,8 @@
 struct libc_tcbhead_t {
 #if defined(__x86__) || defined(__x86_64__)
   char pad[704];
-#elif defined(__arm__) || defined(__aarch64__)
+#elif defined(__arm__) || defined(__aarch64__) || defined(__riscv) || defined(__ILP32__)
   char pad[24 * sizeof(void*)];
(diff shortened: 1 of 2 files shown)


gc00 (approved):
I'll approve this in advance.  But I'll add two comments for your consideration:
1.  `192` and `24*sizeof(void *)` are equivalent for 64-bit architectures..  :-)  But I'm fine with using the latter.
2. Your're also adding `__ILP32__`.  I'm not sure of the motivation here.  `24*sizeof(void *)` presumably works for 32- and 64-bit architectures.  What's special here about singling out 32-bits only?  Also, `__x86__` and `__aarch64__` are both special cases in this logic.  Are you sure that there will never be a future architecture that is also a special case?
3. Based on this, I would suggest removing `__ILP32__`.  And I've never tested for the 32-bit version of `__aarch64__` (which restricts to a subset of the 64-bit instructions, and allows access only to the lower 32 bits of registers). But if this `__ILP__` interferes with a future architecture (such as 32-bit ARM64,, we'll quickly know it.  So, I'm not really concerned.

gc00:
@karya0 , Thanks for the explanation for the PR 'Added tid offset for M32 mode.'
That all makes sense.

PR #1156, opened 2024-09-27
Added Coordinator Plugin Interface.

karya0:
This PR introduces Coordinator plugins to handle specialized tasks within the coordinator. The ultimate goal is to carve out orthogonal features/tasks as plugins to simplify the core business logic inside coordinator and make it easier to enhance and validate changes.

To help with the plugin interface, a new flags struct is created to keep all coordinator flags together and make them available as part of the plugin API.

A future enhancement will pull various coordinator status variables in a similar class and make it available to the plugin API as well as in the key-value store to help with debugging.

### DIFF
--- a/test/autotest.py
+++ b/test/autotest.py
@@ -506,12 +506,12 @@ def getStatus():
   for line in output:
       line=str(line.strip().decode("ascii"))
 
-      m = re.search('NUM_PEERS=(\d+)', line)
+      m = re.search('NUM_PEERS=(\\d+)', line)
       if m != None:
         peers = int(m.group(1))
         continue
 
-      m = re.search('RUNNING=(\w+)', line)
+      m = re.search('RUNNING=(\\w+)', line)
       if m != None:
         running = m.group(1)
         break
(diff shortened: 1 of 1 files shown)


gc00:
I'd like to give this PR a "deep review", since it changes DMTCP in important ways for MANA.

@xuyao0127 , Could you download this, and decide if this PR will allow us to replace the "orphan branch" of DMTCP currently used by MANA?  Then, please do report back to us on issues, bugs, or enhancements needed before it can support MANA?

@xuyao0127 , In particular, you've reported that in your Fedora 39 (gcc-15), the "orphan branch" of DMTCP doesn't compile.  We agreed that rather than fix it, we would upgrade to Kapil's newest DMTCP.  Please tell us if this compiles with Fedora 39 (gcc-15).  In particular, in November, we will see Ubuntu 24.10 with gcc-15, and so we must fix this issue before then, or else we will get a lot of bug reports.

Thanks,
- Gene

gc00:
@karya0 , Thanks for the comment.  I'll wait for when it's no longer a WIP.

I'd still like @xuyao0127 to share the reviewing with me, in case I have blind spots.  But we can review this quickly (without MANA) when it's ready.

gc00 (approved):
I approve.

My comment on style is optional.  Different people will always have different preferred styles.

A minor "ask", is if you can write a short paragraph with each PR, saying somethign about what it does, and the code structure.  This should be a guide for the reviewer, so that they can quickly get a mental map of what each piece of code is trying to do.

I think I've figured out this one.  You've created perhaps two classes, and at least one of them is collecting together all the possible flags and parameters in a single class.  This makes the code clearer (and "more local" :-) ), in that one can read all the flags and parameters in one place, before seeing their usage later in the code.

[ Sorry for my delayed review.  Too many things happening.  I'll get to the other PR soon. ]

PR #1161, opened 2024-10-12
Fixed broken interval-ckpt logic in coordinator.

karya0:
This commit got omitted during the creation/merge of PR #1156  which broke interval ckpt. This PR fixes it.


gc00 (approved):
Thanks.  This owrks.
I also tested with
> `dmtcp_coordinator --stale-timeout 20 --daemon`

and 
> `dmtcp_coordinator --timeout 20 --daemon`.

Those also work.  So, I believe that we no longer have any regressions.

LGTM!

I'll create a new PR for DMTCP-3.1.2.  Please review it when you push this in.

Then, after you push in this PR, I'll push in the PR for a new release.

Issue #1163, opened 2024-10-14
Bug in pthread2 (seen on RISC-V, glibc-2.37): detached threads


gc00:
On restart, usually on thread 4, we see:

> (gdb) where
#0  0x0000003ff7bd2c2e in list_del (elem=0x3fef8bc220) at ../include/list.h:55
#1  __GI___nptl_stack_list_del (elem=0x3fef8bc220) at nptl-stack.c:32
#2  __GI___nptl_deallocate_stack (pd=pd@entry=0x3fef8bc160) at nptl-stack.c:116
#3  0x0000003ff7bd2e1a in __GI___nptl_free_tcb (pd=0x3fef8bc160) at nptl_free_tcb.c:42
#4  0x0000003ff7bd5422 in start_thread (arg=<optimized out>) at pthread_create.c:575
#5  0x0000003ff7c2bfca in __thread_start () at ../sysdeps/unix/sysv/linux/riscv/clone.S:85  

glibc wants to free the stack and tcb of a detached head, before starting the new thread start function (threadMain).
In doing this, the pd (type: struct pthread) has a list field (type: list_t)..

Most of the 'pd' for threads have addresses like ` (struct pthread *) 0x3fef8bc160`..  But for thread 4:
> (gdb) p pd->list
$7 = {next = 0x3f00009c6c, prev = 0x3fef9bd220} 

and `0x3f00009c6c` is not mapped.

Most other threads have a stack:
> (gdb) where
#0  futex_wait (private=0, expected=2, futex_word=0x140d0 <dl_main+8766>) at ../sysdeps/nptl/futex-internal.h:146
#1  __GI___lll_lock_wait (futex=futex@entry=0x140d0 <dl_main+8766>, private=0) at lowlevellock.c:49
#2  0x0000003ff7bd7fe6 in lll_mutex_lock_optimized (mutex=0x140d0 <dl_main+8766>) at pthread_mutex_lock.c:48
#3  ___pthread_mutex_lock (mutex=0x140d0 <dl_main+8766>) at pthread_mutex_lock.c:93
#4  0x000000000001135a in process_dl_debug (state=<optimized out>, dl_debug=<optimized out>) at rtld.c:2537
Backtrace stopped: frame did not save the PC 

Thread 1 is still in rtld:
> (gdb) where
#0  0x0000003ff7c01094 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,          
[...]

PR #1136, opened 2024-09-08
Riscv dmtcp experiment

aayushi363:
This branch is the same as the previous RISC-V branch, but I have split one large commit into several smaller ones.

PR #1165, opened 2024-11-05
Fixes for pthread handling on RISCV.

karya0:
The RISCV pthread trouble described in #1163 had two main issues:
1. Race condition related to a thread performing exit during SUSPEND message processing in the ckpt-thread. The first commit addresses this issue by polling active threads twice to ensure we wait for a user thread to complete pthread_exit before going forward with checkpoint.

2. Improper TID offset calculation done in #1136 which somehow both @gc00 and I missed during our review of the PR. The offset  was mistakenly written as 192 when it should have been `sizeof (void*) * (24 + 2)` (i.e., 208). 
[ FROM @gc00: PR #1140 should have allowed `./configure` to catch this error.  But `util/check-pthread-tid-offset` already had the correct 208 value, while the source had 192.  The two codes were not synchronized.  :-(  ]

All pthread tests now pass on RISCV machine provided by @gc00.

### DIFF
--- a/src/tls.cpp
+++ b/src/tls.cpp
@@ -172,11 +172,6 @@ TLSInfo_GetTidOffset(void)
 
 //FIXME:Define offset for __aarch64__
 
-#ifdef __riscv
-  offset = 192;
-  return offset;
-#endif
-
   if (glibcMinorVersion() >= 10) {
     offset = 26 * sizeof(void *);  // sizeof(__padding) + sizeof(list_t)
   } else {
@@ -364,13 +359,13 @@ tls_get_thread_area(Thread *thread)
 {
   unsigned long int addr;
   asm volatile ("addi %0, tp, 0" : "=r" (addr));
-  thread->tlsInfo.tlsAddr = addr - 1856;  // sizeof(struct pthread)=1856
+  thread->tlsInfo.tlsAddr = addr;
 }
 
 static void
 tls_set_thread_area(Thread *thread)
 {
-  unsigned long int addr = thread->tlsInfo.tlsAddr + 1856;
+  unsigned long int addr = thread->tlsInfo.tlsAddr;
   asm volatile("addi tp, %[gs], 0" : : [gs] "r" (addr));
 }
 #endif  /* end __riscv */


aayushi363 (approved):

gc00 (commented):
@karya0 , Thanks for pinging me.  I'm coming back to this now.  (Busy week.)  And thank you for providing these solutions to a subtle problem!

This is all looking good.  I'd just like a little more time to inspect it carefully.  And after that, I'll most likely approve it.

QUESTION:
    In the discussion, you said that there was a bug in which we set the thread descriptor offset to 192, when it should have been 208.  I agree that it's 208.
    But after the original commit for RISC-V (PR #1136), I had added a commit (PR #1140) for configure.ac, that was supposed to verify that the tid offset was correct.  And you can run it manually by doing:
> `./check-pthread-tid-offset  && echo "IS CORRECT"`

or else by doing:
> `./check-pthread-tid-offset -v`

QUESTION:  Why didn't `./configure' discover that we had the wrong tid offset?  I'd like to know, so as to simplify future debugging.

Thanks.

src/threadlist.cpp
```diff
@@ -534,6 +543,31 @@ ThreadList::suspendThreads()
...
+          // Thread exited. Let's remove it from the list.
+          ThreadList::threadIsDead(thread);
+        } else {
+          needsRescan = 1;
+        }
+      }
+    }
+
+    if (needsRescan) {
+      usleep(1000);
```
gc00, line 566: Minor, but POSIX declared usleep to be obsolete.  The POSIX way to do it is:
```
#include <time.h>
  struct timespec req = {.tv_sec = 0; .tv_nsec = 1000000;};
  nanosleep(&req, NULL();
```

src/dmtcpworker.cpp
```diff
@@ -390,7 +390,7 @@ DmtcpWorker::waitForCheckpointRequest()
 
 // now user threads are stopped
```
gc00, line 391: Most of my comments below can be ignored, if you agree to do the following refactoring.  I think it's simpler.

You used to have::
```
threadlist.cpp:checkpointhread():
  DmtcpWorker::preCheckpoint(); 
```

Right now, you have split `preCheckpoint` into `releaseLocks` and `preCheckpoint`:
```
threadlist.cpp:checkpointhread():
  DmtcpWorker::releaseLocks();
  ThreadList::waitForExitingThreads();
  DmtcpWorker::preCheckpoint(); 
```

But in this case, why not keep the old `preCheckpoint`, and simply insert waitForExitingThreads()` into the right location inside `preCheckpoint`:

```
  JTRACE("Threads suspended");
  WorkerState::setCurrentState(WorkerState::SUSPENDED);

  ThreadSync::releaseLocks();

  if (exitInProgress) {
    // There is no reason to continue checkpointing this process as it would
    // simply die right after resume/restore.
    // Release user threads from ckpt signal handler.
    ThreadList::resumeThreads();
    ckptThreadPerformExit();
  }

  ThreadList::waitForExitingThreads();

  ...
```

gc00 (approved):
I'm approving the PR in advance, but only on the condition that you agree that the proposed refactoring of the code is simpler.  :-)

If the proposed refactoring is not possible (maybe I missed an important point), then could we discuss it again before pushing this in?

And once this PR is pushed in, I will do another "dot release" to incorporate this.  Thanks very much!  Nice analysis.

src/dmtcpworker.cpp
gc00: Sorry for my delay.  I'll look at this carefully on Wednesday.
Probably the right thing, then, is to add a comment saying that long-term, we want ThreadList to call DmtcpWorker, and never vice versa.  I can look at this closely on Wednesday, and suggest a comment to document this (and maybe a strategy to prevent future entanglement in case someone again tries to call ThreadList from DmtcpWorker).

If you have a way of making future PRs to have more difficulty in calling TThreadList from DmtcpWorker, that's interesting too.

gc00:
@karya0 , Sorry that it took me so long to get back to this.
I understand your issue that either we have threadlist.cpp as the higher layer, and dmtcpworker.cpp as the lower layer; or vice versa.

I'd like to refactor the code, and then apply the _simple_ version of your fix that I propose at the beginning of this conversation.

What I'm seeing is that we could move two methods FROM threadlist.cpp TO dmtcpworker.cpp.  After that, we can apply the siple version of your fix.

I'll describe my proposal here.  If you like this proposal, let me kow, and I'll create a draft code for it.

```
dmtcpworker.cpp above threadlist.cpp
  MOVE ThreadList::checkpointhread() to the DmtcpWorker class
  MOVE ThreadList::waitForAllRestored() to the DmtcpWorker class
```

Here are the current thread dependencies.  Notice how this untangles it and allows your ThreadList::waitForExitingThreads() to be integrated in the simple manner that I outline.

```
Thread *ThreadList::ckptThread = NULL;
ThreadList::checkpointhread()->DmtcpWorker::waitForCheckpointRequest()/preCheckpoint/postCheckpoint
ThreadList::checkpointhread()->ThreadList::writeCkpt()->prepareMtcpHeader()->ThreadList::postRestart()
ThreadList::checkpointhread()->ThreadList::waitForAllRestored()->DmtcpWorker::postRestart() [never calls ThreadList]
ThreadList::postRestart()->ThreadList::postRestartWork->SharedData::postRestart()    
```

With moving the two methods above,  your `  ThreadList::waitForExitingThreads()` can be easily inserted in `DmtcpWorker::preCheckpoint()`

**Does this work for you?**
If so, I'll go ahead and submit a proposed commit.

PR #1172, opened 2024-12-28
Added support for pthread_cancel handling.

### LINKED ISSUE #1171: pthread_cancel not working
In pthread_cancel it will get the tid from pthread_t structure, but the tid field has been changed to the virtual tid in thread_start=>procssChildThread=>dmtcp_init_virtual_tid, then tgkill syscall will fail in such case.

image

Please note it still works in dtmcp(2.6.0) version, as you can see the picture above with strace result, "p.1" is the working case while "p.2" is not working.

karya0:
Libc uses pthread_kill internally to signal to-be-canceled thread. This call fails because pthread_kill uses virtual tid stored in the pthread descriptor. We get around this issue by making an explicit call to pthread_kill _after_ libc has processed the rest of the pthread_cancel call. This is based on the observation (i.e. libc:pthread_cancel implementation) that pthread_kill is the last step in pthread_cancel implementation.

While fixing this issue, found two more long-standing corner cases in virtual pid table implementation that are fixed here as well. A third issue in virtual id implementation is left as is, albeit with a todo.

Added a new test case which fails without this fix and passes with this fix.

Fixes #1171

### DIFF
--- a/test/autotest.py
+++ b/test/autotest.py
@@ -69,6 +69,13 @@ disabled_tests = [
   #"vfork1", # We simulate vfork using fork.
   # vfork1 and vfork2 tests were failing due to a previous bug in vfork.c
   #"vfork2",
+
+  # The emacs test currently fails on github during restart due to huge ckpt
+  # image sizes (~500MB). A proper fix would be to diagnose the cause for the
+  # unexpectedly large image size. If the large sizes are inevitable, we can
+  # edit this script to ignore failures on ckpt images that are >~100MB.
+  "emacs",
+
   # This test needs to be fixed.  It is not really running.
   # ERROR: ld.so: object '/home/gene/dmtcp.git/test/plugin/example-db/dmtcp_example-dbhijack.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
   "dmtcp_example-db"


test/pthread6.c
```diff
@@ -0,0 +1,70 @@
...
+void *
+start_routine(void *arg)
+{
+  int threadId = *(int*)arg;
+  free(arg);
+  printf("ThreadId: %d alive\n", threadId);
+
+  sem_post(&threadInitialized);
+
+  // Will get canceled in here.
```
gc00, line 21: Suggest adding to the above comment:
```
// Will get canceled in here.  'sleep()' is a cancellation point, checked at end of sleep.
```

test/pthread6.c
```diff
@@ -0,0 +1,70 @@
...
+    int res = pthread_create(&thread, &attr, start_routine, arg);
+    if (res != 0) {
+      fprintf(stderr, "error creating thread: %s\n", strerror(res));
+      return -1;
+    }
+
+    // Wait for thread to finish initialization.
+    sem_wait(&threadInitialized);
+
+    res = pthread_cancel(thread);
```
gc00, line 53: ADD COMMENT:
```
// We could do a sem_post() here, and then do a sem_wait() before 'sleep(10)'
// in the other thread, in order to guarantee that pthread_cancel()
// executes before sleep(10) completes.  But that would defeat
// the purpose of this test.  We want to execute 'pthread_cancel()' before
// checkpoint and then to have 'sleep(10)' finish executing after restart. 
```

gc00 (approved):
Two suggested comments to add for clarity.
But I'm approving this PR in advance.

PR #1174, opened 2025-01-07
Bug fixes for mana

xuyao0127:
This is a WIP PR that includes partial fixes that MANA requires. This will be cleaned later. Only essential changes will be left before merging into the main branch.


gc00:
@karya0 , Is this PR still current?  If so, could you rebase for the current DMTCP?

PR #1175, opened 2025-01-09
Handle dlopen(NULL, ...) calls.

karya0:
dlopen(NULL) returns a handle to the main program.

### DIFF
--- a/src/plugin/dl/dlwrappers.cpp
+++ b/src/plugin/dl/dlwrappers.cpp
@@ -136,6 +136,7 @@ dlopen_try_paths(const char *filename, int flag, string path)
 extern "C"
 void *dlopen(const char *filename, int flag)
 {
+  // TODO(kapil): Replace with scoped lock once we have it.
   bool lockAcquired = dmtcp_libdlLockLock();
   string rpath;
   string runpath;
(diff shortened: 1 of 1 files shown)


src/plugin/dl/dlwrappers.cpp
```diff
@@ -136,15 +136,22 @@ dlopen_try_paths(const char *filename, int flag, string path)
...
+  // 0. If filename is null, call libc dlopen. The user just wants a handle to
+  // the main program.
+  if (filename == NULL) {
+    ret = _real_dlopen(filename, flag);
+  }
+
   // 1. Attempt dlopen using RPATH.
-  if (!rpath.empty()) {
+  if (ret == NULL && !rpath.empty()) {
     ret = _real_dlopen(filename, flag);
```
gc00, line 155: I suggest:
> `if filename == NULL || !rpath.empty()) {`

on the above line, and then remove the changes at 149 to 152.

In summary, the new code should look like the following.  This avoids duplicating the logic of the call to `_real_dlopen()`:
```
// 1a. If filename is NULL, call libc dlopen. The user just wants a handle to // the main program.
// 1b. If RPATH exists, call libc dlopen using RPATH.
if filename == NULL || !rpath.empty()) {`
  ret = _real_dlopen(filename, flag); }
```

gc00 (changes requested):
I think my proposed revised logic is simpler.  Am I missing something?

src/plugin/dl/dlwrappers.cpp
```diff
@@ -136,15 +136,22 @@ dlopen_try_paths(const char *filename, int flag, string path)
...
   string runpath;
   void *ret = NULL;
 
   getRpathRunPath(__builtin_return_address(0), &rpath, &runpath);
 
-  // 1. Attempt dlopen using RPATH.
-  if (!rpath.empty()) {
+  if (filename == NULL) {
+    ret = _real_dlopen(filename, flag);
+  }
```
gc00, line 149: Sorry to be picky, but isn't the above code redundant?  If `filename == NULL`, then the code below will already do:
> `ret = _real_dlopen(filename, flag);`

gc00 (changes requested):
I think we still have a duplicate code.  Can we fix this?

gc00 (approved):
LGTM.  Thanks.

PR #1176, opened 2025-01-09
Added configure flag to specify fixed addr for mtcp_restart text.

karya0:
The linker flags for specifying text address are required for MANA to force mtcp_restart into a "safe" address with respect to the kernel-loader, upper half, and lower half regions. For the mainstream DMTCP applications, this fixed address approach may disrupt how heap is handled post restart. In some cases, the heap assigned by the kernel is at a lower address than the this mtcp_restart fixed address, causing sbrk calls to fails when it attempts to shrink heap.


gc00 (changes requested):
I tjhink this commit modifies 11 files because it's doing two things at once.  I eventually gave up trying to trace your changes, and the changes for autoconf 2.72.

Please split this into two commits.  One alternative is:
> COMMIT A:  update from autoconf 2.71 to 2.72  (change many files)
> COMMIT B:  Then apply your changes, so that we can see what the commit is doing (change just a few files, related to the bug fix)

Or if you prefer, an alternative is just to stay with autoconf 2.71, and figure out what changes to configure are being applied by your new configure.ac.  Then modify configure _only_ with the changes based on your changed configure.ac.

configure.ac
```diff
@@ -389,6 +389,17 @@ if test "$use_dlsym_wrapper" = "yes"; then
   AC_DEFINE([ENABLE_DLSYM_WRAPPER],[1],[Enable dlsym wrapper in libdmtcp.so])
 fi
 
+AC_ARG_ENABLE([mtcp_fixed_textaddr],
+            [AS_HELP_STRING([--enable-mtcp-fixed-textaddr],
+                            [Loads mtcp_restart at a fixed text addr, as
+                            required by MANA.])],
```
gc00, line 395: I also like your description in the PR of the reason for this:
> In some cases, the heap assigned by the kernel is at a lower address than this mtcp_restart fixed address, causing sbrk calls to fail when they attempt to shrink heap.

Could we change the above lines to:
```
            [AS_HELP_STRING([--enable-mtcp-fixed-textaddr],
                            [Loads mtcp_restart at a fixed text addr, as
                            required by MANA.  WARNING:   In some cases,
                            the heap assigned by the kernel is at a lower
                            address than this mtcp_restart fixed address,
                           causing sbrk calls to fail when they attempt to
                           shrink heap.])],
```

gc00 (approved):
@karya0 , I'm approving this in advance, but please enlarge the comment, as requested.

@xuyao0127 , Please note that when this is pushed in, MANA will then require configuring DMTCP with `--enable-mtcp-fixed-textaddr`.  Please add this flag to the MANA configure now, as a comment, so that we don't lose this important information.

configure.ac
xuyao0127: @gc00 I can add this to MANA when configuring DMTCP. But MANA no longer needs mtcp_restart for restart. I think this flag will make no difference in MANA.

PR #1180, opened 2025-02-13
Set WorkerState to RESTARTING in src/dmtcprestartinternal.cpp

xuyao0127:
The src/dmtcprestartinternal.cpp was a duplicate of mtcp_restart.cpp
  for MANA, but it was not updated. This change only affects MANA.


gc00 (approved):
LGTM

PR #1184, opened 2025-02-22
Added support for vvar_vclock ckpt.

karya0:
On new kernels, there is a new area `[vvar_vclock]` that needs to be handled in the same way as `[vvar]`. This PR duplicates the `[vvar]` handling to support `[vvar_vclock]`. We can refactor the code to remove duplication down the road.


xuyao0127 (approved):
LGTM

PR #1188, opened 2025-02-27
Fixed readdmtcp.sh and minor cleanup for restore buf handling.

karya0:
PR #1144 broke readdmtcp.sh working which is now fixed in this PR.

Also cleaned up a bit of logic around restore buffer handling.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -84,7 +84,6 @@ static int read_one_memory_area(int fd, VA endOfStack);
 static void restorememoryareas(RestoreInfo *rinfo_ptr);
 static void restore_brk(RestoreInfo *rinfo);
 static int doAreasOverlap(Area *area, MemRegion *memRegion);
-static int hasOverlappingMapping(VA addr, size_t size);
 static int mremap_move(void *dest, void *src, size_t size);
 static void remapExistingAreasToReservedArea(RestoreInfo *rinfo, void (*restore_func)(RestoreInfo*));
 static void remapMtcpRestartToReservedArea(RestoreInfo *rinfo,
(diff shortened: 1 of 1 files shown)


src/writeckpt.cpp
```diff
@@ -171,16 +171,10 @@ mtcp_writememoryareas(int fd)
    *        restart() is heavily overloaded, but assuming it's only relevant
    *        to processinfo.cpp, we have it used in: ProcessInfo::restart()
    *        and 'case DMTCP_EVENT_RESTART' in processInfo_EventHook().
-   *
```
gc00, line 174: And this is the related comment.  See my above inline comment.

Conceptually, we have non-local code.  Data structures must be maintained consistently between writeckpt.cpp and mtcp_restart.c  So, we need to document the consistency requirements in both places.  Otherwise, new developers have a hard time to understand the code.

src/writeckpt.cpp
```diff
@@ -190,14 +184,7 @@ mtcp_writememoryareas(int fd)
     // will invoke mmap if the JAlloc arena is full. Similarly, for STL objects
     // such as vector and string.
```
gc00, line 186: And I have a similar worry about removing this comment, and not documenting our consistency/design requirements between writeckpt.cpp and mtcp_restart.c.

gc00 (commented):
I'm worried aboutt the second commit.  It says `Minor cleanup for restoreBuf handling.`.  But it seems liek we're removing various comments that would help other developers.  Usually, I don't anticipate that a "Minor cleanup" should remove documentation, instead of updaating it to the most recent code.  :-)  (joke)

Could you update the comments?  If we're no longer worried about overlap leaving us no place for a restorebuf hole on restart, ccould you document why this will never be a problem?  Even better, can you also add a JASSERT to check for any implicit assumptions that you're making (along with comments about why)?  Thanks.

gc00 (approved):
LGTM.

Thanks for pointing out that the comment about three times the restoreBuf is still there -- just in mtcp_restart.c now.  That works.  And as you point out, the writeckpt.cpp code initially just creates an arbitrary buffer size, and mtcp_restart.c is the one that views it as three separate buffers, in case of overlap.

Thanks.

PR #1191, opened 2025-03-01
Fix shmget wrapper to preserve original key values on first use.

gkliu:
When wrapping the shmget function, this PR fixes the handling of non-IPC_PRIVATE keys. 

Previously, when VIRTUAL_TO_REAL_SHM_KEY returns -1 (indicating the key is being used for the first time by user code), we need to preserve the original key value rather than attempting to translate it. This ensures that explicit key values specified by the application are maintained correctly.

This fix is important for applications that rely on specific key values for their shared memory regions, as it ensures the DMTCP wrapper doesn't modify these values on first use.


gkliu:
Many thanks, @karya0. I tested my use cases and they are working well. Really appreciated it!

PR #1192, opened 2025-03-02
Fixed IPC_PRIVATE handling for SysV Shm.

karya0:
IPC_PRIVATE can [now] be called multiple times and it will in turn create new handlers every time.

Also updated sysv-shm1 test to create shm segments with IPC_PRIVATE key.

Thanks @gkliu for reporting the bug and creating an initial PR (#1191) with the fix.

### DIFF
--- a/include/shareddata.h
+++ b/include/shareddata.h
@@ -198,7 +198,8 @@ void updateDlsymOffset(int32_t dlsymOffset, int32_t dlsymOffset_m32 = 0);
 int32_t getDlsymOffset(void);
 int32_t getDlsymOffset_m32(void);
 
-int32_t getRealIPCId(int type, int32_t virt);
+int32_t getRealIPCId(int type, int32_t virt,
+                     bool insertIfNotFound = false);
 void setIPCIdMap(int type, int32_t virt, int32_t real);
 
 pid_t getRealPid(pid_t virt);
--- a/src/plugin/svipc/sysvipc.cpp
+++ b/src/plugin/svipc/sysvipc.cpp
@@ -470,8 +470,8 @@ SysVShm::virtualToRealKey(key_t k)
   if (_keyMap.find(k) != _keyMap.end()) {
     return _keyMap[k];
   } else {
-    int realId = SharedData::getRealIPCId(SYSV_SHM_KEY, k);
-    if (realId != -1) {
+    int realId = SharedData::getRealIPCId(SYSV_SHM_KEY, k, true);
+    if (realId == k) {
       updateKeyMapping(k, realId);
     }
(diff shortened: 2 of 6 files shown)


include/shareddata.h
```diff
@@ -198,7 +198,8 @@ void updateDlsymOffset(int32_t dlsymOffset, int32_t dlsymOffset_m32 = 0);
 int32_t getDlsymOffset(void);
 int32_t getDlsymOffset_m32(void);
 
-int32_t getRealIPCId(int type, int32_t virt);
+int32_t getRealIPCId(int type, int32_t virt,
+                     bool insertIfNotFound = true);
```
gc00, line 202: If I'm reading this right, the default value for 1insertIfNotFound` is 'true', and there is one invocation in which we make the argument explicit, inside `SysVShm::virtualToRealKey(key_t k)`.  But that invocation also specifies 'true' for this optional argument.

Is there any case where the optional argument would be false?  If not, then let's just hardwire the code to always insert, if not found. 

I also prefer removing the optional argument, because we eliminiate one of the non-local issues in the design of C++.  If we remove the optional arg, the reader no longer has to look at the .h file to determine if there is a missing default argument, and what is the default value.

gc00 (changes requested):
@karya0  (and  @gkliu), Thanks for this PR.  It's an important fix.

@karya0 , I have one suggestion:  Since the optional argument is always 'true', why not remove the optional argument?  But if you have a reason for keeping the optional arg, let me know, and I'll still approve it.

include/shareddata.h
karya0: That's a good catch. It is supposed to be `false` to ensure we don't break any older scenarios with plugins like MANA. I have fixed it now.

gc00:
I'm not sure if I'm the one who should be replying after an update.  Just in case, this is still the only concern that I have:
> @karya0 , I have one suggestion: Since the optional argument in SharedData::getRealIPCId() is always 'true', why not remove the optional argument? But if you have a reason for keeping the optional arg, let me know, and I'll still approve it.

karya0:
> I'm not sure if I'm the one who should be replying after an update.  Just in case, this is still the only concern that I have:
> 
> > @karya0 , I have one suggestion: Since the optional argument in SharedData::getRealIPCId() is always 'true', why not remove the optional argument? But if you have a reason for keeping the optional arg, let me know, and I'll still approve it.

@gc00 : I already fixed a typo and added a response to your comment about the optional argument. Please take a look and approve if it looks good.

gc00 (approved):
Thanks.  I have no mroe comments.
LGTM.

PR #1193, opened 2025-03-02
Use static buffer for motherofall.

karya0:
When using JALLOC_MALLOC for motherofall allocation, we can get into infinite recursion if a user application has put wrappers around mmap. The fix is to use a static buffer.


gc00 (approved):
LGTM.

Nice!  This might explain the few instances where I've seen an infinite recursion before 'main()'.

PR #1195, opened 2025-03-06
Handle applications with user-defined mmap wrappers.

karya0:
This PR addresses several wrapper initialization issues exposed on 64-bit ARM machines as observed by a user. Overall, the initialization logic is made leaner with code removal related to static initialization, etc. In addition, a couple of new tests, mmap1 and selinux1 are introduced that will help catch future regressions. The commits are all self-contained and don't have dependencies. However, they are bundled together here because they are all addressing the same general theme and all were required to unblock the user.


jalib/jfilesystem.cpp
```diff
@@ -151,6 +151,46 @@ jalib::Filesystem::DirName(const dmtcp::string &str)
...
+{
+  size_t len = strlen(fname);
+
+  if (strcmp(fname, "/") == 0 || strcmp(fname, ".") == 0 || fname[0] == '\0') {
+    strcpy(dirname, fname);
+    return;
+  }
+
+  if (strcmp(fname, "..") == 0) {
+    strcpy(dirname, ".");
```
gc00, line 164: Is this a bug?  If fname is '..', shouldn't the dirname also be '..'?

jalib/jfilesystem.cpp
```diff
@@ -151,6 +151,46 @@ jalib::Filesystem::DirName(const dmtcp::string &str)
...
+    return;
+  }
+
+  // Remove trailing slashes
+  while (len > 0 && fname[len - 1] == '/') {
+    len--;
+  }
+
+  size_t lastSlash = 0;
```
gc00, line 173: I suggest: `int lastSlash = -1;`  It means that you will also change to:
> `for (int i = 0; i < len; i++) {`
below.  The constant 0 could be confused by the reader with index 0, even though you test for `fname == '/' `above.

src/plugin/dl/dlwrappers.cpp
```diff
@@ -71,29 +72,41 @@ getRpathRunPath(void *caller, string *rpathStr, string *runpathStr)
...
+    Util::replace(rpathStr, &strtab[rpath->d_un.d_val], "$ORIGIN", dirname);
   } else {
-    rpathStr->clear();
+    rpathStr[0] = '\0';
   }
 
   if (runpath != NULL) {
-    *runpathStr = strtab + runpath->d_un.d_val;
-    *runpathStr = dmtcp::Util::replace(*runpathStr, "$ORIGIN", dirname);
+    Util::replace(runpathStr, &strtab[runpath->d_un.d_val], "$ORIGIN", dirname);
```
gc00: 'man ld.so' says that `${ORIGIN}` is also valid.  We should also test for that.

src/plugin/dl/dlwrappers.cpp
```diff
@@ -71,29 +72,41 @@ getRpathRunPath(void *caller, string *rpathStr, string *runpathStr)
   ASSERT_NOT_NULL(strtab);
 
   if (rpath != NULL) {
-    *rpathStr = strtab + rpath->d_un.d_val;
-    *rpathStr = dmtcp::Util::replace(*rpathStr, "$ORIGIN", dirname);
+    Util::replace(rpathStr, &strtab[rpath->d_un.d_val], "$ORIGIN", dirname);
```
gc00: 'man ld.so' says that `${ORIGIN}` is also valid.  We should also test for that.

Also, we're not supporting `$LIB`, `${LIB}`, `$PLATFORM`, or `${PLATFORM`.
We should at least provide a FIXME comment here.
Ooptionally, we could add code to test for this and issue a warning/error.

gc00 (approved):
In the "Conversation" tab for this PR, could you write a little bit about the motivation?  I'm guessing that when an end user adds a wrapper for 'mmap', there can be an infinite recursion.  Is that the issue?

I've also noted inline what may be a bug concerning converting `..` to `.`.

I assume that the 'fixup! Dl-wrapper memory handling fixes.' commit will be rebased into the prior commit before pushing this in.

Anyway, I really like this commit.  I can see where it also fixes other corner cases, that were needed to support 'mmap'.

I'm approving in advance.

test/mmap1.c
```diff
@@ -0,0 +1,45 @@
+// _DEFAULT_SOURCE for mkstemp  (WHY?)
+#define _DEFAULT_SOURCE
```
gc00, line 2: You also need here:
> `#define _GNU_SOURCE  /* Needed for RTLD_NEXT */`
At least, that's what I'm seeing for WSL Ubuntu 22.04.

gc00:
And there also seems to be some issue with the build.  I'm not sure what it is, but it would be nice to fix this before pushing it in.

Locally, this works for WSL Ubuntu 22.04.

If you're not able to figure out why the build is failing, let's talk together.  I'm trying to instrument autotest.py better.

karya0:
Thanks for the comments, @gc00. I am still working on this PR in a rush, so will update the conversation tab once I am done with testing. In a nutshell, the issue is with a user who has custom mmap wrappers in a selinux environment on arm64 that breaks with dmtcp.

> In the "Conversation" tab for this PR, could you write a little bit about the motivation? I'm guessing that when an end user adds a wrapper for 'mmap', there can be an infinite recursion. Is that the issue?
> 
> I've also noted inline what may be a bug concerning converting `..` to `.`.
> 
> I assume that the 'fixup! Dl-wrapper memory handling fixes.' commit will be rebased into the prior commit before pushing this in.
> 
> Anyway, I really like this commit. I can see where it also fixes other corner cases, that were needed to support 'mmap'.
> 
> I'm approving in advance.

jalib/jfilesystem.cpp
karya0: For '..' it should return '.' according to the manpage. But there was a separate bug and I also followed your suggestion about lastslash.

src/plugin/dl/dlwrappers.cpp
karya0: Addressed `${ORIGIN}` and added asserts for $LIB, etc.

test/mmap1.c
karya0: Fixed.

PR #1197, opened 2025-03-17
Fixed a bug in Util::mmap_fixed_noreplace

xuyao0127:
Util::mmap_fixed_noreplace emulates the mmap() with MAP_FIXED_NOREPLACE
  flag for old systems. It uses a regular mmap() with the given address
  and check if the return value is the same as the given address. However,
  we found a case that the kernel returns a different address although the
  requirest memory range is available. Therefore, the mmap_fixed_noreplace
  function returns an error message EEXIST, which is not expectec.

  Now we check if the requested memory range is available by calling mlock()
  page by page. The mlock function will return -1 with an errono of ENOMEM.
  If there's no already mapped pages in the request memory range, the function
  will call mmap() with MAP_FIXED. Otherwise, the function will return
  MAP_FAILED with an errno of EEXIST.

### DIFF
--- a/src/util_misc.cpp
+++ b/src/util_misc.cpp
@@ -729,30 +729,51 @@ Util::replace(const string &in, const string &match, const string &replace)
 //         check that in the function, and either issue an error in that
 //         case, or else simulate the action of MAP_FIXED_NOREPLACE.
 void*
-Util::mmap_fixed_noreplace(void *addr, size_t len, int prot, int flags,
-                           int fd, off_t offset)
+Util::mmap_fixed_noreplace_private(void *addr, size_t len, int prot, int flags,
+                                   int fd, off_t offset)
 {
   if (flags & MAP_FIXED) {
     flags ^= MAP_FIXED;
   }
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 17, 0)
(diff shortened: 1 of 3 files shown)


src/util_misc.cpp
```diff
@@ -729,30 +729,51 @@ Util::replace(const string &in, const string &match, const string &replace)
...
-#define MAP_FIXED_NOREPLACE 0x100000
+# ifndef MAP_FIXED_NOREPLACE
+#   define MAP_FIXED_NOREPLACE 0x100000
+# endif
+  return mmap(addr, len, prot, flags | MAP_FIXED_NOREPLACE, fd, offset);
 #endif
-  // This flag should force: 'addr == addr2' or 'addr2 == MAP_FAILED'
-  flags |= MAP_FIXED_NOREPLACE;
-#endif
-  void *addr2 = mmap(addr, len, prot, flags, fd, offset);
```
karya0, line 745: Since this kernel issue happens rarely, can we try the previous route of using vanilla mmap and see if the kernel returns the correct address. If it doesn't return the correct address, we can go ahead and try the slow route of using mlocks on each page to see what's going on.

xuyao0127: I will add this in another PR. This function is only used when creating the sharedDatatHeader for now.

PR #1201, opened 2025-04-21
Logging improvements

karya0:
* DMTCP_LOG_FILE env var to specify a path for log file. This allows multiple processes to use the same log file and is important when debugging distributed computations.
* Single-line logging -- when logging to single file, one line per log message makes it easier to parse log files using external tooling.
* Bug fixes to make --enable-logging work.
* Some non-fatal asserts replaced with warnings.


xuyao0127 (approved):
LGTM

PR #1205, opened 2025-05-01
Coord: Fixed epoll_wait corner case.

karya0:
When processing epoll_wait events, EPOLLIN should be processed before EPOLLHUP. This ensures no messages are dropped if a client disconnects immediately after sending a message.


xuyao0127 (approved):
I'm approving to unblock your progress. I'm not familiar with this part of DMTCP. Maybe Gene can provide more feedback.

karya0:
Thanks @xuyao0127. I'll merge it now. If @gc00 has any comments, I'll create a follow-up PR to address them.

PR #1214, opened 2025-07-30
Looping for futex until interrupt in ReaderWriterLock

aayushi363:
Changed the DmtcpRWLockRdLock and DmtcpRWLockWrLock such that it retries the futex_wait if the call is interrupted by EINTR or EAGAIN signal.

But this leads to deadlock as soon as we try to take a checkpoint. A deadlock occurs when the DMTCP checkpoint thread tries to acquire an exclusive write lock on the _wrapperExecutionLock while an application thread already holds a read lock on it.  This happens only if the application is multithreaded. 

Currently make check is failing on all multithreading codes in test like pthread1, pthread2, pthread3 and so on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of locking operations by ensuring that interruptions do not cause premature exits from wait states.

### DIFF
--- a/src/rwlock.cpp
+++ b/src/rwlock.cpp
@@ -24,6 +24,7 @@ int DmtcpRWLockRdLock(DmtcpRWLock *rwlock)
 
   DmtcpRWLockStatus oldStatus;
   DmtcpRWLockStatus newStatus;
+  int ret;
 
   __atomic_load(&rwlock->status, &oldStatus, __ATOMIC_RELAXED);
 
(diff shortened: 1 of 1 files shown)


karya0:
> @karya0 , I talked to @aayushi363 , and she told me that you suggested for this PR to use a loop for EINTR, but not to loop for EAGAIN. Could you write here why EAGAIN should not try again in a loop? Thanks.

@gc00: Futex_wait will return EAGAIN if the current futex value no longer matches our cached value. This is expected in a RW lock scenario, where a reader has just incremented nReadersQueued and is now waiting for the writer to call futex_wake. If in between these two events, if the writer already released the lock, the readerFutex will be set to a different value. This would cause the futex_wait call to return with EAGAIN. The reader is then considered to have acquired the read lock and can continue.

src/rwlock.cpp
```diff
@@ -79,6 +83,7 @@ int DmtcpRWLockWrLock(DmtcpRWLock *rwlock)
 
   DmtcpRWLockStatus oldStatus;
   DmtcpRWLockStatus newStatus;
+  int ret;
```
karya0, line 86: same as above.

src/rwlock.cpp
```diff
@@ -24,6 +24,7 @@ int DmtcpRWLockRdLock(DmtcpRWLock *rwlock)
...
```
karya0, line 27: Move it inside the if block.

karya0 (approved):
LGTM modulo declaration scope.

PR #1216, opened 2025-08-06
Changed --mpi to --kernel-loader, updated comments

xuyao0127:
Both MANA and CRAC needs to set the `ADDR_NO_RANDOMIZE` personality, and use `UH_PRELOAD` env var instead of `LD_PRELOAD`. I changed the flag for `dmtcp_launch` from `--mpi` to `--kernel-loader` and updated related comments so that both MANA and CRAC can use the same `--kernel-loader` flag.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new `--kernel-loader` command-line option for advanced plugin support.

* **Improvements**
  * Updated help messages to reflect the new option and clarify environment variable usage.
  * Enhanced environment setup to use `UH_PRELOAD` when kernel loader mode is enabled, improving compatibility with certain plugins.


gc00 (approved):
LGTM.

PR #1223, opened 2025-10-17
Use double_fork() for gzip compression

gc00:
This is required for forked checkpointing. Previously, we forked a gzip, and did a waitpid for it. So, the parent process was stuck in waitpid while the forked process writing to a ckpt image.

By using a new function `double_fork()`, the code is simpler to follow, 35 lines shorter, and it fixes the waitpid bug for forked checkpointing.

And then the forked checkpointing code uses `double_fork()` in a later commit.  This cleans up the code.

**COMMITS:**
4fa3ade: If FORKED_CKPT, grandchild renames .tmp to .dmtcp
&nbsp;&nbsp;&nbsp;&nbsp; (The parent process in dmtcpworker.cpp was renaming, even while the grandchild was still saving the ckpt.)
b15389a: Use double_fork() for gzip compression
f97b275: Use double_fork() if FORKED_CHECKPOINTING
&nbsp;&nbsp;&nbsp;&nbsp; (Note that this commit has the original process send the temp ckpt filename (\*.dmtcp.temp) to the coordinator, and the grandchild currently cannot send the real filename (\*.dmtcp) because the coordinator doesn't recognize it as part of the original computation group.  An alternative would be for the original process to send the real filename (*.dmtcp) even while the grandchild process continues to append to that file.)
8c53751: Remove some compiler warnings

**This is now ready for review.  @karya0, I'll trade you a review of your PR for a review of this PR. :-)**

## Summary by CodeRabbit

* **Refactor**
  * Reworked checkpointing internals for more robust fork-based checkpoint handling and signal management.
  * Internal cleanup of file-descriptor iteration and filesystem path handling for steadier behavior.

* **Bug Fixes**
  * Corrected checkpoint filename handling when forked checkpointing is enabled so final filenames are delivered consistently.

* **Chores**
  * Removed unused local variables in a test to simplify code.

### DIFF
--- a/jalib/jfilesystem.cpp
+++ b/jalib/jfilesystem.cpp
@@ -165,7 +165,7 @@ void jalib::Filesystem::DirName(char *dirname, const char *fname)
     return;
   }
 
-  size_t len = strlen(fname);
+  int len = strlen(fname);
 
   // Remove trailing slashes
   while (len > 0 && fname[len - 1] == '/') {
@@ -173,8 +173,8 @@ void jalib::Filesystem::DirName(char *dirname, const char *fname)
   }
 
   // Find last slash.
-  size_t lastSlash = -1;
-  for (size_t i = 0; i < len; i++) {
+  int lastSlash = -1;
+  for (int i = 0; i < len; i++) {
     if (fname[i] == '/') {
       lastSlash = i;
     }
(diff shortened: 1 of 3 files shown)


xuyao0127 (approved):
The current solution looks good enough to me. So I'm approving this PR. But I think we should maintain a consistent filename on disk and in the coordinator. I'm not sure if the filename reported to the coordinator will be used anywhere, but it may cause confusion when debugging in the future.

PR #1225, opened 2025-11-11
configure: --disable-prgname-prefix

gc00:
* ADDED:  ./configure --disable-prgname-prefix
  * PRGNAME_PREFIX macro created in config.h; The 'DMTCP:' prefix will be prepended to the comm field in 'ps' on restarted processes _only_ if this macro is defined (default).
  * This is needed for the CUDA ckpt API, which uses the 'comm' field on restart
  * The environment variable DMTCP_DISABLE_PRGNAME_PREFIX will have the same effect.

## Summary by CodeRabbit

* **New Features**
  * Added a build-time option (--enable-prgname-prefix / --disable-prgname-prefix) to control applying a process-name prefix. When enabled, builds set a compile-time flag and default to enabled (use_prgname_prefix = yes). Runtime can also honor a disable signal.

* **Documentation**
  * Updated configure help and docs to explain the new option, default, and an environment-variable alternative to disable the prefix at runtime.


configure
```diff
@@ -8073,6 +8079,20 @@ then :
 fi
 
 
+# Check whether --enable-prgname-prefix was given.
```
xuyao0127: In line 1539 you defined `--disable-prgname-prefix`, but here you are checking `--enable-prgname-prefix`. Is this the correct logic?

xuyao0127 (approved):
LGTM

Issue #1232, opened 2026-01-14
Issues with my program using DMTCP_PATH_MAPPING(src/plugin_pathtranslator.cpp) feature during restart


ishanaro4:
Hi @karya0 , My program hangs during restart upon using DMTCP_PATH_MAPPING.
Let me explain my scenario- So lets suppose my path to original program is `/a/b/c` and now after using checkpointing , i copied all the files in `c` folder to `/a/b/d`.I also deleted the older folder `/a/b/c`.
So my PWD = `/a/b/d` and now I define DMTCP_PATH_MAPPING environment variable as `"/a/b/c:/a/b/d"` and i ran dmtcp_restart command with all checkpoint files, but it hangs there forever.
During debugging with gdb attach command with restart processes IDs I see that 

1. DMTCP coordinator - 
`(gdb) bt

```

- #0  0x00001533924a737b in epoll_wait () from /lib64/libc.so.6
- #1  0x000000000040d629 in dmtcp::DmtcpCoordinator::eventLoop (this=0x675e80 <theCoordinator>) at dmtcp_coordinator.cpp:1487
- #2  0x000000000040f934 in main (argc=0, argv=0x7fffff46d538) at dmtcp_coordinator.cpp:1864
```

`
2. DMTCP 'mtcp_restart' thread
`(gdb) bt

```
- #0  0x000014c449fb725e in pthread_barrier_wait () from /lib64/libpthread.so.0
- #1  0x000014c44aa68dfc in dmtcp::SharedData::waitForBarrier (barrierId=...) at shareddata.cpp:318
- #2  0x000014c44aa32f83 in dmtcp_local_barrier (barrier=0x14c44b244ad0 "Socket::Restart_Ns_Send_Queries") at dmtcpplugin.cpp:470
- #3  0x000014c44b234e03 in dmtcp_SocketConnList_EventHook (event=DMTCP_EVENT_RESTART, data=0x0) at ipc/socket/socketconnlist.cpp:77
- #4  0x000014c44aa43d29 in dmtcp::PluginManager::eventHook (event=DMTCP_EVENT_RESTART, data=0x0) at pluginmanager.cpp:137
- #5  0x000014c44aa351a6 in dmtcp::DmtcpWorker::postRestart (ckptReadTime=0) at dmtcpworker.cpp:533
- #6  0x000014c44aa4d6a8 in dmtcp::ThreadList::waitForAllRestored (thread=0x14c448a43008) at threadlist.cpp:718
[...]

PR #1238, opened 2026-02-26
Fix virtual to real path mapping in Plugin Translator on dmtcp_restart using DMTCP_PATH_MAPPING

ishanaro4:
**Problem Description**

This PR addresses an issue previously reported in https://github.com/dmtcp/dmtcp/issues/1232.
When using DMTCP_PATH_MAPPING to relocate checkpoint files to a new directory, repeated path translations result in incorrect path expansion.

**Scenario**

Suppose I have my application executable and related files in /a/b/c, So at the time of checkpoint, it would save this path to all the files inside /a/b/c.
Now I copies contents inside "c" folder including checkpoint files to new location , suppose /q/w/e. Now if I delete the older location and try to run dmtcp_restart with DMTCP_PATH_MAPPING variable i.e below command 
`DMTCP_PATH_MAPPING="/a/b/c:/q/w/e" dmtcp_restart ckpt_*`

**Observed Behaviour**

The first path translation works correctly: `/a/b/c → /q/w/e`.
However, subsequent translations of the same virtual path result in incorrect path construction:
`/a/b/c → /q/w/ee
→ /q/w/eee
→ /q/w/eeee...`
This eventually causes:

> - Invalid paths
> - File not found errors
> - dmtcp hanging during restart

**Root Cause**

The issue appears to originate from the following line:
`int n = snprintf(data->virtualToRealPath.path, PATH_MAX, "%s%s", newp.c_str(), suffix);`
Repeated translations append the suffix multiple times, causing cumulative path corruption.
Additionally, there was no guard preventing already-translated paths from being reprocessed.

**Fix**

1. Corrected the snprintf usage to avoid unintended concatenation.
2. Added a guard condition to prevent re-translation of paths that already contain the mapped prefix.
`!(Util::strStartsWith(virtPath, newp.c_str()))`

**Testing**

> - Reproduced the issue using the scenario described above.
> - Verified correct behavior after fix:
> - Path translation happens only once.
> - No cumulative suffix growth.
> - Restart completes successfully.
Please let me know if additional test cases or automated tests should be added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an edge case in path translation that could produce incorrect translated paths in certain overlaps; preserves existing boundary checks and path-length validation.
  * No changes to public interfaces or exported signatures; behavior and validations remain unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dmtcp/dmtcp/pull/1238)

### DIFF
--- a/src/plugin_pathtranslator.cpp
+++ b/src/plugin_pathtranslator.cpp
@@ -137,7 +137,7 @@ pathTranslator_VirtualToReal(DmtcpEventData_t *data)
   for (const auto &mapping : *pathMapping) {
     const string &oldp = mapping.first;
     const string &newp = mapping.second;
-    if (Util::strStartsWith(virtPath, oldp.c_str()) &&!(Util::strStartsWith(virtPath, newp.c_str()))) {
+    if (Util::strStartsWith(virtPath, oldp.c_str())) {
       // boundary check: next char must be '/' or end
       char next = virtPath[oldp.size()];
       if (next == '/' || next == '\0') {
--- a/src/plugin_pathtranslator.cpp
+++ b/src/plugin_pathtranslator.cpp
@@ -141,12 +141,12 @@ pathTranslator_VirtualToReal(DmtcpEventData_t *data)
       // boundary check: next char must be '/' or end
       char next = virtPath[oldp.size()];
       if (next == '/' || next == '\0') {
-        string original = virtPath;  
-        string replaced = newp + original.substr(oldp.size());
+        string suffix = &virtPath[oldp.size()];  
         int n = snprintf(data->virtualToRealPath.path,
                          PATH_MAX,
-                         "%s",
-                         replaced.c_str());
+                         "%s%s",
+                         newp.c_str(),
+                         suffix.c_str());
         JASSERT(n > 0 && n < PATH_MAX).Text("Translated path exceeds PATH_MAX");
       }
     }
--- a/src/plugin_pathtranslator.cpp
+++ b/src/plugin_pathtranslator.cpp
@@ -137,12 +137,16 @@ pathTranslator_VirtualToReal(DmtcpEventData_t *data)
   for (const auto &mapping : *pathMapping) {
     const string &oldp = mapping.first;
     const string &newp = mapping.second;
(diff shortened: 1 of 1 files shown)


karya0:
> The first path translation works correctly: `/a/b/c → /q/w/e`. However, subsequent translations of the same virtual path result in incorrect path construction: `/a/b/c → /q/w/ee → /q/w/eee → /q/w/eeee...` This eventually causes:

I a not sure what do you mean by subsequent translations. Since the mapping has only one key-value pair, the for loop should only be executed once. 

I think the root cause (as you have also hinted at) is that we are overwriting the original string. If we are translating with mapping "/a/b/c:/q/w/sss" and the input string is "/a/b/c/d.txt",  snprintf will never terminate because we have suffix pointing to "/d.txt". So in a way, the source and destination overlaps will cause infinite loop. The right fix here would be to make a copy of suffix before snprintf is called. I'll add a suggestion.

src/plugin_pathtranslator.cpp
```diff
@@ -137,12 +137,16 @@ pathTranslator_VirtualToReal(DmtcpEventData_t *data)
...
       char next = virtPath[oldp.size()];
       if (next == '/' || next == '\0') {
-        const char* suffix = virtPath + oldp.size();
-        int n = snprintf(data->virtualToRealPath.path, PATH_MAX, "%s%s", newp.c_str(), suffix);
+        string original = virtPath;  
+        string replaced = newp + original.substr(oldp.size());
+        int n = snprintf(data->virtualToRealPath.path,
+                         PATH_MAX,
+                         "%s",
+                         replaced.c_str());
```
karya0:
```suggestion
        string suffix = &virtPath[oldp.size()];  
        int n = snprintf(data->virtualToRealPath.path,
                         PATH_MAX,
                         "%s%s",
                         newp.c_str(),
                         suffix.c_str());
```

karya0 (dismissed):
Please let me know if this fix would work.

gc00 (approved):
LGTM.  Updating for new dot release.

gc00 (approved):
LGTM

PR #1240, opened 2026-03-11
C23 fix for syscallsreal.c and pid_syscallsreal.c

gc00:
`src/syscallsreal.c` and `src/plugin/pid/pid_syscallsreal.c` were going to break when we use a compiler enforcing C23.  C23 requires signatures in fnc pointers (to be aligned with C++).  Luckily, C23 also makes the built-in, `typeof()`, part of the C standard.  GCC and CLANG always provided `typeof()`.  But now, all compliant compilers must provide it.  So, it's safe for us to use it.

And I also cleaned up some minor errors along the way:
 1. sched_setparam:  We were discarding the second argument, param, when we interposed
 2. close_range was created only with Linux 5.9 and glibc-2.34.  We were not testing this
 3. We were missing several include files that had the declarations for the functions that we were interposing on.  (Since C23 requires the signature for fnc. pointers, we now need to include the proper `.h` files.)

I also tested on Rocky Linux 8 (Linux 4.18 and glibc-2.28).  This is an older O/S.  So, hopefully, we're not breaking DMTCP on older Linuxes.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated wrapper macros across system call handlers for improved type safety and consistency.
  * Refactored function pointer types to use more specific signatures rather than generic types.
  * Reorganized system headers and function declarations for clearer code structure.

* **Bug Fixes**
  * Enhanced error handling to abort immediately when required library symbols are missing, improving diagnostics.

### DIFF
--- a/src/dmtcp_coordinator.cpp
+++ b/src/dmtcp_coordinator.cpp
@@ -1578,10 +1578,8 @@ DmtcpCoordinator::addDataSocket(CoordClient *client)
 // Copy name+suffix into short_buf of length len, and truncate name to fit.
 // This keeps only the last component of name (after last '/')
 char *short_name(char short_buf[], char *name, unsigned int len, char *suffix) {
-  char name_copy[strlen(name)+1];
-  memcpy(name_copy, name, strlen(name)+1);
-  char *base_name = strrchr(name_copy, '/') == NULL ?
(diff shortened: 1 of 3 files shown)


xuyao0127 (commented):
LGTM,

xuyao0127 (approved):
LGTM

PR #1246, opened 2026-05-01
Fix missing file creation mode on restart

briangough:
This PR fixes the file creation mode for missing files on restart, by saving the file mode in the FileConnection.

I noticed that files recreated after a restart sometimes get strange modes:

```
-r-x--S--T 1 tex tex    8589 May  1 10:21 out.log    # recreated file
-rw-r----- 1 tex tex    8589 May  1 08:11 out.log.0  # original
```

If the file is removed/renamed after the checkpoint is taken then on restart the variadic function `_real_open` is called with `O_CREAT` but no `mode` argument.

https://github.com/dmtcp/dmtcp/blob/6896e12276a9fe449edb0cf206203ce01b19efe6/src/plugin/ipc/file/fileconnection.cpp#L373

This causes the `va_arg`  call in `_real_open` to take an undefined `mode` value off the stack, leading to unexpected file permissions.

https://github.com/dmtcp/dmtcp/blob/6896e12276a9fe449edb0cf206203ce01b19efe6/src/nosyscallsreal.c#L419-L432

## Summary by CodeRabbit

* **Bug Fixes**
  * Files opened for writing are now properly recreated with their original permissions during checkpoint restart, even if the files are deleted before the restart occurs.

* **Tests**
  * Added test coverage for file mode preservation across restart scenarios.

### DIFF
--- a/src/plugin/ipc/file/fileconnection.cpp
+++ b/src/plugin/ipc/file/fileconnection.cpp
@@ -127,6 +127,7 @@ FileConnection::drain()
   // Read the current file descriptor offset
   _offset = lseek(_fds[0], 0, SEEK_CUR);
   fstat(_fds[0], &statbuf);
+  _mode = statbuf.st_mode;
   _st_dev = statbuf.st_dev;
   _st_ino = statbuf.st_ino;
   _st_size = statbuf.st_size;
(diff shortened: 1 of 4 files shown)


gc00 (approved):
LGTM.  Adding for the new dot release.

PR #1248, opened 2026-05-13
SysV IPC fixes.

karya0:
These changes are backported from the main branch and tailored towards 2.5 branch in particular.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -130,14 +130,14 @@ check: tests
 	  echo "'make check' will fail if run as root."; \
 	  exit 1; \
 	fi
-	@ if python -c 'print "Python exits."' > /dev/null; then \
-	  bash -c "$(LIMIT) && $(top_srcdir)/test/autotest.py ${AUTOTEST} $*"; \
+	@ if python3 -c 'print("Python exits.")' > /dev/null; then \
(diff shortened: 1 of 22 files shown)


gc00:
@karya0 ,
    I've reviewed this.  Since it's an old branch, we don't need a formal github review.  It all looks good.  I have two minor comments to consider before you push it in:

1.   The last commit says: "Randomize the keyed sysv-shm2 key to avoid collisions with concurrent test runs."  Would it be easy to check if a keyed sysv-shm2 key is already reserved by some other concurrent test?  If so, it might be worth adding that test, and warning the user.  Otherwise, it might be difficult to understand a random bug that appears.  (But since this is a back-port, it's fine if you decide you don't need this.)
2.  COMMENT ONLY (NOT A QUESTION):  I notice that you're updating this to work with AARCH64 -- for example a libc_fork wrapper, since AARCH64 doesn't have a fork kernel call.  I'm just curious.  Does that mean that this will often be used with AARCH64?  (Many clusters now are using AARCH64, since it's relatively low power, and still competes in performance with x86_64.  So, if we want to support just one other CPU architecture besides x86_64, then AARCH64 would definitely be the one to choose.)

PR #1257, opened 2026-06-09
Modernize DMTCP C++20 utility usage

karya0:
## Summary
- Add C++20 parsing helpers for integer, port, PID, version, and proc-style parsing paths.
- Replace local membership/string/vector idioms with C++20 helpers such as contains(), starts_with(), string_view, and std::erase where the change is mechanical.
- Simplify DmtcpAlloc aliases and allocator API now that the project builds as C++20.

## Test Plan
- ./configure
- make -j4
- ./test/autotest.py -v dmtcp1
- git diff --check origin/main..HEAD

ASSERT cleanup and test harness changes are intentionally left for separate PRs.

## Summary by CodeRabbit

* **Refactor**
  * Safer, modernized string and container handling for more reliable option/argument/path matching and membership checks.
  * Hardened parsing (integers, ports, version parsing) and improved physical-memory detection to reduce false warnings.

* **Chores**
  * Streamlined allocator and container aliases and tightened environment-variable handling, including added virtual/real PID env vars for improved process identity mapping.


include/dmtcpalloc.h
```diff
@@ -44,216 +44,92 @@ template<typename T>
...
     // Allocate raw memory
-    pointer allocate(size_type n, const void * = NULL)
+    T *allocate(std::size_t n)
     {
-      // void* p = malloc( n * sizeof(T) );
-      // if( p == NULL )
-      // throw std::bad_alloc();
-      // return pointer(p);
-      return pointer(jalib::JAllocDispatcher::malloc(n * sizeof(T)));
+      return static_cast<T *>(jalib::JAllocDispatcher::malloc(n * sizeof(T)));
```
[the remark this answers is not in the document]
karya0: Fixed in the rewritten branch: allocate() now checks element-count overflow before multiplying by sizeof(T), with the required standard includes.

src/plugin/ssh/ssh.cpp
```diff
@@ -372,8 +373,8 @@ prepareForExec(DmtcpEventData_t *data)
   size_t commandStart = 2;
   for (size_t i = 1; i < nargs; ++i) {
     string s = argv[i];
-    if (strcmp(argv[i], "-o") == 0) {
-      if (strcmp(argv[i + 1], "StrictHostKeyChecking=no") == 0) {
+    if (Util::strEquals(argv[i], "-o")) {
+      if (Util::strEquals(argv[i + 1], "StrictHostKeyChecking=no")) {
         noStrictChecking = true;
```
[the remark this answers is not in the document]
karya0: Fixed in the rewritten branch: SSH option parsing now stops before the argv NULL sentinel and returns without rewriting when -o has no argument.

src/plugin/ssh/ssh.cpp
```diff
@@ -601,7 +606,7 @@ updateCoordHost()
         continue;
       } else {
         JASSERT(sizeof localhostIPAddr == sizeof s->sin_addr);
-        if ( strncmp( name, hostname, sizeof hostname ) == 0 ) {
+        if (std::string_view(name) == std::string_view(hostname)) {
```
gc00: Keep the old code here.  As pointed out by coderabbit, the C string hostname is of unlimited size.  The previous code uses `sizeof hostname` to bound a very long hostname.

src/plugin/ssh/ssh.cpp
```diff
@@ -617,7 +622,7 @@ updateCoordHost()
     }
     if (at_least_one_match) {
       success = true;  // Call it a success even if hostname != name
-      if ( strncmp( name, hostname, sizeof hostname ) != 0 ) {
+      if (std::string_view(name) != std::string_view(hostname)) {
```
gc00: [ Same as earlier comment ]
Keep the old code here.  As pointed out by coderabbit, the C string hostname is of unlimited size.  The previous code uses `sizeof hostname` to bound a very long hostname.

src/dmtcp_launch.cpp
```diff
@@ -395,8 +397,7 @@ processArgs(int *orig_argc, const char ***orig_argv)
       personality(ADDR_NO_RANDOMIZE);
 #endif
       shift;
-    } else if ((s.length() > 2 && s.substr(0, 2) == "--") ||
-               (s.length() > 1 && s.substr(0, 1) == "-")) {
+    } else if (s.length() > 1 && s.starts_with("-")) {
```
gc00: Elsewhere, I mentioned the coderabbit comment.  This `else if` and the next one should be reversed.  coderabbit already has a suggested fix.

src/dmtcprestartinternal.cpp
```diff
@@ -760,8 +760,7 @@ DmtcpRestart::DmtcpRestart(int argc, char **argv, const string& binaryName, cons
       // Just in case a non-standard version of setenv is being used:
       setenv(ENV_VAR_QUIET, getenv(ENV_VAR_QUIET), 1);
       shift;
-    } else if ((s.length() > 2 && s.substr(0, 2) == "--") ||
-               (s.length() > 1 && s.substr(0, 1) == "-")) {
+    } else if (s.length() > 1 && s.starts_with("-")) {
```
gc00: similar issue to above:
Elsewhere, I mentioned the coderabbit comment. This else if and the next one should be reversed. coderabbit already has a suggested fix.

include/util.h
```diff
@@ -113,6 +118,91 @@ int changeFd(int oldfd, int newfd);
...
+{
+  return ch >= '0' && ch <= '9';
+}
+
+template <typename T>
+inline bool parseInteger(std::string_view text, T *value)
+{
+  auto result = std::from_chars(text.data(), text.data() + text.size(), *value);
+  return result.ec == std::errc() && result.ptr == text.data() + text.size();
+}
```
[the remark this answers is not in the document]
gc00: coderabbit proposes:
```
if (value == nullptr) {
    return false;
}
```
But maybe we should do JASSERT here, instead of blindly returning `false`.
Should the caller of parseInteger currently checks this, but it should be enforced, instead of relying on the caller to check.

include/util.h
```diff
@@ -113,6 +118,60 @@ int changeFd(int oldfd, int newfd);
...
+
+inline bool parsePortNumber(std::string_view text, int *port)
```
gc00: I would prefer the name `parsePortInteger`.  It's an integer, not just a number.  (And the include file would need to be changed too.)

include/dmtcpalloc.h
```diff
@@ -44,216 +46,95 @@ template<typename T>
...
-      ::new ((void *)p)X(std::forward<Args>(args) ...);
-    }
-
-    // In-place destruction
-    template<typename X>
-    void destroy(X *p)
-    {
-      p->~X();
-    }
+    DmtcpAlloc& operator=(const DmtcpAlloc&) noexcept = default;
```
gc00, line 63: overloading with `operator=` can be a trap.  Users see `=` and don't realize that there's a non-local definition of a keyword operator.  For people whose native lang is not C++, I'm worried.  But maybe comments about `noexcept` and `default` will help.

src/plugin/file/fileconnection.cpp
```diff
@@ -908,7 +908,7 @@ PosixMQConnection::drain()
     unsigned prio;
     ssize_t numBytes = _real_mq_receive(_fds[0], buf, attr.mq_msgsize, &prio);
     JASSERT(numBytes != -1) (JASSERT_ERRNO);
-    _msgInQueue.push_back(jalib::JBuffer((const char *)buf, numBytes));
+    _msgInQueue.emplace_back(buf, buf + numBytes);
```
gc00, line 912: Just because `emplace_back` is not familiar to those using C, and the name is obscure, can we add a small comment, perhaps on the same line as `emplace_back`:
`// append buf to end of vector`

src/plugin/svipc/sysvipc.cpp
```diff
@@ -1134,8 +1134,9 @@ MsgQueue::preCheckpoint()
     for (size_t i = 0; i < _qnum; i++) {
       ssize_t numBytes = _real_msgrcv(_realId, msgBuf, size, 0, 0);
       JASSERT(numBytes != -1) (_id) (JASSERT_ERRNO);
-      _msgInQueue.push_back(jalib::JBuffer((const char *)msgBuf,
-                                           numBytes + sizeof(msgBuf->mtype)));
+      char *msgData = (char *)msgBuf;
+      _msgInQueue.emplace_back(msgData,
```
gc00, line 1139: And here also, can we add a comment:
`// append msgData to end of vector`

gc00 (approved):
COMMENT:  I do like the limited use of C++20 extensions here.  It fixes some of my prior complaints about C++.  :-)  The API is intuitive, and the API will be transparent when debugging with GDB.

----

coderabbit caught an important flaw.  Even though it's unrelated to this PR, let's fold it into this PR (and I also comment inline in the code):
  src/dmtcp_launch.cpp (1)
399-404: ⚠️ Potential issue | 🟠 Major | ⚡ Quick win

 `--` is now treated as invalid instead of end-of-options.

  The starts_with("-") invalid-argument branch runs before the s == "--" check, so -- can never reach its intended handler.
[ And coderabbit has a suggested fix that you can plug in ]

----

include/dmtcpalloc.h
karya0: Added a comment.

src/plugin/ssh/ssh.cpp
karya0: Fixed with a fixed size view.

src/plugin/ssh/ssh.cpp
karya0: Added fixed-size view.

PR #1262, opened 2026-06-23
Fix inverted diagnostic conditions

karya0:
## Summary
- Fix KVDB warning polarity so long IDs trigger the truncation warning instead of valid IDs
- Fix the TLS auxv stack sanity check to use the intended upper bound, matching mtcp_check_vdso.c
- Clarify ASSERT/WARN documentation: raw macros log when the condition is false

## Audit
- Ran parallel read-only audits over core, plugins, and assert/logger helpers for WARN/ASSERT polarity issues
- Plugin and helper audits found no additional runtime polarity bugs

## Validation
- python3 util/dmtcp-style.py src/kvdb.cpp src/tls.cpp src/dmtcp_assert.h
- git diff --check
- make -j8
- make -C test
- make -C test unit/dmtcp_unit_tests
- ./test/autotest.py -v dmtcp1
- ./test/autotest.py --suite unit

## Summary by CodeRabbit

* **Documentation**
  * Clarified assertion helper documentation to detail behavior differences among assertion types and error value handling.

* **Bug Fixes**
  * Fixed inverted condition in ID validation warning logic.
  * Corrected bounds checking validation in system information processing.

### DIFF
--- a/src/dmtcp_assert.h
+++ b/src/dmtcp_assert.h
@@ -25,10 +25,11 @@ inline constexpr int kAssertFailureExitCode = 99;
 
 /*
  * Assertion helper guide:
- * - ASSERT(cond, "msg {}", arg) logs and exits; WARN(cond, ...) logs and
- *   continues.  NOTE(...) and TRACE(...) use the same formatter and are
- *   controlled by the runtime log level.  ASSERT_ERRNO/WARN_ERRNO add the
- *   errno value captured at the failing check.
+ * - ASSERT(cond, "msg {}", arg) logs and exits when cond is false;
+ *   WARN(cond, ...) logs and continues when cond is false.  NOTE(...) and
+ *   TRACE(...) use the same formatter and are controlled by the runtime log
+ *   level.  ASSERT_ERRNO/WARN_ERRNO add the errno value captured at the
+ *   failing check.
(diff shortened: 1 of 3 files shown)


gc00 (approved):
I will approve it for now.  But **please** re-check the logic at line 390 of src/tls.cpp.
**Then please add to the conversation why the issue below is not an issue, even if you change nothing.**

Something here is very strange.  The code worked fine in the old code when we asserted that:
> `(void **)stack[-1] >= stack + 100000`

**That should have asserted out.**

Now, you've fixed the inequality.  But I'm nervous about pasting over a potentially more serious bug.  **Why was the assert not erroring out in the previous code?**

karya0:
> I will approve it for now. But **please** re-check the logic at line 390 of src/tls.cpp. **Then please add to the conversation why the issue below is not an issue, even if you change nothing.**
> 
> Something here is very strange. The code worked fine in the old code when we asserted that:
> 
> > `(void **)stack[-1] >= stack + 100000`
> 
> **That should have asserted out.**
> 
> Now, you've fixed the inequality. But I'm nervous about pasting over a potentially more serious bug. **Why was the assert not erroring out in the previous code?**

As I wrote in the email, this is legacy dead code that's why this path was never exercised. I'll leave the fix as is for now.

PR #1264, opened 2026-06-25
Skip autotest on Python < 3.7

karya0:
## Summary
- Skip autotest on Python < 3.7 with a clear message instead of carrying a Python 3.6 compatibility shim.
- Keep Python 3.7-compatible replacements for newer helpers used by harness diagnostics.
- Add a regression test for the Python < 3.7 skip path.

## Test Plan
- `/tmp/Python-3.6.15/python test/autotest.py --list dmtcp1`
- `python3 test/autotest.py --list dmtcp1`
- `python3 -m unittest autotest_test.DmtcpTestHarnessUnitTest.test_autotest_skips_before_python37 autotest_test.DmtcpTestHarnessUnitTest.test_command_failure_details_are_capped_and_actionable autotest_test.DmtcpTestHarnessUnitTest.test_autotest_suite_all_runs_command_suites_then_integration`
- `python3 -m py_compile test/autotest.py test/autotest_test.py`
- `/tmp/Python-3.6.15/python -m py_compile test/autotest.py test/autotest_test.py`
- `git diff --check`

## Notes
- Full `python3 test/autotest_test.py` still has pre-existing local config-sensitive failures: missing `autotest_config.HAS_SYS_MQ_OPEN`, and `waitid-syscall` is not registered in this configured tree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Autotest now exits early with a clear message when run on Python versions earlier than 3.7.
  * Command and execution failure messages are now easier to read, with improved shell-style formatting in logs and summaries.
  * Display names for coordinator-related runs now render consistently without the extra prefix.

### DIFF
--- a/test/autotest_test.py
+++ b/test/autotest_test.py
@@ -1628,6 +1628,47 @@ class DmtcpTestHarnessUnitTest(unittest.TestCase):
         self.assertEqual(result.returncode, 2)
         self.assertIn("No tests selected", result.stderr)
 
+    def test_autotest_skips_before_python37(self):
+        with tempfile.TemporaryDirectory() as tempdir:
+            temp_path = pathlib.Path(tempdir)
+            (temp_path / "sitecustomize.py").write_text(
+                "import collections\n"
+                "import sys\n"
+                "VersionInfo = collections.namedtuple(\n"
+                "    'version_info', 'major minor micro releaselevel serial')\n"
+                "sys.version_info = VersionInfo(3, 6, 8, 'final', 0)\n"
+            )
+            (temp_path / "dataclasses.py").write_text(
+                "raise ModuleNotFoundError('No module named dataclasses')\n"
+            )
+            env = os.environ.copy()
+            env["PYTHONPATH"] = os.pathsep.join(
+                [str(temp_path), env.get("PYTHONPATH", "")]
+            )
+
+            result = subprocess.run(
+                [
+                    sys.executable,
+                    str(ROOT / "test" / "autotest.py"),
+                    "--list",
+                    "dmtcp1",
+                ],
+                cwd=str(ROOT),
+                env=env,
+                text=True,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                check=False,
+                timeout=10,
+            )
+
+        self.assertEqual(result.returncode, 0)
+        self.assertEqual(
+            result.stdout,
+            "Skipping autotest; Python 3.7+ is required.\n",
+        )
+        self.assertEqual(result.stderr, "")
+
     def test_autotest_list_rejects_unknown_test(self):
         result = subprocess.run(
             [
--- a/test/autotest.py
+++ b/test/autotest.py
@@ -17,11 +17,16 @@ import tempfile
 import threading
 import time
 import traceback
-from dataclasses import dataclass, field, replace
 from random import sample
 from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 import autotest_config
+
+if sys.version_info < (3, 7):
+    print("Skipping autotest; Python 3.7+ is required.", flush=True)
+    sys.exit(0)
+
+from dataclasses import dataclass, field, replace
 
 
 ROOT = pathlib.Path(__file__).resolve().parents[1]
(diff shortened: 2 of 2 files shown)


gc00 (commented):
@karya0,
    dataclasses and dataclass are not available in Python 3.6.  So, it's still failing in Rocky Linux 8 with Python 3.6.

If it would be difficult to support the new autotest on Python 3.6, it's fine to just configure and build, but tell the user that autotest requires Python 3.7.  Install Python 3.7+ if they want to use autotest.

gc00 (changes requested):
@karya0 , I tested this on our local login host.  When I even do `make check-dmtcp1`, it randomly fails on either the first restart or the second restart.  At that time, I'm seeing the following processes:
> gene     1411569  2.2  0.0  88144 18824 pts/1    T    11:42 python3 ./test/autotest.py dmtcp1
> gene     1411572  0.7  0.0      0     0 pts/1    Z    11:42   0:00 [dmtcp1] <defunct>
> gene     1411588  0.4  0.0      0     0 pts/1    Z    11:42   0:00 [DMTCP:dmtcp1] <defunct>

[ Several cycles of ckpt/restart work on dmtcp1, running manually without autotest. ]

It's also failing:
   test_cleanup_signals_worker_process_group
   test_start_coordinator_passes_extra_arg
   barrier        ckpt:PASSED; rstr:  [HANGS]

My advice is to simply tell the user that we don't support Python 3.6.  They can upgrade to Python 3.7, or run without doing autotest and unit tests.  It's not worth our effort to temporarily support Python 3.6, until it becomes obsolete.

In case you want to try a quick fix, but I advise changing it and just warn the user that we don't support Python 3.6, Rocky Linux 8 is running:
Rocky Linux release 8.10
Python 3.6.8 
gcc version 11.2.1 20220127 (Red Hat 11.2.1-9) (GCC)
CFLAGS: -g -O2 -std=gnu11 -Wall -g3 -O0 -DDEBUG -pthread
CXXFLAGS: -g -O2 -Wall -g3 -O0 -DDEBUG
CPPFLAGS: -Wall -g3 -O0 -DDEBUG
ldd (GNU libc) 2.28
LDFLAGS:
openjdk version "26" 2026-03-17
OpenJDK Runtime Environment (Red_Hat-26.0.0.0.35-2) (build 26+35)
OpenJDK 64-Bit Server VM (Red_Hat-26.0.0.0.35-2) (build 26+35, mixed mode, sharing)
javac 26

gc00 (approved):
Thanks.  I tested on Rocky Linux 8, and it always replies:  `Skipping autotest; Python 3.7+ is required.`
This is perfect.

xuyao00:
On Fedora 43 (python 3.14.4) and Ubuntu 26.04 (python 3.14.4), this test failed with the following error message:
```
Harness tests
-------------
  autotest-unit  FAIL. exit=1 Ran 113 tests in 2.430s
    command: /usr/bin/python3 ./autotest_test.py
    cwd: /home/yaoxu/Workspace/dmtcp/test
    output tail (last 14 lines):
      .....................F...........................................................................................
      ======================================================================
      FAIL: test_autotest_skips_before_python37 (__main__.DmtcpTestHarnessUnitTest.test_autotest_skips_before_python37)
      ----------------------------------------------------------------------
      Traceback (most recent call last):
        File "/home/yaoxu/Workspace/dmtcp/test/./autotest_test.py", line 1665, in test_autotest_skips_before_python37
          self.assertEqual(result.returncode, 0)
          ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
      AssertionError: 1 != 0

      ----------------------------------------------------------------------
      Ran 113 tests in 2.430s

      FAILED (failures=1)
```

gc00:
@karya0 ,
    There is a strange bug on this PR, that seems to exhibit only for aarch64/ARM64.  @xuyao0127 is testing it on a Linux VM on his Mac.  The unit test, above, is failing for him.
    Also, the assertEqual statement below has a trailing comma in its arguments, as if it has 3 args.  That seems weird.
    Does the bug seem obvious to you?

@xuyao0127 :
    I don't see this bug on Ubuntu 22.04.  But there is something funny going on, when I try to isolate the path components of the file path.  Could you do:
    cd DMTCP_ROOT/test
    python3 -mpdb autotest_test.py
    (pdb) break 1632
and then step through that function line by line, checking what is going wrong.  Good things to print out as you step through it might be:
    print( temp_path / "sitecustomize.py" )    # read that file
    print( str(ROOT / "test" / "autotest.py") )
    print( sys.executable )
    print( result.returncode )  # and if it's an error code, try to find out what the process error was.

And the statement:
  self.assertEqual(
            result.stdout,
           "Skipping autotest; Python 3.7+ is required.\n",
  )
seems suspicious.  First, it has a trailing comma (illegal syntax?).  Second it's not clear what it's comparing for equality.  But your test seems to show that it doesn't even get that far on aarch64.

test/autotest_test.py
```diff
@@ -1628,6 +1628,47 @@ def test_autotest_run_rejects_empty_selection(self):
         self.assertEqual(result.returncode, 2)
         self.assertIn("No tests selected", result.stderr)
 
+    def test_autotest_skips_before_python37(self):
```
xuyao00, line 1631: Bad name, should be "test_if_autotest_skips_before_python37"

PR #1265, opened 2026-06-26
close_range() is not defined prior to glibc-2.34

gc00:
[ This is a trivial one-liner to review.  The comments here say it all. ]

Some platforms like Rocky Linux 8 (on glibc-2.28) don't yet have glibc-2.34.
This conditionally defines close_range for `src/wrappers.cpp`.  We missed that one in conditionally defining elsewhere.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved platform compatibility by only enabling a specific system call wrapper on supported Linux and library versions.
  * Prevents build or runtime issues on environments that do not meet the minimum version requirements.

### DIFF
--- a/src/wrappers.cpp
+++ b/src/wrappers.cpp
@@ -393,6 +393,7 @@ close(int fd)
   return ret;
 }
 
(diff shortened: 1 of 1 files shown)


xuyao0127 (approved):
LGTM

PR #1266, opened 2026-06-26
Fix StackHead, used for dwcas, cmpxchg16b, etc.

gc00:
[ This is a trivial one-liner to review.  The commit comments (shown here) say it all. ]

 * jalloc.cpp uses __atomic_compare_exchange()
 * That compiles to inline Intel assembly (cmpxchg16b) (and similarly for aarch64/ARM64 and RISC-V, but sometimes needs -mcx16 or -march=native to force it.
 * For 16-byte data (StackHead), it must be 16-byte aligned for inline assembly; this guarantees that.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory allocation stability in low-level operations, reducing the risk of rare alignment-related issues.

### DIFF
--- a/jalib/jalloc.cpp
+++ b/jalib/jalloc.cpp
@@ -326,7 +326,7 @@ class JFixedAllocStack
         char buf[N];
       };
     };
-    struct StackHead {
+    struct alignas(16) StackHead {
       uintptr_t counter;
       FreeItem* node;
     };


gc00:
I won't do the coderabbit suggestion.  The whole point of StackHead is that it will _always_ be a 16-byte value, to fit in 128-bit inline compare (dwcas).  It will never have an additional field.

xuyao0127 (approved):
LGTM

PR #1269, opened 2026-06-28
Install version.h under the dmtcp include directory

felixonmars:
dmtcp.h includes "dmtcp/version.h", but the automake install rule flattened include/dmtcp/version.h to $includedir/version.h.

Install version.h under $includedir/dmtcp instead so dmtcp.h resolves its own include and the installation does not pollute the global include namespace with version.h.

## Summary by CodeRabbit

* **Chores**
  * Updated the installed header layout so the main public header and the version-specific header are placed into separate include locations.
  * Primary header remains in the standard include path, while the version header is now installed under a dedicated `dmtcp` subdirectory to keep header organization clearer and more predictable during installation.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -15,6 +15,7 @@ AUTOMAKE_OPTIONS = foreign
 
 jalibdir=$(top_srcdir)/jalib
 dmtcpincludedir=$(top_srcdir)/include
+dmtcpversionincludedir = $(includedir)/dmtcp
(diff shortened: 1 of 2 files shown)


gc00 (approved):
@felixonmars , Thanks for this valuable PR.  We were always testing only locally, and failed to notice the problem when doing global install.
I also propagated from `src/Makefile.am` to `src/Makefile.in` for your commit, while keeping your authorship.

PR #1271, opened 2026-07-02
Handle thread stack guard page in glibc >= 2.42

xuyao0127:
Previously, PR https://github.com/dmtcp/dmtcp/pull/1234 tried to solve this problem introduced in glibc 2.42. The old fix has two issues:
1. Completely removes the guard page for all thread stacks. Therefore, permanently disables stack-overflow protection for every thread.
2. The "guard_size" returned from the function `pthread_attr_getguardsize()` may not reflect the correct guard page size. On some 64-bit ARM machines, the actual guard page is 64K, but the function still returns 4k.

This PR fixed both problems.
1. This PR only remove the guard page at checkpoint and will restore the guard page after restart or resume.
2. This PR uses max of guard page size and EXEC_PAGESIZE to determine the actual guard page size.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved checkpoint and restart reliability by properly managing per-thread lightweight stack guard regions during checkpoint writes and restart.
  * Prevented memory scanning from faulting on protected guard pages while creating checkpoints.
  * Reinstalled stack guard protection before threads resume after restart.
* **New Features**
  * Added automatic tracking of lightweight per-thread stack guard metadata on supported glibc versions for safer, more accurate guard handling.

### DIFF
--- a/src/threadinfo.h
+++ b/src/threadinfo.h
@@ -88,6 +88,14 @@ struct Thread {
 
   uint32_t wrapperLockCount;
 
+  // Lightweight stack guard region installed by glibc >= 2.42 at the low end
+  // of the thread's stack via madvise(MADV_GUARD_INSTALL). Recorded when the
+  // thread is created so that the checkpoint thread can temporarily remove it
+  // while writing the checkpoint image and reinstall it afterward and on
+  // restart. guardSize == 0 means no guard region is tracked for this thread.
+  void *guardAddr;
+  size_t guardSize;
+
   Thread *next;
   Thread *prev;
 };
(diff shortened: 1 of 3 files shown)


gc00:
@xuyao0127 , Please note that PR #1284 has a first commit that wants to "[Avoid guarded statics in early plugin init](https://github.com/dmtcp/dmtcp/pull/1284/changes/fa27d71f19ad60b80a7373b69752f0f22a72ce83)".  This older PR is also about guard pages.  Could you briefly look at PR #1284, and talk to @karya0 about whether your PR and the first commit of his PR are both working on the same issue?  Thanks.

PR #1272, opened 2026-07-02
Fixed harness test with Python 3.13+

xuyao0127:
PR https://github.com/dmtcp/dmtcp/pull/1264 creates a bug with Python 3.13+.

On Python 3.13+, importing concurrent.futures transitively imports dataclasses (via _colorize -> traceback -> logging -> concurrent.futures._base). When simulating Python 3.6 via a fake sitecustomize.py (which overrides sys.version_info and provides a dummy dataclasses.py that raises ModuleNotFoundError), the script crashes at `import concurrent.futures` before the version check on line 25 is ever reached. Moving the version check (which only needs sys) to the very top ensures it runs before any module resolution can hit the fake dataclasses.

## Summary by CodeRabbit

* **Chores**
  * Improved startup checks to fail faster with a clear message when Python 3.7 or newer isn’t being used.
  * Removed a redundant duplicate import and consolidated version validation for cleaner execution.
* **Tests**
  * Updated the relevant unit test name to better reflect its behavior around skipping when Python is below 3.7.

### DIFF
--- a/test/autotest.py
+++ b/test/autotest.py
@@ -1,5 +1,11 @@
 #!/usr/bin/env python3
 
+import sys
+
+if sys.version_info < (3, 7):
+    print("Skipping autotest; Python 3.7+ is required.", flush=True)
+    sys.exit(0)
+
 import argparse
 import concurrent.futures
 from collections import Counter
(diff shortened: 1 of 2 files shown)


gc00 (approved):
LGTM

PR #1273, opened 2026-07-04
Avoid recreating pre-existing pipe fds on restart

xuyao0127:
Libraries like Hydra create pipes that are not saved or recreated by DMTCP. When recreating pipes for CUDA at restart, avoid recreating these pre-existing pipes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA checkpoint/restore behavior so saved file descriptor slots are restored more safely when other components already use some of them.
  * Reduced the chance of restart failures caused by file descriptor conflicts.

* **Tests**
  * Added build support for a CUDA test target to help validate checkpoint/restore behavior.

### DIFF
--- a/plugin/cuda/cuda-ckpt.cpp
+++ b/plugin/cuda/cuda-ckpt.cpp
@@ -112,14 +112,25 @@ int recreate_pipes(pipe_info_t *pipe_fd_array, int num_fds) {
   pipe_map_t pipe_map[MAX_PIPE_FDS] = {0};
   int map_count = 0;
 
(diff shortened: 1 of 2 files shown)


gc00 (approved):
Before pushing this in, please add a FIXME comment saying that eventually, we hope that DMTCP-5.0 will stop promoting pipes to sockets.  At that time, we can delete certain code here, when we do the port of this plugin from 4.2 to 5.0.

Coderabbit still reviewing at this time.

Approving now.

PR #1281, opened 2026-07-11
mtcp_restart: print colliding region details

gc00:
mmap_fixed_noreplace() would fail with only "error 17 mapping ..." (no indication of what's already there), forcing a live gdb attach to diagnose. Add printOverlappingAreas(): on any EEXIST (or the pre-4.17 kernel emulation path), scan /proc/self/maps and print every region overlapping the requested range (address, perms, flags, name), so a collision is diagnosable directly from mtcp_restart's own stderr output.

*(This is trivial to review, but quite useful.)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostics for memory-mapping conflicts and failures.
  * Added details about overlapping memory regions when a requested mapping cannot be placed at the specified address.


gc00:
This commit is now inside PR #1279.

PR #1284, opened 2026-07-14
Fix TSAN pthread startup under DMTCP

karya0:
Summary:
- Fix TSAN startup under DMTCP by making thread initialization robust when
  sanitizer/libc helper paths enter DMTCP before the normal pthread wrapper
  handoff is complete.
- Consolidate current-thread state setup through `ThreadList::init()` and
  `dmtcp_get_current_thread()`, so initialization is idempotent regardless of
  whether it is reached from startup, pthread child entry, wrapper-lock paths,
  or pthread lifecycle events.
- Keep `motherofall` setup centralized and serialized while allowing other
  early threads to acquire their own DMTCP `Thread` records safely.
- Avoid guarded function-local static initialization in early plugin enablement
  checks, since TSAN startup can trip over the compiler-emitted
  `__cxa_guard_*` path before the sanitizer runtime is fully initialized.
- Simplify virtual TID lifecycle handling by carrying pthread/tid data through
  pthread events and letting the PID plugin update mappings from that single
  event path.
- Add launch-only TSAN coverage with `tsan-target`, including the harness
  metadata needed to keep address-space-limit tests serial.

Motivation:
ThreadSanitizer can trigger allocator, pthread, and libc helper activity before
DMTCP has reached the narrow call path that historically created per-thread
state. That left early wrapper entry dependent on ordering details in the
loader/sanitizer startup sequence. This PR makes the thread initialization
sequence self-healing: if a DMTCP path needs the current thread and it has not
been initialized yet, the shared init path creates the right state exactly once.

The init routine causing deadlock/crash with guarded statics like in this example:
```c++
bool pluginEnabled()
{
  static bool enabled = internalPluginEnabled("PID");
  return enabled;
}
```

That shape asks the compiler to emit `__cxa_guard_acquire/release` around the
first call. Under TSAN, the guard path can enter sanitizer locking while TSAN is
still initializing. The fix is to use file-local POD cache state and prime it
during DMTCP's single-threaded initialization path:

```c++
static int pluginEnabled = -1;

int pluginEnabledCached()
{
  if (pluginEnabled == -1) {
    pluginEnabled = internalPluginEnabled("PID") ? 1 : 0;
  }
  return pluginEnabled;
}
```

Tests:
- `git diff --check`
- `make -j8`
- `./test/unit/dmtcp_unit_tests`
- `python3 test/autotest_test.py`
- `./test/autotest.py --jobs 8 -v tsan-target mmap-noreserve dmtcp1 pthread1 pthread2 pthread5 pthread6 gettid pthread_atfork1 pthread_atfork2 forkexec vfork1 vfork2`

Note:
- `tsan-target` is launch-only for now; checkpointing the TSAN target still
  stalls before producing an image.

## Summary by CodeRabbit

* **New Features**
  * Added thread and process identity information to pthread-related events.
  * Improved thread initialization, lifecycle tracking, and checkpoint/restart handling.
  * Added a ThreadSanitizer-based threading test.

* **Bug Fixes**
  * Improved synchronization behavior during startup and thread creation.
  * Preserved syslog identity and checkpoint directory settings more reliably across lifecycle events.
  * Improved cleanup of exited thread mappings.

* **Tests**
  * Added automated coverage for thread synchronization, restart scenarios, and pthread event data.


karya0:
@coderabbitai review

karya0:
@coderabbitai review

gc00:
@karya0 , I assume that you're still working on this PR.  When it's ready for review, could you add a better explanation of what this is doing?  I'm seeing changes to very many files in some kind of pattern, but I'm not seeing the initial summary that explains why propagating this pattern to many files is useful.
Thanks.

karya0:
Hi @gc00 , I have updated the PR summary. For reviewing this PR, it will be best to review the commits separately. The first commit is mostly trivial static-init fix. The second commit is more involved with thread initialization route refactor/simplication

src/threadlist.cpp
```diff
@@ -199,25 +187,62 @@ ThreadList::resetOnFork()
...
-    prepareThread(th, NULL, NULL);
-    /* Set up caller as one of our threads so we can work on it */
-    initThread(th);
-    // Publish last (release), so a waiting thread sees it fully initialized.
-    __atomic_store_n(&motherofall, th, __ATOMIC_RELEASE);
+    motherofall = &motherofallStorage;
+    th = motherofall;
+  } else {
+    th = (Thread *) JALLOC_MALLOC(sizeof(Thread));
   }
```
gc00, line 205: Shouldn't the next few lines (until `ThreadList::addToActiveList(th);`) be moved to an auxiliary function with a meaningful name, to enable people reading the code?  
Maybe an appropriate function might be:
    ``ThreadList::initializeNewThreadForActiveList(th);`

For inspiration, note that previously, we seemed to use a combination of `ThreadList::getNewThread()` and `ThreadList::prepareThread()`, and this new code combines both functions.

src/threadwrappers.cpp
```diff
@@ -111,18 +91,16 @@ pthread_create(pthread_t *pth,
...
 
-  Thread *newThread = ThreadList::getNewThread(start_routine, arg);
-  ThreadSync::wrapperExecutionLockLockForNewThread(newThread);
-  ASSERT_NE(0u,
-            newThread->wrapperLockCount,
-            "new thread wrapper lock was not pre-acquired: tid={}",
-            newThread->tid);
+  PthreadStart *start = (PthreadStart *) JALLOC_MALLOC(sizeof(*start));
+  start->fn = start_routine;
+  start->arg = arg;
```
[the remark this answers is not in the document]
gc00: As coderabbit says, "minor but quick win".

src/threadlist.cpp
karya0: That's a valid observation. I want to create a follow-on PR to cleanup this - creating a constructor in Thread which then avoids the malloc, etc., here along with the trivial initialization of fields. For this PR, I wanted to keep the "everything is local" theme at the expense of longer functions.

gc00 (changes requested):
*This comment is obsolete.  Please see the new comment, lower down.*

src/threadlist.cpp
gc00: For ease of reading the code, can you add a helper function now, and then revise or delete it later?
One can think of the name of the helper function as a kind of "executable comment" telling the reader what we're going to do for the next 20 or 30 lines of code.

gc00:
@karya0 , I've now created a new PR #1288.  Once we agree, that will replace the old PR #1261.

This new PR is based on top of your own PR #1284.  I'd like to jointly review both PRs (1284 and 1288), since they're tightly related.  In particular, PR #1288 has a commit "TSAN: detection of TSAN helper (background) thread" that interacts closely with your own commit.  I want to make sure that you've looked closely at my commit, and see how it is rebased on top of yours.  After that, we can choose to modify my commit or your PR or both, once we've analyzed the interaction.

Separately, note that my PR has a more advanced autotest for TSAN.  So, my PR is removing any changes you made in the autotest directory for TSAN, before adding my own changes.

src/threadlist.cpp
karya0: @gc00 : Just added ThreadInfo and a default constructor so this is more readable now. I want to do a little more refactor but that might conflict with your PR, so let's chat more on the Saturday call.

gc00: The ThreadInfo works well for now (before your refactor).

gc00 (changes requested):
This is clearly a much-needed PR.  It makes the thread initialization portion much more robust.  I'm listing this as "Request changes" for now (Copilot has multiple suggestions), but ping me if you want a quick "review and approve".

There are a few changes I'd like to see:
1. Coderabbit has a review outstanding about " A NULL check immediately after JALLOC_MALLOC and returning EAGAIN keeps the change minimal while avoiding a null dereference on allocation failure".  This seems valid to me, but you can decide.
2. I asked Copilot to also do a review.  Copilot found multiple interesting issues.  Could you look them over and do them or comment?
3. Could you rebase from main?  This will also help my PR #1288, which is stacked on top of this PR.

Mainly, I'm still listing "Request changes" because Copilot found multiple other issues to look at.  I haven't looked at those closely yet.  I can return to "review and approve" after we've looked at that.  (Some of the issues that the TSAN work is uncovering can be subtle.)

src/pluginmanager.cpp
```diff
@@ -103,7 +107,7 @@ static InternalPluginEntry internalPlugins[] = {
   { &dlPlugin, false }
 };
 
-static pthread_once_t internalPluginInitOnce = PTHREAD_ONCE_INIT;
+static bool internalPluginStateInitialized = false;
```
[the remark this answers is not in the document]
karya0: Verified against this branch. I am not adding a pthread_once/atomic init state here: internal plugin state is initialized during early loader/plugin setup, and dmtcp_initialize_plugin() primes the wrapper-level caches before user threads run. The current push adds a code comment documenting that invariant. Comment prepared with Codex.

src/util_misc.cpp
```diff
@@ -137,31 +137,34 @@ Util::readBooleanEnv(const char *envName, bool defaultValue)
...
+  const char *minorEnd = strchr(minorStart, '.');
+  if (minorEnd == NULL) {
+    minorEnd = minorStart + strlen(minorStart);
+  }
+  ASSERT(parseInteger(std::string_view(minorStart, minorEnd - minorStart),
+                      &parsed.minor),
+         "failed to parse glibc minor version: version={}",
+         versionText);
+  cachedVersion = parsed;
+  return parsed;
```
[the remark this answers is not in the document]
karya0: Fixed the branch-introduced lazy-init issues in the current push: util glibc/page caches now use explicit atomic loads/stores, pageMask no longer caches separately, syslog/processinfo/nohijack string pointers publish with compare-exchange, ConnectionIdentifier::null no longer caches a heap object, and TLSInfo_GetPidOffset computes directly from the TID offset. Comment prepared with Codex.

src/dmtcpworker.cpp
```diff
@@ -132,11 +134,15 @@ DmtcpWorker::determineCkptSignal()
...
-    sig = strtol(tmp, &endp, 0);
-    if ((errno != 0) || (tmp == endp)) {
+  if (!ckptSignalEnvInitialized) {
+    ckptSignalEnv = getenv(ENV_VAR_SIGCKPT);
+    ckptSignalEnvInitialized = true;
+  }
+
+  if (ckptSignalEnv != NULL) {
+    sig = strtol(ckptSignalEnv, &endp, 0);
+    if ((errno != 0) || (ckptSignalEnv == endp)) {
```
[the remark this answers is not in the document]
karya0: This is called during single-threaded operation.

karya0:
> This is clearly a much-needed PR. It makes the thread initialization portion much more robust. I'm listing this as "Request changes" for now (Copilot has multiple suggestions), but ping me if you want a quick "review and approve".
> 
> There are a few changes I'd like to see:
> 
> 1. Coderabbit has a review outstanding about " A NULL check immediately after JALLOC_MALLOC and returning EAGAIN keeps the change minimal while avoiding a null dereference on allocation failure".  This seems valid to me, but you can decide.
> 2. I asked Copilot to also do a review.  Copilot found multiple interesting issues.  Could you look them over and do them or comment?
> 3. Could you rebase from main?  This will also help my PR #1288, which is stacked on top of this PR.
> 
> Mainly, I'm still listing "Request changes" because Copilot found multiple other issues to look at. I haven't looked at those closely yet. I can return to "review and approve" after we've looked at that. (Some of the issues that the TSAN work is uncovering can be subtle.)

The Coderabbit NULL-check issue is old - the new code doesn't have JALLOC_MALLOC.

Addressed copilot issues and/or added comment suggesting why the issue is irrelevant.

PR #1286, opened 2026-07-21
Disable ckpt when using kvdb

xuyao0127:
kvdbRequest creates a temporary socket to communicate with the coordinator,which is not tracked by DMTCP. Therefore, it's unsafe to checkpoint when user's application, including DMTCP plugins, is requesting the KVDB.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved key-value database request handling during checkpoint operations, reducing the risk of interference and improving request reliability.
  * Preserved existing request validation and processing behavior, ensuring no changes to how valid requests are prepared or handled.

### DIFF
--- a/src/kvdb.cpp
+++ b/src/kvdb.cpp
@@ -8,6 +8,7 @@
 #include "jconvert.h"
 #include "kvdb.h"
 #include "threadinfo.h"
+#include "wrapperlock.h"
 
 namespace dmtcp
 {
@@ -45,6 +46,7 @@ KVDBResponse request(KVDBRequest request,
   msg.valLen = val.length() + 1;
   msg.extraBytes = msg.keyLen + msg.valLen;
 
+  WrapperLock wrapperLock;
   return CoordinatorAPI::kvdbRequest(msg, key, val, oldVal);
 }
 


gc00 (approved):
LGTM

src/kvdb.cpp
```diff
@@ -45,7 +45,13 @@ KVDBResponse request(KVDBRequest request,
   msg.valLen = val.length() + 1;
   msg.extraBytes = msg.keyLen + msg.valLen;
 
-  return CoordinatorAPI::kvdbRequest(msg, key, val, oldVal);
+  // kvdbRequest creates a temporary socket to communicate with the coordinator,
+  // which is not tracked by DMTCP. Therefore, it's unsafe to checkpoint when
+  // user's application, including DMTCP plugins, is requesting the KVDB.
+  DMTCP_PLUGIN_DISABLE_CKPT();
```
karya0: Please use the WrapperLock convention. Then you don't have to use disable/enable pairs.

karya0 (approved):
LGTM modulo the comment.

src/kvdb.cpp
xuyao00: @karya0 Can you point me to some code use WrapperLock as example?

xuyao00: Fixed

PR #1287, opened 2026-07-22
Cuda plugin

xuyao0127:
# Add CUDA checkpoint plugin

## Summary

This PR adds an optional plugin (plugin/cuda) that transparently checkpoints and restarts CUDA GPU state using NVIDIA's CUDA checkpoint API (cuCheckpointProcess*, introduced in CUDA 12.4), plus three small DMTCP-core changes required to make it work correctly.

The plugin locks and dumps GPU memory/state to host memory at DMTCP_EVENT_PRESUSPEND and restores it after resume/restart. It also saves and recreates the CUDA driver's pipes (a workaround until DMTCP 5.0 stops promoting pipes to sockets), reserves the /dev/nvidia* device-file mappings across checkpoint, and refuses to launch when the DMTCP: program-name prefix is enabled (the CUDA checkpoint API matches processes by name on restart).

### Plugin (1 commit)
- plugin/cuda/ — the plugin, its build (Makefile, C++20), a README.md, and an end-to-end GPU checkpoint/restart test (test/counter.cu + make check).

### Core changes (3 commits) — the part most worth reviewing closely:
1. Skip thread-state update for the checkpoint thread in the pthread_create wrapper. The CUDA checkpoint API spawns short-lived helper threads from the checkpoint thread during DMTCP_EVENT_PRESUSPEND. The wrapper's RUNNING → THREAD_CREATE state transition assumes the caller is a user thread (ST_RUNNING); the checkpoint thread is ST_CKPNTHREAD, so the transition would fail. We skip the transition when the caller is the checkpoint thread — those helper threads exit before the checkpoint is taken, so this is safe.
2. Disable the DMTCP: program-name prefix by default. The CUDA checkpoint API matches a process by name on restart, so the prefix breaks CUDA restart. Users can re-enable it with --enable-prgname-prefix or override at runtime via DMTCP_DISABLE_PRGNAME_PREFIX. Changed in the configure.ac source (not just generated configure).
3. Add DMTCP_EVENT_RUNNING_AFTER. A new event fired after DMTCP_EVENT_THREAD_RESUME and before DMTCP_EVENT_RUNNING, processed in reverse plugin order (the resume/restart ordering). The CUDA plugin restores GPU state here; it is also required by a companion NCCL plugin (not in this repo) that must re-initialize after the GPU is restored. Basic CUDA checkpointing works without it, but it is needed for correct layered plugin ordering.

## Notes for reviewers

- Optional plugin, not built by default. plugin/cuda has its own hand-written Makefile (cd plugin/cuda && make); it is not wired into the default build or make
check, and requires CUDA ≥ 12.4 (nvcc) + an NVIDIA GPU. Nothing in the default r when the plugin is unused.
- Pipe handling is a temporary workaround. inspect_pipes/recreate_pipes exist oy promotes pipes to sockets; there's a FIXME noting this can be removed oncethat changes in 5.0.

## Testing

On CUDA 13.0 + an NVIDIA RTX 3070:
- ./configure && make -j — core builds; ENABLE_PRGNAME_PREFIX off by default.
- cd plugin/cuda && make — plugin + test build cleanly under C++20.
- cd plugin/cuda && make check — PASS: the GPU __device__ counter resumes from its checkpointed value after restart.
- make check-dmtcp1 check-pthread1 — PASS (confirms the pthread_create wrapper threading).

## Summary by CodeRabbit

* **New Features**
  * Added CUDA checkpoint/restart support for CUDA applications, including programs that run GPU kernels.
  * Added plugin lifecycle events for “running after” launch, resume, and restart.
* **Documentation**
  * Added instructions for building, testing, and running the CUDA plugin.
* **Tests**
  * Added a CUDA counter test that exercises checkpoint and restart behavior.
  * CUDA support requires CUDA Toolkit 12.4 or newer.

### DIFF
--- a/src/threadwrappers.cpp
+++ b/src/threadwrappers.cpp
@@ -118,13 +118,14 @@ pthread_create(pthread_t *pth,
             "new thread wrapper lock was not pre-acquired: tid={}",
             newThread->tid);
 
-  ASSERT(Thread_UpdateState(thread, ST_THREAD_CREATE, ST_RUNNING),
+  ThreadState thread_state = (ThreadState) thread->state;
+  ASSERT(Thread_UpdateState(thread, ST_THREAD_CREATE, thread_state),
          "Failed to mark thread (tid:{}) from RUNNING to THREAD_CREATE",
          thread->tid);
 
   int retval = _real_pthread_create(pth, attr, thread_start, newThread);
 
-  ASSERT(Thread_UpdateState(thread, ST_RUNNING, ST_THREAD_CREATE),
+  ASSERT(Thread_UpdateState(thread, thread_state, ST_THREAD_CREATE),
          "Failed to mark thread (tid:{}) from THREAD_CREATE to RUNNING",
          thread->tid);
 


include/dmtcp.h
```diff
@@ -90,6 +90,7 @@ typedef enum eDmtcpEvent {
   DMTCP_EVENT_RESUME,
   DMTCP_EVENT_RESTART,
```
gc00, line 91: Let's change these two event names to:  `DMTCP_EVENT_RUNNING_AFTER_LAUNCH` and `DMTCP_EVENT_RUNNING_AFTER_CKPT_RESTART`.

@karya0 , We had mentioned that we need these two different events.  The reason is that events are run in forward order on launch/ckpt and in reverse order on restart.  But right now, you have `DMTCP_EVENT_RUNNING` as a single event.  Currently, it fails to be run in reverse order during RESTART.

@karya0, You're the best person to comment on this proposal to add _two_ events.  One would run after launch _only_, and one would run after restart _only_.  I'm guessing that you will want to modify this part of the code to handle that.  We need the new event for CUDA support.  Please suggest how you'd like to modify this PR, before we push it in.  :-)

plugin/cuda/Makefile
```diff
@@ -0,0 +1,138 @@
...
+	  echo "       but the CUDA checkpoint API requires CUDA >= $(CUDA_MIN_MAJOR).$(CUDA_MIN_MINOR)."; \
+	  exit 1; \
+	fi; \
+	echo "Using CUDA $$ver from $(CUDA_HOME)"
+
+${LIBOBJS}: | cuda-version-check
+tests: | cuda-version-check
+
+${LIBNAME}.so: ${LIBOBJS}
+	${LINK} -shared -fPIC -o $@ $^ -lcuda -ldl
```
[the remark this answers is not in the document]
gc00: @xuyao0127 , This comment by Code Rabbit seems important.  Some users will have only the driver library, and not a CUDA toolkit installed.

plugin/cuda/Makefile
```diff
@@ -0,0 +1,138 @@
+CC=gcc
```
gc00, line 1: QUESTION:  What happens for this Makefile if a user is on a machine with no CUDA library (and maybe no GPU)?

1. Is this CUDA plugin compiled by default?
2. Even if it's not compiled by default, does the Makefile finish with a graceful error message (e.g., CUDA library not found) and allow the rest of any build to contiue?

plugin/cuda/README.md
```diff
@@ -0,0 +1,21 @@
+CUDA checkpoint plugin
+======================
+
```
gc00, line 3: Suggested rewording of first sentence, below:
`This DMTCP plugin supports CUDA programs (programs that also execute CUDA kernels).`

src/threadwrappers.cpp
```diff
@@ -113,20 +113,31 @@ pthread_create(pthread_t *pth,
 
   Thread *newThread = ThreadList::getNewThread(start_routine, arg);
   ThreadSync::wrapperExecutionLockLockForNewThread(newThread);
-  ASSERT_NE(0u,
+  ASSERT_NE(0,
             newThread->wrapperLockCount,
             "new thread wrapper lock was not pre-acquired: tid={}",
             newThread->tid);
```
gc00, line 120: Is the code change below required for the CUDA?  I'm wondering if this was generated by the AI in porting from 4.2-dev to 5.0, and maybe it's not actually necessary.  I don't see any actual reference to CUDA in this code.
Or maybe this code is needed, and the AI comment doesn't explain well why this is needed for CUDA.

configure.ac
```diff
@@ -194,13 +194,14 @@ AC_ARG_ENABLE([experts_only],
 AC_ARG_ENABLE([experts_only_after], [AS_HELP_STRING([ ], [])], [], [])
 
 AC_ARG_ENABLE([prgname-prefix],
```
gc00, line 196: @karya0, Here, we're changing the configure default for _everybody_ to not use the program name prefix "DMTCP:".  We absolutely need this for CUDA ckpt compatibility.  What do you suggest?  Do we need to add a special configure flag (--with-cuda-support) just for this one little case (ugly) and restore the default?  Is there a way to decide at runtime if we're going to use CUDA and decide then?  Or maybe should we offer an environment variable, `DMTCP_SUPPORT_CUDA`, and use that at runtime, if we can't automatically decide at runtime.

Having DMTCP support CUDA "out-of-the-box" for naive users would be a nice feature.  So, automatically deciding at runtime would be nice.  This might interact with whether we automatically compile the CUDA plugin, but I'm just speculating.

src/threadwrappers.cpp
xuyao00: thread_state = ST_CKNTHREAD or ST_RUNNING
Thread_UpdateState(thread, ST_THREAD_CREATE, thread_state)
_real_pthread_create(...)
Thread_UpdateState(thread, thread_state, ST_THREAD_CREATE)

plugin/cuda/cuda-ckpt.cpp
```diff
@@ -0,0 +1,405 @@
...
+#define DMTCP_LOG_COMPONENT "cuda"
+#include "dmtcp_assert.h"
+#include "procselfmaps.h"
+#include "procmapsarea.h"
+#include "trampolines.h"
+
+// Maximum path length for /proc/self/fd/ entries and readlink buffer
+#define PATH_MAX 4096 
+// Max number of file descriptors we'll track (arbitrary limit)
+#define MAX_PIPE_FDS 1024
```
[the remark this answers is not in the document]
gc00: Yes. I agree with the suggested change.  I've been burnt by that before.

plugin/cuda/cuda-ckpt.cpp
```diff
@@ -0,0 +1,405 @@
...
+  int kept = 0;
+  for (int i = 0; i < num_fds; i++) {
+    int fd = pipe_fd_array[i].fd;
+    if (fd != fd_unique && fcntl(fd, F_GETFD) != -1) {
+      continue; // occupied by another subsystem's restored pipe; not ours
+    }
+    pipe_fd_array[kept++] = pipe_fd_array[i];
+    dup2(fd_unique, fd); // Occupy fd; close it later.
+  }
+  num_fds = kept;
```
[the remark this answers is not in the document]
gc00: This one is minor and optional.  It's rare that we would reach the max number of fd's.

include/dmtcp.h
```diff
@@ -90,6 +90,17 @@ typedef enum eDmtcpEvent {
...
   DMTCP_EVENT_RUNNING,
```
gc00, line 93: Maybe add a comment on the above line (`DMTCP_EVENT_RUNNING`): // This event will be replaced by others.

plugin/cuda/cuda-ckpt.cpp
```diff
@@ -325,18 +325,10 @@ static void cuda_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data) {
   switch (event) {
     case DMTCP_EVENT_INIT:
       // The CUDA checkpoint API matches the process by its name on restart, so
```
gc00, line 324: There's a typo on the commit msg for this commit (commit #4):  disbales -> disables

gc00 (approved):
Please also read the conversation tab, where I added my own comments, and some comments on the suggestions of coderabbit.  I also asked Copilot to review this, and it will add more comments.

Anyway, all the comments are minor .  So, I'm approving it now.

plugin/cuda/Makefile
```diff
@@ -0,0 +1,151 @@
...
```
karya0, line 1: Any reason for not using Makefile.am or Makefile.in here?

Is this going to be standalone for now? I don't see configure.ac or top-level Makefile.in changes.

xuyao00: Hi Kapil. I was thinking not all users need this plugin and not all computer support this. Currently, I added this plugin as an optional and standalone plugin. Users who need CUDA support need to manually compile this plugin before using. Alternatively, I can add automatic detecting or an extra flag in the configure script to include this plugin. Which approach do you prefer?

plugin/cuda/Makefile
xuyao00: This CUDA plugin is not compiled by default. User needs to cd into the CUDA plugin directory and compile manually. On machines that doesn't have CUDA and nvcc, it will fail because there's no nvcc. There no separate error message.

plugin/cuda/README.md
xuyao00: Fixed.

include/dmtcp.h
```diff
@@ -90,6 +90,17 @@ typedef enum eDmtcpEvent {
...
+  // Currently, they are triggered at the same time as DMTCP_EVENT_RUNNING.
+  // Plugins using these events needs to determine the current status by
+  // themselves.
+  //
+  // DMTCP_EVENT_RUNNING_AFTER_LAUNCH calls plugins in forward order.
+  // DMTCP_EVENT_RUNNING_AFTER_RESUME/RESTART calls plugins in the
+  // reverse order.
+  DMTCP_EVENT_RUNNING_AFTER_LAUNCH,
+  DMTCP_EVENT_RUNNING_AFTER_RESUME,
+  DMTCP_EVENT_RUNNING_AFTER_RESTART,
```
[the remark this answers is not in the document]
xuyao00: No plugin uses explicit numeric assignments.

plugin/cuda/cuda-ckpt.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+static void restore_gpu() {
+  int pid = getpid();
+  // Release reserved memories for NVIDIA device files.
+  for (Area device_area : nvidia_device_files) {
+    munmap(device_area.addr, device_area.size);
+  }
+  CUcheckpointUnlockArgs unlock_args = {0};
+  CUresult ret = cuCheckpointProcessRestore(pid, &restore_args);
+  ASSERT(ret == CUDA_SUCCESS, "cuCheckpointProcessRestore failed: {}", (int)ret);
```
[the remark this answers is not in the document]
xuyao00: Actually we don't need to call cuInit() manually. So we can remove cuda_initialized.

plugin/cuda/cuda-ckpt.cpp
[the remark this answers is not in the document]
xuyao00: Fixed.

plugin/cuda/cuda-ckpt.cpp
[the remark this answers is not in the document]
xuyao00: Ignored.

include/dmtcp.h
xuyao00: Fixed.

include/dmtcp.h
```diff
@@ -90,6 +90,17 @@ typedef enum eDmtcpEvent {
...
   DMTCP_EVENT_THREAD_RESUME,
```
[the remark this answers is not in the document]
xuyao00: fixed.

plugin/cuda/cuda-ckpt.cpp
xuyao00: fixed

PR #1294, opened 2026-07-31
Support lightweight guard pages

xuyao0127:
## Problem

  Linux 6.13 added lightweight guard regions (`madvise(MADV_GUARD_INSTALL)`), and
  glibc 2.42 uses one at the low end of every new thread stack.

  A guard page raises SIGSEGV on any access, and the kernel reports a guard PTE with
  the *swap* bit set. So `Util::scanOccupiedRangeBatch()` classifies it as occupied,
  `writeCkptAll()` reads it, `Util::writeAll()` gets EFAULT, and the ASSERT at
  `writeckpt.cpp:79` aborts the checkpoint. Symptom: `pthread1`-`pthread4` fail on
  glibc >= 2.42.

  Commit e654d4b81d worked around this by calling `MADV_GUARD_REMOVE` right after
  every `pthread_create`. That permanently disables stack-overflow detection for the
  whole process, and only covers glibc's own guards -- an application calling
  `madvise(MADV_GUARD_INSTALL)` still aborts the checkpoint.

  ## Solution

  Detect guard pages from the kernel rather than tracking them:

  - `Util::guardPagesSupported()` / `Util::scanNextGuardRange()` read the
    `PAGE_IS_GUARD` category via `ioctl(PAGEMAP_SCAN)` (Linux 6.15+), or pagemap
    bit 58 on the `pread()` fallback.
  - `GuardPages` (new `src/guardpages.{h,cpp}`) records the ranges during the
    existing pre-pass in `mtcp_writememoryareas()`, and replays them after restart.
  - Guard pages are never read and never uninstalled. A guard run is written as an
    ordinary `DMTCP_ZERO_PAGE` run, so protection is continuous in the running
    process, the resume path needs no work, and the checkpoint image format is
    unchanged.
  - `GuardPages::reinstall()` runs in `ThreadList::waitForAllRestored()`, after the
    `DMTCP_EVENT_RESTART` hooks (a plugin may copy out of restored memory and
    re-mmap it) and before any user thread is released.
  - The e654d4b81d workaround is deleted.

  The list lives in DMTCP's heap, which is itself checkpointed, so it survives
  restart with no image record -- hence it must be complete before the first area is
  written.

  ## Platform-specific behavior

  - **`madvise(2)` understates where guard pages can appear.** It documents
    MADV_GUARD_INSTALL as working only on writable anonymous private mappings; on
    6.19 it also succeeds on MAP_SHARED anonymous, private and shared memfd,
    `/dev/zero`, and file-backed mappings (only mlock'd fails). Scanning is
    therefore not restricted by mapping type, and the file-backed single-blob write
    path brackets the write with remove/reinstall, since it has no per-run headers
    to carve a guard page out of.
  - Install succeeds on read-only and PROT_NONE mappings, so reinstall needs no
    `mprotect`.
  - `[heap]` and `[stack]` use the *content* scan (`Util::areZeroPages()` reads
    every page) because `residencyScanSafe` requires an unnamed area; the guard
    carve bounds that path too.
  - **Linux 6.13 and 6.14 are not supported**: guard pages exist there but no
    interface reports them. A single runtime probe treats "can detect" as "can
    exist", so on those kernels behavior is unchanged from today. Both are EOL
    upstream.
  - Kernels before 6.13, and older build headers, are handled by fallback
    `#define`s plus the runtime probe; nothing is allocated when unsupported.
  upstream.
- Kernels before 6.13, and older build headers, are handled by fallback
  `#define`s plus the runtime probe; nothing is allocated when unsupported.

## Testing

./configure && make -j$(nproc)
make check AUTOTEST="--retry-once --jobs 8"   # pass=133 fail=1 skipped=14 total=148
make check-unit                                # 76/76 (11 new guard-region tests)
make check-guardpage  make check-guardpage-pread
make clean && make                             # from scratch, warning-free

- New integration test `test/guardpage.c` (+ a `-pread` variant forcing the
  fallback) guards two disjoint ranges in an anonymous region, one in a read-only
  region, and one in `[heap]`, then re-verifies after every checkpoint and restart
  that each guarded page still faults, its neighbours are byte-exact, and no other
  page faults.
- Confirmed the tests fail when the feature is broken: disabling the reinstall
  yields `FAIL: guarded page 5 is readable; guard lost`; disabling the carve
  aborts the checkpoint.
- `pthread1`-`pthread6` pass with the workaround removed. A traced run of
  `pthread2 8` records 18 guard ranges and reinstalls all 18 after restart,
  including glibc's cached thread stacks.

## Summary by CodeRabbit

* **New Features**
  * Added checkpoint and restart support for Linux lightweight guard pages.
  * Preserves guard-page protections and surrounding memory contents across checkpoint cycles.
  * Supports multiple memory-scanning methods and safely skips unsupported kernels.

* **Bug Fixes**
  * Prevents guard pages from being incorrectly included in memory ranges or lost during restart.

* **Tests**
  * Added unit and integration coverage for guard-page detection, restoration, fault behavior, and fallback scanning.


gc00 (changes requested):
Let me know when you want a _final_ review.  In the meantime, Coderabbit seems to have found two important robustness issues.  I find that you can point the AI to the URL for this PR, and then ask it if it agrees with Coderabbit.  After you decide what to do, you can click on "Resolve" for that comment.  Also, I have now asked for a Copilot review, in addition.

(And please rebase with the latest 'origin/main', just in case that affects anything.)

src/guardpages.cpp
```diff
@@ -0,0 +1,173 @@
...
+            removing ? "drop" : "restore",
+            removing ? "reading the area may fault"
+                     : "the application's guard page is no longer in effect",
+            (void *)guardStart, guardEnd - guardStart, area.name);
+    count++;
+    addr = guardEnd;
+  }
+
+  return count;
+}
```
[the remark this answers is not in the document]
xuyao00: Fixed

src/util_misc.cpp
```diff
@@ -602,6 +602,65 @@ Util::areZeroPages(void *addr, size_t numPages)
...
+
+    munmap(page, page_size);
+    return detected;
+#else // ifdef PAGEMAP_SCAN
+    return false;
+#endif // ifdef PAGEMAP_SCAN
+  }();
+
+  return supported;
+}
```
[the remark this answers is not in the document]
xuyao00: Fixed.

xuyao00:
@coderabbitai full review

src/util_misc.cpp
```diff
@@ -602,6 +638,66 @@ Util::areZeroPages(void *addr, size_t numPages)
...
+  struct page_region region;
+  int savedErrno = 0;
+  const long ret =
+    scanForGuardPages(probePage, probePage + page_size, &region, &savedErrno);
+  const bool supported = (ret >= 0);
+
+  TRACE("lightweight guard page support: supported={} errno={}",
+        supported, supported ? 0 : savedErrno);
+  return supported;
+}
```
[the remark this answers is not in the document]
xuyao00: If the program was checkpointed on a platform that support lightweight guard page, we assume it will be restarted on a platform that support the same feature. So we don't need to consider the case that PAGEMAP_SCAN is not supported on restart.

xuyao00:
Hi @gc00, this PR is ready for final review.

src/util_misc.cpp
```diff
@@ -602,6 +638,66 @@ Util::areZeroPages(void *addr, size_t numPages)
...
+Util::guardPagesSupported()
+{
+  // Any page that is certainly mapped will do; this one is in libdmtcp's .bss.
+  static int probeAnchor = 0;
+
+  const size_t page_size = pageSize();
+  const uintptr_t probePage =
+    (uintptr_t)&probeAnchor & ~(uintptr_t)(page_size - 1);
+
+  // The kernel validates the category mask before walking any page table, so
```
gc00: Typo:
  page table, so  ->  page table. So

[Claude doesn't seem to understand that 'so' can be an interjection (here) or a conjunction.]

src/util_misc.cpp
```diff
@@ -602,6 +638,66 @@ Util::areZeroPages(void *addr, size_t numPages)
...
+  // failure here is a pure capability answer.  The anchor page is never
+  // guarded, so a supported kernel returns 0; only the sign matters.
```
gc00: guarded, so ->  guarded. So

src/util_misc.cpp
```diff
@@ -739,6 +835,50 @@ Util::scanOccupiedRangeBatch(uintptr_t start, uintptr_t end,
...
+  struct page_region region;
+  int savedErrno = 0;
+  const long ret =
+    scanForGuardPages(startAddr, endAddr, &region, &savedErrno);
+
+  if (ret == 0) {
+    return false; // no guard page in this range
+  }
+  if (ret < 0) {
+    // Callers gate on guardPagesSupported(), so what is left is a range the
```
gc00: , so -> . So

src/util_misc.cpp
xuyao00: Fixed.

src/util_misc.cpp
xuyao00: Fixed.

src/util_misc.cpp
xuyao00: Fixed.

